PR #14, opened 2017-03-30
Feature/search

youngmit:


djabaay:
Hey Mitch,

It looks like you're creating stuff that I think already exists in ArrayUtils.f90.  Would you mind taking a look and seeing if that is true?  It may or may not exist for 1-D ints, but I think something similar should be there for 1-D reals.

-Dan

youngmit:
TL;DR: I recommending either deprecating or removing the search code in ArrayUtils. ArrayUtils conceptually should be about higher-order array utilities (things like findEleHtAbove()), while raw search functionality should be in Search.f90. We already have array sorting functionality in Sorting.f90, so this is more consistent. Justification below:

I see the code you're talking about. A couple of observations:
 - Many of the search functions aren't being called anywhere within Futility or in MPACT, so they can be safely removed or replaced.
 - The findLowBound and findUpBound interfaces return values, rather than indices. This makes them more limited in their usefulness. Imagine if you wanted to index, you'd have to call findLowBound(), then use findIndex() to get the corresponding index. Returning the index should be preferred, as the client code has access to the original sequence; if they want the value, they can just dereference it with the returned index.
 - The find*Bound interfaces also advertise semantics that disagree with their canonical forms. lower_bound should return the index of the first element that does not compare less than the search value, while upper_bound should return the index of the first element that compares greater than the search value. This distinction is subtle, yet important in the case of non-float datatypes.
 - The actual implementations of the search functions is not very performant. Internal copies are made of the passed search array, and a linear search is used instead of a binary search. I get that making the copy allows for doing fancy things like search an array as if it is specified using deltas instead of absolutes, but such cases could be better served by providing functions in ArrayUtils to generate an array of absolutes from deltas, perform the conversion in the client code, then perform searches on the array of absolutes. Most use cases will perform many searches on the same array, so pre-computing the absolutes is the way to go anyways.
 - All of the extra functionality hinted at in the above complicates the actual interface, which along with the fact that the search is slower than an in-place linear search, probably explains why these aren't getting used anywhere in the first place.
 - We already have a Sorting module, separate from ArrayUtils, so having the Search functionality be separate also makes sense. IMO "ArrayUtils" as a name is pretty vague about its contents. I think it does have merit, though. I think conversions from deltas to absolutes, functions like findEleHtAbove() and findEleHtBelow() are good candidates for living in such a module.

bkochuna:
So Mitch, the search capability in ArrayUtils would just be replaced by calls to Search.f90?

youngmit:
Where search functionality is used --actually called-- in ArrayUtils, yes. And the interface for findLowBound and findUpBound should be simplified to their canonical form. Now that Futility is on its own, we really need to be careful with API design since future changes may break dependent code. The interfaces on findLow and findUp are frankly dangerous to have in the library.

youngmit:
I am only seeing them used by the findEleHtAbove_1DReal and findEleHtBelow_1DReal which also do not appear to be called anywhere, and also have a more complicated interface than really seems necessary. I can imagine most client code that uses these features becoming more complicated than implementing them in-place.

aarograh:
I agree with Mitch that separating and simplifying some of these routines
would be useful.  When I refactored ControlRodMovement, I moved away from
the ArrayUtils calls and did a lot of it in place because some of the
optional arguments and return values made it tricky to ensure I was
actually getting what I needed.  It would be nice to simplify some of the
interfaces and split this functionality into Sorting/Search/Other modules
so it's a bit more user-friendly than it currently is.



---------------------------------------------------------------------------
Aaron Graham
University of Michigan Nuclear Engineering
---------------------------------------------------------------------------

youngmit:
I think this is in a good place to merge now, assuming everyone is happy where things ended up.

src/IntrType.f90
```diff
@@ -98,6 +98,7 @@ MODULE IntrType
   PUBLIC :: SOFTEQ
   PUBLIC :: SOFTEQR
   PUBLIC :: SOFTLE
+  PUBLIC :: SOFTLT
```
youngmit: Implemented SOFTGT, but there doesn't appear to be a direct test for IntrTypes. Might punt that to a separate issue.

src/Search.f90
```diff
@@ -0,0 +1,286 @@
...
+  IMPLICIT NONE
+
+  PRIVATE
+
+  PUBLIC :: upperBound
+  PUBLIC :: lowerBound
+  PUBLIC :: find
+  PUBLIC :: binarySearch
+
+  !> @brief Return the index of the first element of a list that compares
```
youngmit: Yeah, I need to check with doxygen on this. I'm of a mind that it's better to take the inconvenience of separating the documentation from the implementations because it's less likely to end up with implementations with disparate semantics and documentation. In a perfect world, we could manage to keep all of this stuff in sync, but... pobody's nerfect.

src/Search.f90
```diff
@@ -0,0 +1,286 @@
...
+  !> @param list the list to search. The list should be sorted, otherwise
+  !> undefined behavior.
+  !> @param val the value to find
+  !> @param tol the comparison tolerance to use for floating-point
+  !> implementations. Optional
+  !> @returns the index to the first element of the list that compares greater
+  !> than the passed search value
+  !>
+  !> If the value is greater than or equal to the largest element of the list,
+  !> returns SIZE + 1.
```
youngmit: We could. I went with size+1, since it mimics "end" in iterator-based approaches. I also don't want to imply any meaning beyond "not found," such as 0 if value is less than and size+1 if greater than all elements in the list.

src/Search.f90
```diff
@@ -0,0 +1,286 @@
...
+  !> compare less than the passed value
+  !> @param list the list to search. The list should be sorted, otherwise
+  !> undefined behavior
+  !> @param val the value to find
+  !> @param tol the comparison tolerance to use for floating-point
+  !> implementations. Optional
+  !> @returns the index to the first element of the list that does not compare
+  !> less than the passed value
+  !>
+  !> If the value is greater than the last element of the list, returns SIZE+1
```
youngmit: This is correct. lower bound will return the index of the last element if the search value equal the last element. upper_bound wants to find the first element that is greater, therefore it will not find an element if the search value is equal to the last element in the sequence

src/Search.f90
youngmit: Tried building with Doxygen, but there is no Doxygen file for building the Futility docs standalone. There should be, so I created Issue #15. As for the location, I strongly prefer that the documentation live in one place. As a generic interface, the provided documentation should apply to all implementations, and as a user of the library, one shouldn't have to dig down to a specific implementation to see what a generic interface does.

bscollin:
I'm ready to merge.  Any objections need to be reported by 5 pm EST.

PR #17, opened 2017-04-20
Hotfix 4914 close hdf5

djabaay:
Merge request for PHI ticket #4914
https://vminfo.casl.gov/trac/casl_phi_kanban/ticket/4914

Needed for fixing MPACT's ability to read and write to the same restart file in the same case.

Commit descriptions describe the changes.


bkochuna:
Only a couple comments on review.
1. Commit 10ba9c3 - modify the commit short message to state which unit test is being updated.
2. Commit be390b4 - modify commit description to describe the nature of the changes (e.g. fixing indenting). In the short description say which file
3. Commit 7b6bc04 - state which file type has the closure behavior modified

General: remove "." from short message in commit

PR #18, opened 2017-04-25
Polygonize_Cylinder routine added.

mgjarrett:
The new Polygoinze_Cylinder routine is similar to the already-existing Polygonize_Circle routine. This was added for visualizing general cylinders in VTK. Test for cylinder added to Geom_Poly unit test.

### DIFF
--- a/src/Geom_Poly.f90
+++ b/src/Geom_Poly.f90
@@ -122,6 +122,7 @@ MODULE Geom_Poly
 
   INTERFACE Polygonize
     MODULE PROCEDURE Polygonize_Circle
+    MODULE PROCEDURE Polygonize_Cylinder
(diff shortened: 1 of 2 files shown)


src/Geom_Poly.f90
```diff
@@ -1632,6 +1633,57 @@ SUBROUTINE Polygonize_Circle(circle,polygon)
     ENDSUBROUTINE Polygonize_Circle
 !
 !-------------------------------------------------------------------------------
+!> @brief This routine will return a polygon type for a given circular geometry.
+!> @param circle The circle type to be turned into a polygon
```
bkochuna: This comment header needs to be updated

bkochuna:
Mike... Last thing. The commit messages need to be updated to have the correct formatting.

PR #21, opened 2017-05-12
Stop deallocating LU factorization all the time

bscollin:
Current implementation would clear and deallocate the preconditioner whenever the matrix
is changed or the solver failed.  This ends up spending significant time in parameter list
and allocation time for mongoose which is solving a temperature equation many many times
per solve.  The changes represent about an order of magnitude speedup in mongoose.

### DIFF
--- a/src/LinearSolverTypes.f90
+++ b/src/LinearSolverTypes.f90
@@ -828,10 +828,6 @@ MODULE LinearSolverTypes
         TYPE IS(LinearSolverType_Direct)
           IF(ALLOCATED(solver%IPIV)) CALL demallocA(solver%IPIV)
       ENDSELECT
-      IF(ALLOCATED(solver%M)) THEN
-        CALL solver%M%clear()
-        DEALLOCATE(solver%M)
-      ENDIF
     ENDSUBROUTINE updatedA
 !
 !-------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


bscollin:
This is a pretty simple change but I'd like a few pairs of eyes since I'm deleting deallocations.  it looks like myLS%clear() will also clear M and deallocate it so I think we're good as far as memory leaks.  All unit tests seem to work fine too.

bkochuna:
Just summarizing my understanding, but this change should only affect our internal direct solvers.
The attribute for the M matrix is used to store a decomposition of matrix A.
Presently M is deallocated and reallocated each time A is decomposed.

I agree this does not need to happen. It doesn't seem like to me that M should deallocated and reallocated every time the matrix is decomposed, especially for tri-diagonal and dense matrices, the size of M is fixed based on the size of A. If the ILU is ILU(0) (which it looks to be) then the non-zero locations don't change and there is no need to reallocate M.

At most we should just be setting all the values in M to 0 or something each time A changes, not deallocate it.

Change makes sense to me, looks right to me. The only thing I have to add is it looks like a similar change could be made to the top of the `DecomposePLU_Dense` and `DecomposeILU_Sparse`. So maybe just knock those out too and then merge.

bscollin:
Pushed the changes.  Is this acceptable now?

bkochuna:
Yep. I think this looks fine. Shall I merge?

bscollin:
I opened to reply and realized I could just merge it.  Done.  Thanks

PR #22, opened 2017-05-15
VTU Implementation

zdodson94:
The additions made in this branch add the capability to print the geometry and flux results in parallel to vtu files, based on the spatial decomposition of the problem. A pvtu file is created, which acts as a master file when opening the output in a visualization program. To switch on this capability, add an extra "T" to the vis_edits card. New unit tests have been added for this capability and all nightly tests have passed.


unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      CALL e%setStopOnError(.FALSE.)
+      CALL e%setQuietMode(.TRUE.)
+      CALL testVTUFile%e%addSurrogate(e)
+      CALL eVTK%addSurrogate(e)
+    ENDSUBROUTINE SetupError
+!
+!-------------------------------------------------------------------------------
+    SUBROUTINE testInitialize
+      !Test Initialize and clear
+      CALL testVTUFile%clear
```
bkochuna: always put parenthesis after function/subroutine calls.

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      ASSERT(bool,'testVTUFile%initialize(...)')
+      CALL testVTUFile%clear(.TRUE.)
+    ENDSUBROUTINE testInitialize
+!
+!-------------------------------------------------------------------------------
+    SUBROUTINE testFailSTRUCTURED_POINTS
+      !Test writeMesh for STRUCTURED_POINTS (should fail, not supported)
+      CALL testVTUFile%e%addSurrogate(e)
+      CALL testVTUFile%initialize(666,'testVTU1.vtu')
+      CALL SetupTest1_Mesh
```
bkochuna: parentheses

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      bool = .NOT.testVTUFile%hasMesh
+      ASSERT(bool, 'testVTUFile%writeMesh(...) STRUCTURED_POINTS')
+      CALL testVTUFile%clear(.TRUE.)
+    ENDSUBROUTINE testFailSTRUCTURED_POINTS
+!
+!-------------------------------------------------------------------------------
+    SUBROUTINE testUNSTRUCTURED_GRID
+      !Test writeMesh for UNSTRUCTURED_GRID
+      CALL testVTUFile%initialize(666,'testVTU2.vtu')
+      CALL SetupTest2_Mesh
```
bkochuna: parentheses

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      CALL testVTUFile%writeMesh(testVTKMesh)
+      bool = testVTUFile%hasMesh
```
bkochuna: "=" should not have spaces on either side

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      ASSERT(bool, 'testVTUFile%writeMesh(...) UNSTRUCTURED_GRID')
+
+      !Write data on the mesh
+      CALL SetupTest2_Data
```
bkochuna: This should be a component test

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      CALL testVTUFile%writeScalarData(testVTKData)
+      testVTKData%varname='mat_val'
+      testVTKData%vtkDataFormat='float'
+      CALL testVTUFile%writeScalarData(testVTKData)
+
+      !Test writeScalarData
+      testVTKData%vtkDataFormat='long'
+      testVTKData%varname='mat_long'
+      CALL testVTUFile%writeScalarData(testVTKData) !repeat for long
```
bkochuna: should ASSERT something after each of these calls if possible

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      testVTKData%varname='mat_double'
+      CALL testVTUFile%writeScalarData(testVTKData) !write as double
+      CALL testVTUFile%close
+      CALL testVTUFile%clear
+    ENDSUBROUTINE testUNSTRUCTURED_GRID
+!
+!-------------------------------------------------------------------------------
+    SUBROUTINE testMultiVTU
+      !Test writeMesh for UNSTRUCTURED_GRID
+      CALL testPVTUFile%clear
```
bkochuna: parentheses

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+      !Write data on the mesh
+      CALL testPVTUFile%writeScalarData(testVTKData)
+      CALL testPVTUFile%close
+      CALL testPVTUFile%clear
+
+      CALL testPVTUFile%writepvtu(666,'testPVTU',2)
+    ENDSUBROUTINE testMultiVTU
+!
+!-------------------------------------------------------------------------------
+    SUBROUTINE SetupTest1_Mesh
```
bkochuna: parentheses for function/subroutine definitions also, even when there are no arguments.

reasoning is to be able to differentiate variables from functions

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+  TYPE(ExceptionHandlerType),POINTER,SAVE :: e
+  TYPE(VTKMeshType),SAVE :: testVTKMesh
+  TYPE(VTKDataType),SAVE :: testVTKData
+  TYPE(VTUXMLFileType),SAVE :: testVTUFile
+  TYPE(PVTUXMLFileType),SAVE :: testPVTUFile
+  LOGICAL(SBK) :: bool
+
+  CREATE_TEST('Test VTU Files')
+
+  CALL SetupError
```
bkochuna: parantheses

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+
+  REGISTER_SUBTEST('INITIALIZE',testInitialize)
```
bkochuna: You should have a test clear before any of this.

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -0,0 +1,232 @@
...
+  REGISTER_SUBTEST('STRUCTURED POINTS',testFailSTRUCTURED_POINTS)
+  REGISTER_SUBTEST('UNSTRUCTURED GRID',testUNSTRUCTURED_GRID)
+  REGISTER_SUBTEST('MULTIFILE DATASET',testMultiVTU)
+
+  CALL DellocateAll
```
bkochuna: parantheses

src/VTUFiles.f90
```diff
@@ -1,643 +1,497 @@
...
+!> 
+!> This routine supports writing of UNSTRUCTURED_GRIDS. The other types of 
+!> geometry/topology are not yet implemented. The display of this 
+!> data was verified by loading and displaying it with VisIt.
+!>
+    SUBROUTINE writeMesh_VTUXMLFileType(myVTKFile,vtkMesh)
+      CHARACTER(LEN=*),PARAMETER :: myName='writeMesh_VTUXMLFileType'
+      CLASS(VTUXMLFileType),INTENT(INOUT) :: myVTKFile
+      TYPE(VTKMeshType),INTENT(IN) :: vtkMesh
+      CHARACTER(LEN=256) :: aline,sint
```
bkochuna: What is the justification for 256 here? Do you anticipate any problems with this?

bkochuna: What is the justification of 256 here, will that be sufficient?

src/VTUFiles.f90
```diff
@@ -1,643 +1,497 @@
...
+  !
+  !> Module name for error messages
+  CHARACTER(LEN=*),PARAMETER :: modName='VTUFILES'
+  
+  !> @brief Derived type object for files definable by the VTU XML format
+  !>
+  !> This is an extension of the @ref FileType_Fortran "VTKLegFileType".
+  TYPE,EXTENDS(VTKLegFileType) :: VTUXMLFileType
+  !
+  CHARACTER(LEN=256),ALLOCATABLE,DIMENSION(:) :: dataFormatList
```
bkochuna: Why are you not using the StringType here?

src/VTUFiles.f90
```diff
@@ -1,643 +1,497 @@
...
+  INTEGER(SIK) :: dataFormatListLength = 0
```
bkochuna: Whitespace around "="

src/VTUFiles.f90
```diff
@@ -1,643 +1,497 @@
...
+              CASE(VTK_UNSTRUCTURED_GRID)
+                myVTKFile%mesh=vtkMesh
+                myVTKFile%hasMesh=.TRUE.
+                !Write a mesh that is an unstructured grid
+                !Clean-up redundant points
+                CALL myVTKFile%mesh%cleanupPoints()
+                WRITE(sint,'(i64)') myVTKFile%mesh%numPoints; sint=ADJUSTL(sint)
+                WRITE(aline,'(i64)') myVTKFile%mesh%numCells; aline=ADJUSTL(aline)
+                WRITE(funit,'(a)') '  <UnstructuredGrid>'
+                WRITE(funit,'(a)') '    <Piece NumberOfPoints="'//TRIM(sint)//'" NumberOfCells="'//TRIM(aline)//'">'
```
bkochuna: Another formatting guideline is to try to keep lines from extending beyond 80-columns. The purpose of this guideline is to help with readability.

Consider continuing some source lines across multiple lines. e.g.
```
WRITE(*,*) "STuff", &
  "indent 2-spaces continuing to subsequent line"
```

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -49,92 +50,118 @@ SUBROUTINE SetupError
     ENDSUBROUTINE SetupError
 !
 !-------------------------------------------------------------------------------
-    SUBROUTINE testInitialize
-      !Test Initialize and clear
-      CALL testVTUFile%clear
+    SUBROUTINE testClear()
```
bkochuna: In these tests we typically manually set some attributes on the type then call clear.

We then ASSERT everything we can to make sure the object is in its defined uninitialized state. e.g. assert on all attributes (even the inherited ones).

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -49,92 +50,118 @@ SUBROUTINE SetupError
...
+      ASSERT(bool,'testVTUFile%clear(...)')
+    ENDSUBROUTINE testClear
+!
+!-------------------------------------------------------------------------------
+    SUBROUTINE testInitialize()
+      !Test initialize
       CALL testVTUFile%initialize(666,'testVTU0.vtu')
-      bool = testVTUFile%isInit() .AND. testVTUFile%isOpen()
+      bool=testVTUFile%isInit().AND.testVTUFile%isOpen()
       ASSERT(bool,'testVTUFile%initialize(...)')
```
bkochuna: There should be more asserts here. e.g. after initialization does the object have a mesh? etc.

It's better to not combine assertions on attributes. This obfuscates the debugging process. I recommend something like:
```
ASSERT(testVTUFile%isInit(),'%isInit()')
ASSERT(testVTUFile%isOpen(),'%isOpen()')
```
etc.

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -49,92 +50,118 @@ SUBROUTINE SetupError
...
+      CALL testVTUFile%hasData('mat_double','double',bool)
+      ASSERT(bool,'testVTUFile%hasData(...) DOUBLE')
+      !
+      CALL testVTUFile%close()
+      CALL testVTUFile%clear()
     ENDSUBROUTINE testUNSTRUCTURED_GRID
 !
 !-------------------------------------------------------------------------------
-    SUBROUTINE testMultiVTU
+    SUBROUTINE testMultiVTU()
```
bkochuna: This routine needs some ASSERTS as well. If you're unsure about this, the question to ask yourself is "what is the clearest and most direct way to assert that what I am testing works?"

src/VTUFiles.f90
zdodson94: A length of 256 should be sufficient for all feasible real world cases. The only situation in which this would not be enough is if each of the numbers in the connectivity block are over 31 digits long (after accounting for the spaces between each). Therefore, increasing this should not be necessary.

src/VTUFiles.f90
```diff
@@ -393,13 +409,38 @@ SUBROUTINE hasData_VTUXMLFileType(fileobj,varName,varFormat,bool)
...
+      CHARACTER(LEN=*),PARAMETER :: myName='hasFile_VTUXMLFileType'
+      CLASS(VTUXMLFileType),INTENT(INOUT) :: fileobj
+      CHARACTER(LEN=*),INTENT(IN) :: fname
+      LOGICAL(SBK),INTENT(INOUT) :: bool
+      INTEGER(SIK) :: i
+      !
+      bool=.FALSE.
+      IF(ALLOCATED(fileobj%fileList)) THEN
+        DO i=1,fileobj%numFiles
+          IF (fname==fileobj%fileList(i)) bool=.TRUE.
```
bkochuna: This loop will continue to execute after bool has been set to true.

Modify to:
```
IF(fname == fileobj%fileList(i)) THEN
  bool=.TRUE.
  EXIT
ENDIF
```

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -52,17 +51,23 @@ SUBROUTINE SetupError()
...
 !-------------------------------------------------------------------------------
     SUBROUTINE testInitialize()
       !Test initialize
       CALL testVTUFile%initialize(666,'testVTU0.vtu')
-      bool=testVTUFile%isInit().AND.testVTUFile%isOpen()
-      ASSERT(bool,'testVTUFile%initialize(...)')
+      ASSERT(testVTUFile%isInit(),'testVTUFile%isInit()')
+      ASSERT(testVTUFile%isOpen(),'testVTUFile%isOpen()')
+      ASSERT(.NOT.testVTUFile%hasMesh,'testVTUFile%hasMesh')
+      ASSERT(testVTUFile%numDataSet==0,'testVTUFile%numDataSet')
```
bkochuna: More formatting stuff.
- I think its better to remove the "testVTUFile" from the message for assert. This part of the message is kind of redundant.
- Relational operators should have surrounding whitespace (e.g. `==` should be ` == `).

unit_tests/testVTUFiles/testVTUFiles.f90
```diff
@@ -138,10 +143,14 @@ SUBROUTINE testMultiVTU()
       CALL SetupTest2_Mesh()
       CALL testVTUFile%writeMesh(testVTKMesh)
       bool=testVTUFile%hasMesh
-      ASSERT(bool,'testVTUFile%writeMesh(...) UNSTRUCTURED_GRID 0')
+      ASSERT(bool,'testVTUFile%hasMesh')
+      INQUIRE(FILE='testPVTU_0.vtu',EXIST=bool)
```
bkochuna: Be careful here. You should make sure the test deletes this file upon completion. Otherwise multiple executions of the test can lead to fall positives.

PR #23, opened 2017-05-19
Sundials

bscollin:
Unit tests pass with and without sundials enabled.

In order for this to work with sundials:
* Download cvode:  https://computation.llnl.gov/projects/sundials/download/cvode-2.9.0.tar.gz
* Unzip tarball
* Create build directory
* Change into build directory and run cmake:  cmake -DFCMIX_ENABLE=ON -DCMAKE_INSTALL_PREFIX={install path}  {source path}
* make -jN install
* Then when building MPACT or Futility (or VERA or Mongoose) you need to add -DTPL_ENABLE_SUNDIALS=ON -DSUNDIALS_LIBRARY_DIRS={install path}


bscollin:
I've done Futility, MPACT, and Mongoose builds with and without Sundials and Futility unit tests pass.  Also had a mixture of debug and release buidls.

src/ODESolverTypes.f90
```diff
@@ -189,6 +220,146 @@ SUBROUTINE odesolver_sub_absintfc(solver)
...
+        IF(n < 1) THEN
+          CALL eODESolverType%raiseError('Incorrect input to '// &
+            modName//'::'//myName//' - Number of values (n) must be '// &
+              'greater than 0!')
+        ELSE
+          solver%n=n
+        ENDIF
+
+        IF(solvetype == THETA_METHOD) THEN
+          !Error... unsupported option
```
stimpsonsg: Do we want to fill in the error here?

src/ODESolverTypes.f90
```diff
@@ -189,6 +220,146 @@ SUBROUTINE odesolver_sub_absintfc(solver)
 !> @param solver The ode solver to act on
 !> @param Params A parameter list with input options
 !>
```
stimpsonsg: If we want to have all the doxygen data up to date, should add f.

src/ODESolverTypes.f90
```diff
@@ -189,6 +220,146 @@ SUBROUTINE odesolver_sub_absintfc(solver)
...
+      CALL FCVODE(REAL(TF,C_DOUBLE),REAL(ttmp,C_DOUBLE),solver%ytmp,INT(1,C_INT), ierr)
+      CALL yf%set(solver%ytmp)
+#endif
+    ENDSUBROUTINE step_ODESolverType_Sundials
+!
+!-------------------------------------------------------------------------------
+!> @brief Initializes the ODE Solver Type with a parameter list
+!>
+!> @param solver The ode solver to act on
+!> @param Params A parameter list with input options
```
stimpsonsg: Same about f here

src/ODESolverTypes.f90
bscollin: addressed

src/ODESolverTypes.f90
bscollin: addressed

src/ODESolverTypes.f90
bscollin: addressed

bscollin:
Addressed Shane's comments.  Merging with master.

PR #27, opened 2017-06-22
SuperLU Linear Solver

stimpsonsg:
Adding hooks to direct LS type to use SuperLU.  Right now, this is enabled by specifying PETSC as the TPL and LU as the solver method.


bscollin:
Changes look reasonable but a few comments:
 1.  Would be really nice if at configure time you could determine which TPLs PETSC is built with instead of letting PETSc error out
 2.  There is not test of our wrappers in linear solver type.  That should probably be added
That's all I got for now.

stimpsonsg:
Alright, I've added a small test to testLS to account for our hooks in LinearSolverTypes.

PR #28, opened 2017-06-28
Make isNaN and isInf functions elemental.

benyee:
Without this, the functions do not work for array inputs (they can evaluate to
false even if there are nan's present).  There are already places in MPACT
where isNaN and isInf are assumed to be elemental.

### DIFF
--- a/src/IntrType.f90
+++ b/src/IntrType.f90
@@ -330,28 +330,28 @@ MODULE IntrType
 
   !> Definition of external C interfaces defined in CUtils.
   INTERFACE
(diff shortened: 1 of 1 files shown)


benyee:
I compiled with gcc-4.8.3 and there weren't any issues.  I'll try it with an intel compiler and let you know if that works.

benyee:
I just tried compiling with icc-13.0.1 and it seems to work there (built it, compiled it, and checked that it output the correct values for a 2D array input).  Is there anything else I should check?

PR #31, opened 2017-06-30
Making two builds, debug Futility only and mpi release futility + tri…

bscollin:
…linos + all TPLs


bscollin:
ran for 25 minutes... not too bad considering the frequency of a pull request.

bscollin:
Agree on prebuilt libraries of trilinos.  I guess then we would create our own docker container that has trilinos installed?

youngmit:
Exactly

PR #36, opened 2017-07-12
Update sundials ode solver to support multiple ODE solver objects

bscollin:
The sundials fortran interface effectively defines a singleton object
for the ode solver.  Since multiple ODE solvers are needed in Mongoose
in order to solve multiple pins at once.  This requires a lot of sundials
data to be global across all ODE solver types so module data structures are
used to hold a few logicals and data arrays that sundials requires.  Note that
this fix will only work if all ODE solvers have the same number of unknowns.
A better and more permanent fix would be to use the PETSc wrapper for
sundials or directly wrap the C interface for sundials if multiple size ODE solvers
are needed.


youngmit:
This all looks good. It would be nice to store all of the Sundials global state in an object to keep from cluttering the "namespace," though that would be going to a lot of trouble for code that will probably get replaced by something more robust later.

PR #37, opened 2017-07-14
Adds functionality to support writing and reading attributes to HDF5 …

HendersonSC:
…objects.

Description:
  Adds routines to the HDF5 file interface to read and write attribute
  name/value pairs to an object. For the read routines the attribute name must
  be known from the beginning.

  Adds test components to the unit tests for testing the read and write
  attriubute functions.

### DIFF
--- a/src/FileType_HDF5.f90
+++ b/src/FileType_HDF5.f90
@@ -474,16 +474,16 @@ MODULE FileType_HDF5
         write_attribute_d0
       !> @copybrief FileType_HDF5::read_str_attribure_help
       !> @copyoc FileType_HDF5_read_str_attribure_help
-      PROCEDURE,PASS,PRIVATE :: read_attribute_st0_helper 
-      !> @copybrief FileType_HDF5::read_attribure_i0
-      !> @copyoc FileType_HDF5_read_attribure_i0
-      PROCEDURE,PASS,PRIVATE :: read_attribure_i0 
+      PROCEDURE,PASS,PRIVATE :: read_attribute_st0 
+      !> @copybrief FileType_HDF5::read_attribute_i0
+      !> @copyoc FileType_HDF5_read_attribute_i0
+      PROCEDURE,PASS,PRIVATE :: read_attribute_i0 
       !> @copybrief FileType_HDF5::read_attribute_d0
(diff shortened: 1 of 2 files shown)


src/FileType_HDF5.f90
```diff
@@ -6571,5 +6595,338 @@ SUBROUTINE compute_chunk_size(gdims,cdims)
...
+       CHARACTER(LEN=*),INTENT(IN) :: obj_name, attr_name
+       REAL(SDK),INTENT(INOUT) :: attr_val
+        
+#ifdef FUTILITY_HAVE_HDF5
+       INTEGER(HID_T) :: attr_id, dset_id
+       INTEGER(HSIZE_T),DIMENSION(1) :: dims
+       dims(1)=1
+
+       !Prepare the File and object for the attribute
+       CALL open_object(this,obj_name,dset_id)
```
youngmit: This is pretty nit-picky, but if you are going to open_object explicitly outside of the pre_read_attribute routine, it looks scary not to see a matching close_object call. I would recommend calling open_object inside the pre_read_attribute, so that it's more clear that things are getting cleaned up right.

src/FileType_HDF5.f90
```diff
@@ -6571,5 +6595,338 @@ SUBROUTINE compute_chunk_size(gdims,cdims)
...
+    END SUBROUTINE read_attribute_d0
+!
+!-------------------------------------------------------------------------------
+!> @brief Sets up all attribute operations by checking links and opening object` 
+!> 
+!> @param obj_name the relative path to the dataset
+!> @param dset_id the HDF5 system id for the working dataset 
+!>  
+#ifdef FUTILITY_HAVE_HDF5
+    SUBROUTINE open_object(this,obj_name,dset_id)
```
youngmit: Also super nit-picky. Change dset_id to obj_id throughout, where relevant, for consistency.

HendersonSC: Made this edit as written.

src/FileType_HDF5.f90
HendersonSC: Changed the function call and structured the subroutine to capture the intent of this edit.

PR #38, opened 2017-07-14
Restructure Matrix and Vector types

youngmit:
This does a number of things:

- Split implementations of the MatrixTypes and VectorTypes into separate files. The base type and common code lives in the _Base.f90 files. Native, Trilinos and PETSc implementations live in their own files/modules. The MatrixTypes.f90 and VectorTypes.f90 "main modules" USE and expose as PUBLIC all of the relevant types and free functions from the Base and derived modules. In most cases (all cases at present), client code simply USEs the MatrixTypes and/or VectorTypes modules.
- The main modules provide abstract factory routines for allocating new objects using the 'engine' parameter on the passed parameter list. There are also routines for generating matrices and vectors of the same type as those passed in.
- Code that uses the matrix and vector types has been refactored to use these factories so that they interact with them in a more type-agnostic way. This largely eliminates the SELECT TYPEs and \#ifdef guards that used to be all over the place. Just as importantly, this should make it much easier to add new linear algebra engines in the future (read: ForTrilinos)


src/MatrixTypes.f90
```diff
@@ -32,6 +32,7 @@
 !>  - @ref BLAS2 "BLAS2": @copybrief BLAS2
 !>  - @ref BLAS3 "BLAS3": @copybrief BLAS3
 !>
+!> TIBWSFB: This documentation doesnt match the current interface
```
bscollin: I like it but didn't you just change this entire interface... doesn't that mean you should have updated the documentation ;-)

src/MatrixTypes.f90
```diff
@@ -64,337 +65,49 @@ MODULE MatrixTypes
...
+  PUBLIC :: MatrixFactory
+  PUBLIC :: DistributedMatrixFactory
+  PUBLIC :: MatrixResemble
   PUBLIC :: SquareMatrixType
   PUBLIC :: RectMatrixType
+  PUBLIC :: DistributedMatrixType
+  ! Matrix structure enumerations
+  PUBLIC :: SPARSE,DENSESQUARE,DENSERECT,TRIDIAG
+  ! Matrix-Vector engine enumerations
+  PUBLIC :: VM_PETSC,VM_TRILINOS,VM_NATIVE
```
bscollin: what does VM stand for?

src/MatrixTypes_PETSc.f90
```diff
@@ -0,0 +1,533 @@
+!
```
bscollin: Missed the top line in the header... its right in native

src/VectorTypes.f90
```diff
@@ -20,6 +23,7 @@
 !>  - @ref ExceptionHandler "ExceptionHandler": @copybrief ExceptionHandler
 !>  - @ref Allocs "Allocs": @copybrief Allocs
 !>
+!> TIBWSFB: This example doesnt match the current interface
```
bscollin: Again, shouldn't we just fix this now?

src/VectorTypes.f90
```diff
@@ -1563,6 +414,10 @@ SUBROUTINE axpy_scalar_VectorType(thisVector,newVector,a,n,incx,incy)
 !> @param incx the increment to use when looping over elements in @c x
 !> @param incy the increment to use when looping over elements in @c y
 !>
+!> MAINTENANCE NOTE: This is inefficient. It makes a copy of the vectors, which
```
bscollin: can you make an issue pointing to this note?

bscollin:
Also note that @stimpsonsg  updated MPACT which might require updating based on this refactor.

src/MatrixTypes.f90
youngmit: Vector-Matrix

src/MatrixTypes.f90
youngmit: hahaha don't look at me; that was broken when we went from passing an int for size to using the parameter lists. I'm still angling for init routines that use a struct for passing valid params, rather than a full-fledged parameter list, so I'm not itching to re-write the documentation twice.

src/MatrixTypes_PETSc.f90
youngmit: Will fix

src/VectorTypes.f90
youngmit: I used MAINTENANCE NOTE before using TIBWSFB. If you think its a big enough deal, I'll make an issue.

youngmit:
Is Shane's stuff on master now? I can pull that down and make sure everything is good. Also, you're right, I should add a unit test for the factory and Resemble routines. Should be pretty simple; just call them for each of the types, select type to make sure they come out the right type and are initialized.

stimpsonsg:
Yeah, my stuff is on master now.

bkochuna:
I'd like to have a look at this before it gets merged to get myself up to speed. I can look at it over lunch.

src/MatrixTypes_Base.f90
```diff
@@ -0,0 +1,277 @@
...
+  TYPE,ABSTRACT,EXTENDS(SquareMatrixType) :: DistributedMatrixType
+
+    !> creation status
+    LOGICAL(SBK) :: isCreated=.FALSE.
+    !> assembly status
+    LOGICAL(SBK) :: isAssembled=.FALSE.
+    !> MPI comm ID
+    INTEGER(SIK) :: comm=-1
+    !> number of local values
+    INTEGER(SIK) :: nlocal
```
bkochuna: This needs a default value.

src/EigenvalueSolverTypes.f90
```diff
@@ -253,6 +253,52 @@ SUBROUTINE evsolver_sub_absintfc(solver)
...
+      TYPE(ParamType),INTENT(IN) :: params
+      
+      IF(ASSOCIATED(solver)) THEN
+        CALL eEigenvalueSolverType%raiseError(modName//"::"//myName//" - "// &
+          "Solver pointer is already associated")
+        RETURN
+      ENDIF
+
+      SELECTTYPE(M)
+#ifdef FUTILITY_HAVE_PETSC
```
bkochuna: What was the reasoning for ifdefing the TPL provided types? (thinking about this from the point of view of minimizing number places with ifdefs)

I guess you get build errors instead of run time errors if you still provide stub types. That seems like an advantage. Also, I guess you should only need the ifdefs in the factories and modules containing the concrete implementations.

src/VectorTypes_Native.f90
```diff
@@ -6,31 +6,7 @@
...
-!> PROGRAM ExampleVector
-!>   TYPE(RealVectorType) :: vector
-!>
-!>   CALL vector%init(36)
-!>   CALL vector%set(1,10._SRK)
-!>   value=vector%get(1)
-!>   CALL vector%clear()
-!> ENDPROGRAM ExampleVector
-!> @endcode
+!> @brief Futility-native implementations of VectorType
```
bkochuna: Was there a reason you did not keep the example code and dependencies?

youngmit: The example code is in the main modules, since those are the ones that should be user facing anyways. Just didn't get around to updating the module deps after the dust settled. Fixing it now

src/EigenvalueSolverTypes.f90
youngmit: Ifdefing out the types themselves cuts down on the number of ifdefs needed throughout, lets us get rid of all of the errors that are getting thrown way too late in the game anyways, and with the addition of an abstract factory, the types shouldn't be needed in pretty much any client code to begin with. If there is a need for client code to interact with the TPL-backed types directly, they should use their own ifdef guards.

Make sense/seem reasonable?

src/MatrixTypes.f90
youngmit: To clarify, I was worried about cluttering the namespace with overly-general NATIVE, TRILINOS, PETSC enumerators.

src/MatrixTypes_Base.f90
youngmit: Done

src/VectorTypes.f90
youngmit: I'm writing new examples, but I kind of got stuck holding the bag on this; these should have been updated when the original %init routines were changed to take parameter lists instead of a list of arguments. Also, for performance reasons), I don't think we should be using parameter lists anyways  (e.g. where frequently constructing matrices is needed, as in the previous iteration of the Sundials interface. Parameter lists are dreadfully slow, and doing plist validation is especially so. I still advocate for a simple derived type for storing arguments, where each optional argument is initialized with a sane default. Discussion for another day.

src/MatrixTypes_PETSc.f90
bscollin: Missed the "!" at the end of line 2

bscollin:
@bkochuna , I'm fine with all the changes but your review is blocking the ability to merge.  Any chance you can take another look.  Looks like Mitch has addressed your concerns.

bkochuna:
Yeah. I'm happy with the changes. Thanks for taking the time to do this Mitch, the new design is much more maintainable.

PR #43, opened 2017-07-21
Unit test

HendersonSC:
Add new ASSERT macros to unit_test.h to support outputting test and reference values on failure. Add tests for the new macros.

tested locally using HDF attribute read/write functions during setup. Eliminated a few unused variables in the process.


bscollin:
Requesting assert_(gt,ge,lt,le,approxge,approxle)

HendersonSC:
Included the requested macros, and changed the output of from the macros with failing conditions. The 132(?) line length limit makes it difficult to both have information and be consistent with the output. I selected the following format [ val1 /= val2 ] and [ val1 Not < val2 ], where appropriate. Some not all could have been more verbose.

PR #44, opened 2017-07-21
Adding routine to IOutil for mkdir

bkochuna:
Description:
This implements a fairly portable make directory interface for
Fortran. Works for GNU and Intel.

Should work in Windows, but have not tested it there yet.

### DIFF
--- a/src/IOutil.f90
+++ b/src/IOutil.f90
@@ -57,6 +57,7 @@ MODULE IOutil
   PUBLIC :: GET_COMMAND
   PUBLIC :: GET_ENVIRONMENT_VARIABLE
   PUBLIC :: GET_CURRENT_DIRECTORY
+  PUBLIC :: MAKE_DIRECTORY
 
   !Needed for getPWD_c
   INCLUDE 'getSysProcInfo_F.h'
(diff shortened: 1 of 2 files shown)


zdodson94:
I had to change the syntax of the call to SYSTEM because it was causing segfaults.

bkochuna:
I don't think the "CALL" to system works for the Intel compiler. I believe the Intel compiler has only implemented this routine as a function.

zdodson94:
Yes, I saw that and made the fix in the vtu_org branch, which I merged mkdir into. Everything seems to work now on that branch.

src/IOutil.f90
```diff
@@ -244,4 +251,30 @@ SUBROUTINE GET_CURRENT_DIRECTORY_internal(dir,length,status)
...
 !
+!-------------------------------------------------------------------------------
+    SUBROUTINE MAKE_DIRECTORY_string(dir,status)
+      TYPE(StringType),INTENT(IN) :: dir
+      INTEGER(SIK),INTENT(OUT),OPTIONAL :: status
+      CALL MAKE_DIRECTORY_char(CHAR(dir),status)
+    ENDSUBROUTINE MAKE_DIRECTORY_string
+!
+!-------------------------------------------------------------------------------
+    SUBROUTINE MAKE_DIRECTORY_char(dir,status)
```
zdodson94: Note that "status" must be passed into this routine to avoid a segfault. Maybe it should not be optional?

youngmit: Just protect with `if(present(status))`. Bear in mind that present-ness gets forwarded to any subsequent calls that take optional arguments, so you only need to check `present` immediately before accessing an optional variable.

PR #45, opened 2017-07-26
Adding PETSc matrix transpose testMatrixTypes

bkochuna:
Want to see if this passes the CI which uses the VERA TPLs docker container.

### DIFF
--- a/src/MatrixTypes_PETSc.f90
+++ b/src/MatrixTypes_PETSc.f90
@@ -263,6 +263,9 @@ MODULE MatrixTypes_PETSc
       CLASS(PETScMatrixType),INTENT(INOUT) :: matrix
       PetscErrorCode  :: iperr
       IF(.NOT.matrix%isAssembled) CALL matrix%assemble()
+      !This is to avoid a deadlock in IBarrier in MPICH
+      CALL PetscCommBuildTwoSidedSetType(matrix%comm, &
+        PETSC_BUILDTWOSIDED_ALLREDUCE,iperr)
       CALL MatTranspose(matrix%a,MAT_REUSE_MATRIX,matrix%a,iperr)
     ENDSUBROUTINE transpose_PETScMatrixType
 !
--- a/unit_tests/testMatrixTypes/testMatrixTypes.f90
+++ b/unit_tests/testMatrixTypes/testMatrixTypes.f90
@@ -3526,10 +3526,12 @@ PROGRAM testMatrixTypes
 !-------------------------------------------------------------------------------
     SUBROUTINE testTransposeMatrix()
       CLASS(SparseMatrixType),ALLOCATABLE :: testA
-      CLASS(DistributedMatrixType),POINTER :: dmat_p
       LOGICAL(SBK) :: bool
-      REAL(SRK) :: aij
       TYPE(ParamType) :: tmpPlist
+#ifdef FUTILITY_HAVE_PETSC
+      CLASS(DistributedMatrixType),POINTER :: dmat_p
+      REAL(SRK) :: aij
+#endif
       ALLOCATE(SparseMatrixType :: testA)
 
 
--- a/unit_tests/testMatrixTypes/testMatrixTypes.f90
+++ b/unit_tests/testMatrixTypes/testMatrixTypes.f90
@@ -3526,11 +3526,14 @@ PROGRAM testMatrixTypes
 !-------------------------------------------------------------------------------
     SUBROUTINE testTransposeMatrix()
       CLASS(SparseMatrixType),ALLOCATABLE :: testA
+      CLASS(DistributedMatrixType),POINTER :: dmat_p
       LOGICAL(SBK) :: bool
+      REAL(SRK) :: aij
       TYPE(ParamType) :: tmpPlist
       ALLOCATE(SparseMatrixType :: testA)
 
 
+      COMPONENT_TEST('Sparse')
       CALL Plist%clear()
       CALL tmpPlist%add('MatrixType->n',4_SIK)
       CALL tmpPlist%add('MatrixType->matType',SPARSE)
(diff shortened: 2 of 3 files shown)


bkochuna:
Ok, looks like the test MatrixTypes hung on this run of the CI

youngmit:
Nuggets... can you replicate this locally?

bkochuna:
Doing that now (I'm guessing yes).

bkochuna:
I am able to reproduce myself within the docker container. First thing I checked was to make sure we were calling the interface correctly, and as far as I can tell we are.
In Futility the PETSc transpose routine is:
```
    SUBROUTINE transpose_PETScMatrixType(matrix)
      CHARACTER(LEN=*),PARAMETER :: myName='transpose_PETScMatrixType'
      CLASS(PETScMatrixType),INTENT(INOUT) :: matrix
      PetscErrorCode  :: iperr
      IF(.NOT.matrix%isAssembled) CALL matrix%assemble()
      CALL MatTranspose(matrix%a,MAT_REUSE_MATRIX,matrix%a,iperr)
    ENDSUBROUTINE transpose_PETScMatrixType
```
And in the file `$PETSC_DIR/include/finclude/ftn-auto/petscmat.h90` the interface is defined as:
```
      subroutine MatTranspose(a,b,c,z)
       Mat a ! Mat
       MatReuse b ! MatReuse
       Mat c ! Mat
       integer z
       end subroutine
```

I'll also note that in our regular dev env that is not in docker, we use the same PETSc distribution (petsc-3.5.4)

So I'm pretty sure we are calling the interface correctly. Not sure what else the issue might be...

youngmit:
Maybe try running it in debug under gdb. wait for it to hang, do a ctrl-c to send SIGINT. Then you can do `bt` to figure out where its stuck.

bkochuna:
A few follow on comments:

Earlier when Damien was putting this image together, we had a BUNCH of failing tests initially, which we traced to the HDF5 library not getting built correctly. I'm wondering if there could also be an issue with the PETSc library build.

I'm also testing now with debug and valgrind. I'll try the gdb thing shortly. Thanks for the suggestion

bkochuna:
I'll also probably update the testTPLPETSc to include a call to MatTranspose before I merge this

bkochuna:
Backtrace from GDB is:
```
#0  0x00007ffff467270d in poll () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007ffff5731c0e in MPID_nem_tcp_connpoll () from /usr/lib/x86_64-linux-gnu/libmpich.so.12
#2  0x00007ffff571b347 in MPIDI_CH3I_Progress () from /usr/lib/x86_64-linux-gnu/libmpich.so.12
#3  0x00007ffff570af31 in MPID_Iprobe () from /usr/lib/x86_64-linux-gnu/libmpich.so.12
#4  0x00007ffff56776a5 in PMPI_Iprobe () from /usr/lib/x86_64-linux-gnu/libmpich.so.12
#5  0x00007ffff60f1184 in PetscCommBuildTwoSided ()
   from /tools/vera/gcc-5.4.0/tpls/opt/petsc-3.5.4/lib/libpetsc.so.3.5
#6  0x00007ffff625ebb6 in PetscSFSetUp_Basic () from /tools/vera/gcc-5.4.0/tpls/opt/petsc-3.5.4/lib/libpetsc.so.3.5
#7  0x00007ffff6262f1c in PetscSFSetUp () from /tools/vera/gcc-5.4.0/tpls/opt/petsc-3.5.4/lib/libpetsc.so.3.5
#8  0x00007ffff62679d9 in PetscSFReduceBegin () from /tools/vera/gcc-5.4.0/tpls/opt/petsc-3.5.4/lib/libpetsc.so.3.5
#9  0x00007ffff66f35e8 in MatTranspose_MPIAIJ () from /tools/vera/gcc-5.4.0/tpls/opt/petsc-3.5.4/lib/libpetsc.so.3.5
#10 0x00007ffff637ef33 in MatTranspose () from /tools/vera/gcc-5.4.0/tpls/opt/petsc-3.5.4/lib/libpetsc.so.3.5
#11 0x00007ffff63b4f7e in mattranspose_ () from /tools/vera/gcc-5.4.0/tpls/opt/petsc-3.5.4/lib/libpetsc.so.3.5
#12 0x0000000000541cbc in matrixtypes_petsc::transpose_petscmatrixtype (matrix=...)
    at /home/bkochuna/MPACT/Futility/src/MatrixTypes_PETSc.f90:266
#13 0x0000000000410377 in testtransposematrix ()
```
So its pretty deep into the call stack... not what to do about that. Thoughts?

bscollin:
Is the version of MPI the same?

bkochuna:
The MPI version is 3.2-6build1. Its a from the package manager.

bscollin:
The BT seems to suggest its hanging in MPI.  In theory it could be some weird interaction with docker or a different version of mpich in general.  @youngmit, don't you have this same compiler stack on yupana?

youngmit:
I should have, yeah. Its not hanging on yupana, though

bkochuna:
I'm working on setting up a different Ubuntu docker container to test if this was a problem related to how the TPL was installed (since Mitch doesn't have this issue on his software stack).

bkochuna:
Well I run into the same issue in the other docker container I created.

Mitch, is it possible for you to provide some info about how you built PETSc on yupana?

bkochuna:
Hmm... so I also ran into this issue in the Centos 7 image that uses different compilers.

I'm wondering if this might be a PETSc + docker problem that happens because of the virtualization?

bkochuna:
Positive update! I emailed petsc-maint about this issue and there is a known problem with Ibarrier in MPICH. They suggested adding the command line option ` -build_twosided allreduce`.

Turns out if you run the unit tests with these options they do not hang.

So now how to figure out how to work around this issue... some options might be:

1. Do not use the mpich provided by the package manager.
2. Figure out if there is a PETSc configure option to make this a default.
3. Figure out if PETSc has a mechanism to specify this through the API.

I'll work on investigating these things.

bkochuna:
Think I've got it through the API.

bscollin:
Should we open a ticket on this and start discussing with infrastructure?

bkochuna:
@bscollin I have a ticket open, not sure if its an INF ticket though. I'll update it tomorrow and cc Sieger and Damien etc.

clarnokt:
Fantastic - thanks!

bkochuna:
@bscollin the PETSc people originally gave me a command line option to try with the executable, so that's probably what you were thinking of.

After searching the PETSc documentation I found the right calls through their API.

Anyway all the transient cases should work now under Docker. I'll follow up in the kanban tickets about re-enabling those tests for Docker.

p.s. might have found a workaround for the DA32 unit test =O

PR #46, opened 2017-07-27
Removes compilier warnings from unused variables

HendersonSC:
Removes unused variables from recent changes.


bscollin:
This test should fail until everything is clear

PR #47, opened 2017-08-03
Hotfix da32

bkochuna:
In testing stuff for the PETSc issue I also looked at this.

Looks like my original code works in the docker container and all the problems were for previous versions of GCC.

Original code for GCC 5.4.0 is a-ok. So, just modified an ifdef here. And the travis ci to re-enable this test (and remove dos line endings)


bkochuna:
Nuts. The CI for this branch didn't use the {{{.travis.yml}}} file on the branch. Will have to confirm manually. I think I actually fixed this a better way anyway which would have gotten to the underlying issue.

.travis.yml
```diff
@@ -1,19 +1,19 @@
...
+
+services:
+ - docker
+
+before_install:
+ - echo $PWD
+ - docker pull dalg24/vera_tpls
+
+script:
+ - docker run -v $PWD:/mnt/Futility dalg24/vera_tpls /bin/bash -c "mkdir /build; cd /build; cmake /mnt/Futility; make -j2 |& tee /mnt/Futility/make.out; ctest -E DA32 -j2"
```
bscollin: Did you miss this one?

Is like 19 all that changed here?

bkochuna: I did miss one. The reason the whole file looks changed is the dos2unix. I did only remove the `-E DA32` option from (now) both instances in the .travis.yml

If you click the individual commits here it'll show you what changed more specifically (in the one commit) rather than just diffed with master.

bkochuna:
Cool the test did run on this branch. I probably just didn't think it had before because I only changed the one line in the `.travis.yml` file.

PR #48, opened 2017-08-14
Refactor to use Tpetra matrix and vector types, as well as solver packages that use them

youngmit:
This replaces the Epetra matrix and vector classes with those provided by Tpetra. The Anasazi code has been refactored to use these instead. The old preconditioner packages (Ifpack, ML) needed to be replaced with Ifpack2 and MueLu instead. Non-linear solvers (NOX/LOCA and Anderson) have been disabled, since they too will require refactoring, but aren't in use in the wild.


youngmit:
Two changes:
 - Added Anderson acceleration and JFNK stuff back in. It still uses the Epetra interfaces, since the amount of added effort needed to use the tpetra/thyra interface is impractical for now. When we start using this more, we can find someone that works with the thyra model evaluator stuff to help us implement it properly
 - Added ML to the list of disabled packages. It was the only one we can do without. Once the Anderson stuff gets moved to thyra, the Ifpack dependency can go, too.

bscollin:
Something really weird is happening with the docker when building Trillinos.

PR #49, opened 2017-08-14
Partition graph

apfitzge:
This branch contains changes which add a new type `PartitionGraphType`. This type can be used to partition a graph in an arbitrary number of groups (domains).

Additionally, the graph may have some spatiality to it. This is beneficial in cases in which the graph represents a mesh.  In general this spatial dependence may be n-dimensional, although only 2D or 3D cases are expected.


bkochuna:
Andrew, did you look at Geom_Graph.f90 to see if there was any overlapping design?

apfitzge:
Brendan,

I had looked at `Geom_Graph.f90` when I first designed the type, but there doesn't seem to be a lot of overlap.

Some major differences.
1. GraphType requires coordinates, `PartitionGraphType` has optional coordinates
2. `GraphType%quadEdges` makes no sense for `PartitionGraphType`
3. `GraphType%edgeMatrix` is a full edgeMatrix, which would not be efficient for Partitioning when we expect some upper bound (low relative to number of vertices) for the maximum number of neighbors.
4. `GraphType` orders it's vertices in clockwise manner, no need to do this in partitioning.

Additionally nearly all the routines on GraphType would not benefit the PartitionGraph, there is no need to order the vertices in some clockwise (spatial) manner, or get MCB, etc. Likewise, using any of the partitioning algorithms on the geometric graph also doesn't really make sense.

In summary, the types are different enough to justify not sharing a common base type.

bkochuna:
Thanks Andrew! All good points, and I agree. I just couldn't recall if you'd looked at that before.

PR #51, opened 2017-09-01
Added new MPI reduce wrapper 'trueForAny' to ParallelEnv which checks…

zdodson94:
… if a boolean is true on any rank.


bkochuna:
Looks great Zack! Nice and clean. As soon as the CI server says the its ok to merge, I'll merge it in!

PR #52, opened 2017-09-11
Common config

bscollin:
Making Futility and MPACT configurations the same.  Fixed an error in calling PROJECT() in an include file instead of the top level CMakeLists.txt file.  Merged latest version of TriBITS in.  This is to close out CASL ticket 4901.


bscollin:
Note that now configuration starts at line 300 but checking for compilers happens at 415 after processing all of the folders which is consistent with the rest of CASL.

bscollin:
Requesting squash-merge on this.

CMakeLists.txt
```diff
@@ -9,13 +9,11 @@
...
-  IF(DEFINED ${PROJECT_NAME}_ENABLE_Teuchos)
-    IF("${${PROJECT_NAME}_ENABLE_Teuchos}" STREQUAL "OFF")
-      SET(${PROJECT_NAME}_ENABLE_ML OFF)
-      SET(${PROJECT_NAME}_ENABLE_Triutils OFF)
-    ENDIF()
-  ENDIF()
+
+  # CMake requires that you declare the CMake project in the top-level file and
+  # not in an include file :-(
+  PROJECT(${PROJECT_NAME} NONE)
```
bkochuna: Why is the project name NONE instead of Futility?

cmake/FutilityBuildQuickRefTemplate.rst
```diff
@@ -0,0 +1,48 @@
...
+
+The set of Futility extra repos is officially defined in the file:
+
+  `Futility/cmake/ExtraRepositoriesList.cmake <cmake/ExtraRepositoriesList.cmake>`_
+
+Configurations of Futility tend to keep complex configuration inside of CMake
+options fragment files and are read in using the TriBITS variable
+``Futility_CONFIGURE_OPTIONS_FILE`` (see `Basic configuration`_ for details).
+Trying to set up a basic configuration for Futility from scratch is very
+difficult.  Thankfully, there are standard configuration for Futility located
```
bkochuna: configurations <- configuration

cmake/FutilityBuildQuickRefTemplate.rst
```diff
@@ -0,0 +1,48 @@
...
+=========================
+
+Below, configure options specific to Futility are given.  The later sections
+give more generic options that are the same for all TriBITS projects.
+
+First note that the location of ``TRIBITS_BASE_DIR`` in Futility is::
+
+  TRIBITS_BASE_DIR=Futility/cmake/tribits
+
+ToDo: Fill in Futility-specific options.
```
bkochuna: Would like to see this be a little more polished before merging. If you're going to create some documentation we should try to have it be complete.

bscollin: This is exactly what has been in MPACT for several years.  I only put it in because you asked me to put in something similar to what was in mpact.

CMakeLists.txt
bscollin: The project name gets processed in TriBITS.  Ross says this is the way you have to do it for cmake to be happy.  Just following TriBITS documentation.... actually... the comment is from the documentation too.

cmake/FutilityBuildQuickRefTemplate.rst
bscollin: honestly.. I don't get the point of this BuildQuickRefTemplate file.  I can't find this in any of the TriBITS or CASL projects other than MPACT.  I'm inclined to say we should just get rid of this file in both Futility and MPACT.  We probably do need to write some build instructions but I feel like its outside of this merge and the underlying ticket.

bkochuna: I mean I think the purpose of the quick build ref is just like "hey here are these 5 commands to get you running". It isn't necessary that this needs to be an rst, html, and pdf. It could just be like a README or the first section in the BuildReference.

Either way, I agree addressing the larger issue of updating the installation/build reference documentation deserves its own task. Do you mind making a PHI Kanban ticket for this?

In the mean time I'll squishy merge this.

bkochuna:
Forgot to add that it looks like TriBITS eliminated the build_quick_ref. Looking through the TriBITS commit logs Ross has:
```
commit a51b98d560fb3152b72961e86226fbe5199c59f2
Author: Roscoe A. Bartlett <bartlettra@ornl.gov>
Date:   Fri May 8 09:10:24 2015 -0400

    Finish name change from "BuildQuickRef" to "BuildReference" (TriBITS #68)

commit 64c3b3da1486f07e396b9954c855caacb9fe850d
Author: Roscoe A. Bartlett <bartlettra@ornl.gov>
Date:   Fri May 8 08:59:04 2015 -0400

    Start name change from "BuildQuickRef" to "BuildReference" (TriBITS #68)

    This document is far to long to be a "quick reference".  This document could
    contain a project-specific quick reference but the entire document is too big
    for that.

    This just changes the names of the files so that git can follow the renames.
```
And the page for this issue is https://github.com/TriBITSPub/TriBITS/issues/68

cmake/FutilityBuildQuickRefTemplate.rst
bscollin: makes sense.  I'd rather see this in a better place than buried in the cmake folder (because as a general rule... who looks there?)

PR #53, opened 2017-09-16
Dbc

bscollin:


unit_tests/testDBC/testDBC.f90
```diff
@@ -0,0 +1,81 @@
...
+!> @brief Tests the REQUIRE macro and associated function calls
+!>
+    SUBROUTINE require_pass()
+      WRITE(*,*) "Testing REQUIRE Passing"
+      DBC_REQUIRE(5==5)
+
+    ENDSUBROUTINE require_pass
+!
+!-------------------------------------------------------------------------------
+!> @brief Tests the ENSURE macro and associated function calls
```
bkochuna: Commenting unit tests!? What wizardry is this?

src/DBC.f90
```diff
@@ -0,0 +1,62 @@
...
+!                          Futility Development Group                          !
+!                             All rights reserved.                             !
+!                                                                              !
+! Futility is a jointly-maintained, open-source project between the University !
+! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
+! can be found in LICENSE.txt in the head directory of this repository.        !
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief description
+!>
+!>
```
bkochuna: Can you fill in this header?

src/DBC.f90
```diff
@@ -0,0 +1,62 @@
...
+!> @par
+!> @code
+!> @endcode
+!>
+!>
+!> @author Benjamin Collins
+!>    @date 11/17/2012
```
bkochuna: change date

cmake/tribits/dashboard_driver/TribitsDriverDashboard.cmake
```diff
@@ -253,7 +253,7 @@ if (NOT TDD_IN_TESTING_MODE)
     SET(PULL_OUT_FILE "${CTEST_BINARY_DIRECTORY}/${EXTRA_PULL_DIR}.pull.out")
     set(CTEST_NOTES_FILES ${CTEST_NOTES_FILES} ${PULL_OUT_FILE})
     MESSAGE("Pull extra updates in '${EXTRA_PULL_DIR_ABS}' ...")
-    EXTRAREPO_CLEAN_FETCH_RESET("${git_exe}" "${EXTRA_PULL_DIR_ABS}")
+    TRIBITS_UPDATE_GIT_EXTRAREPO("${git_exe}" "${EXTRA_PULL_DIR_ABS}")
```
bkochuna: These look like what was in the common config branch? Is this just replicating them?

cmake/Futility_Configurations.cmake
```diff
@@ -269,7 +274,7 @@ ELSEIF(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
        )
 
     SET(Fortran_FLAGS_RELEASE
-        ${CSYM}Ofast
+        ${CSYM}O3
```
bkochuna: I recall at the stand-up Mitch thought there was one option within Ofast causing problems with ORIGEN. Could we try including more options from Ofast and add to O3 that don't cause problems?

(This could be done on separate branch)

src/DBC.f90
bscollin: Time travel... I figured it out

cmake/tribits/dashboard_driver/TribitsDriverDashboard.cmake
bscollin: I snapshotted tribits in one of them... maybe both... I'll take a look.

cmake/Futility_Configurations.cmake
bscollin: you're right.  We could move this into a different branch.  I'll change it back.

youngmit: Fine if we move this to a separate branch, but we probably shouldn't use anything from Ofast. From the gcc man pages:
>  Disregard strict standards compliance.  -Ofast enables all -O3
>  optimizations.  It also enables optimizations that are not valid
>  for all standard-compliant programs.

The only optimization that it enables that I would consider safe is `-fstack-arrays`, but even that has caused us stack overflow issues in the past, and might be considered harmful.

src/DBC.f90
bscollin: Fixed

src/DBC.f90
bscollin: Fixed... with code sample :)

unit_tests/testDBC/testDBC.f90
bscollin: I can delete if desired?? ;-)

bscollin:
This should be ready for re-review.  All comments addressed.

bkochuna:
Hmm... Looks like the CI fails from ML getting enabled. I'm not sure what failed looking at the CI output (maybe it took too long?)

Once this passes the CI, I'm good with merging.

bscollin:
So, this ML thing is weird so maybe that's why it was disabled in the first place.  Since its unrelated, I'm just reverting this change. Mitch has an active pull request that gets rid of ML completely so I'll let him handle this.

PR #55, opened 2017-09-29
Add -Werror=unused-variables flag for GNU

youngmit:
This should help in the dev process so we dont have to wait until
running on the CI before we find out that we missed something. We can
add more -Werror= flags as we squash more classes of warning


bscollin:
I'm a little concerned about making this the default.  This means we can't ever have an undefined variable while developing or debugging.  This likely happens very often.  I'm in favor of a variable that would enforce this "-D<PROJECT_NAME>_WARNING_ERRORS=ON  or something.  But I'm concerned about its impact on normal development.

bkochuna:
I'm with Ben on this one... I don't think we want it to be a default flag for DEBUG configs.
I think it would be better to update the automated testing configs to enable this option when evaluating pull requests. Either with the flag Ben suggests or by just setting the CMAKE_<>_COMPILER_FLAGS variable from the command line.

PR #56, opened 2017-09-30
Removes DBC_ from REQUIRE and ENSURE interface

bscollin:
Aaron discovered that having DBC_ in front of ENSURE and REQUIRE messes up the line numbers in gfortran 4.8.
It appears that pretty much any other combintation will not result in having the wrong line numbers.  He
tried not having DBC_, using ABC_, appending _DBC, even appending DbC_, and several other combinations.
Since the previous review made prepending DBC_ as optional, I'm reverting the names back to the original name.

### DIFF
--- a/src/DBC.h
+++ b/src/DBC.h
@@ -8,9 +8,9 @@
 !++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
 
 #ifdef FUTILITY_DBC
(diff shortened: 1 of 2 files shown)


bkochuna:
That IS weiiiird.

How were the line numbers wrong? Like just totally wrong?

aarograh:
@bkochuna The line numbers were always 4 greater than they should have been.  I spent quite a bit of time trying to figure out where that particular number came from and never came up with an answer.  It was a consistent offset that could be worked around, but very annoying when trying to debug things.

PR #58, opened 2017-10-19
Changes to REB partitioning algorithm.

apfitzge:
Description:
Changes to REB partitioning algorithm to give more convex domains. Updated unit-test.
Additionally, some quality changes.

1. Consider weight AFTER finding next expansion vertex (then determine whether to expand)
2. Only search neighboring vertices not already in domain
3. Rather than distance from original vertex, now takes distance from a reference point:
  i. Reference point starts as original vertex
 ii. If weight of external connections decreases (from previous expansion) switch reference point
     to the expansion.

CASL Ticket # - #5029

Updated PartitionGraph unit-test

Description:
Updated PartitionGraph unit-test to reflect changes in previous REB commit.

CASL Ticket # - #5029


bkochuna:
Thanks guys!

PR #59, opened 2017-10-19
Adding implementation of SlashRep for StringTypes

bkochuna:
This helps the VTU stuff, and should have been done forever ago.


bkochuna:
Testing revealed an unused variable.

Also needed to clean up some comments.

PR #60, opened 2017-10-20
Added capability to assign an integer to a StringType.

zdodson94:


src/Strings.f90
```diff
@@ -1030,6 +1036,38 @@ PURE SUBROUTINE assign_StringType2A_to_StringType2A(thisStr,s)
...
+!> @param i short integer
+!>
+!> The intent is that this will overload the assignment operator so a short
+!> (32-bit) integer can be assigned to a StringType.
+!>
+    PURE SUBROUTINE assign_Short_Integer_to_StringType(thisStr,i)
+      TYPE(StringType),INTENT(INOUT) :: thisStr
+      INTEGER(SNK),INTENT(IN) :: i
+      CHARACTER(LEN=11) :: l
+      WRITE(l,'(i11)') i
```
bkochuna: Does a "short" (4-byte?) int have up to 11 digits?

src/Strings.f90
```diff
@@ -1030,6 +1036,38 @@ PURE SUBROUTINE assign_StringType2A_to_StringType2A(thisStr,s)
...
+!> @param i long integer
+!>
+!> The intent is that this will overload the assignment operator so a long
+!> (64-bit) integer can be assigned to a StringType.
+!>
+    PURE SUBROUTINE assign_Long_Integer_to_StringType(thisStr,i)
+      TYPE(StringType),INTENT(INOUT) :: thisStr
+      INTEGER(SLK),INTENT(IN) :: i
+      CHARACTER(LEN=20) :: l
+      WRITE(l,'(i20)') i
```
bkochuna: Similar question here about the number of digits for the 64-bit integer.

Also do not forget leaving an extra place for negative numbers.

src/Strings.f90
bkochuna: I checked, and the answer is yes.

src/Strings.f90
bkochuna: I checked and the widths do include space for the "-" sign.

unit_tests/testStrings/testStrings.f90
```diff
@@ -81,6 +83,30 @@ PROGRAM testStrings
...
+!
+!Test assigning a long integer to string
+  COMPONENT_TEST('Assignment 64-bit Integer to String')
+  testSLK=1_SLK
+  testString=testSLK
+  ASSERT(CHAR(testString) == '1','testString=testSLK : 1')
+  testSLK=-9223372036854775807_SLK
+  testString=testSLK
+  ASSERT(CHAR(testString) == '-9223372036854775807','testString=testSLK : lowest')
+  testSLK=9223372036854775807_SLK
```
bkochuna: Instead of using the largest value use `HUGE(testSLK)` and `-HUGE(testSLK)` on RHS of assignment.

bkochuna:
Oh, one last thing would be to just amend your one commit.

bkochuna:
To amend your commit and push the changes you'll have to do:

1. make the changes
2. `git add` the changes
3. `git commit --amend`
4. Then `git push --force origin INT_StringType`

zdodson94:
Made the requested changes. Also added new tests for arrays.

PR #65, opened 2017-11-01
Fixes an issue with getMemUsage in Allocs

aarograh:
When using certain options in MPACT, the memstring variable returned
by getMemUsageChar_default had an unreadable value that caused an
ungraceful crash.  While this may ultimately be an issue with the
usage of Allocs in MPACT, these changes both eliminated redundant
code and resolved the ungraceful exit in MPACT.

### DIFF
--- a/src/Allocs.f90
+++ b/src/Allocs.f90
@@ -641,26 +641,9 @@ MODULE Allocs
 !>
     FUNCTION getMemUsageChar_default() RESULT(memstring)
       CHARACTER(LEN=ALLOC_MEMSTRING_LENGTH) :: memstring
-      CHARACTER(LEN=5) :: unit
-      REAL(SRK) :: mem,bytes
-      REAL(SRK),PARAMETER :: KB2bytes=1024_SRK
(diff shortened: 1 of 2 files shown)


stimpsonsg:
Brendan, I've reviewed this.  See my comments, which are mostly pertinent to future changes.  This can likely be merged as is.

bscollin:
Should we change that to a string instead of using a fixed length character array?

stimpsonsg:
Probably would help with the ****** issues.  

Currently set to 14:
INTEGER(SIK),PARAMETER :: ALLOC_MEMSTRING_LENGTH=14

bkochuna:
I'd prefer that we update the memory edits from MPACT to use GB instead of MB, maybe increase the size of ALLOC_MEMSTRING_LENGTH.

I'd like to avoid adding the dependency on StringTypes.

Since that's an MPACT issue, I'm going to go ahead and merge this.

bkochuna:
Thanks Aaron!

PR #69, opened 2017-11-10
Trilinos tpl

bscollin:


bscollin:
Alright, this can't be merged yet because it won't work with the current CASL version of Trilinos.  But I should point out that the version that uses the TPL takes 12 minutes in the CI compared to the previous version which builds Trilinos and takes between 40 and 50+ minutes.

bkochuna:
I will look at this today

cmake/ExtraRepositoriesList.cmake
```diff
@@ -1,10 +1,16 @@
...
-  Trilinos           ""  GIT  ${Futility_URL_REPO_BASE}Trilinos            PRE         Continuous
-  TeuchosWrappersExt ""  GIT  ${Futility_URL_REPO_BASE}TeuchosWrappersExt  PRE         Continuous
-)
+IF(NOT ${PROJECT_NAME}_TRILINOS_TPL)
+  TRIBITS_PROJECT_DEFINE_EXTRA_REPOSITORIES(
+    Trilinos           ""  GIT  ${Futility_URL_REPO_BASE}Trilinos            PRE         Continuous
+    TeuchosWrappersExt ""  GIT  ${Futility_URL_REPO_BASE}TeuchosWrappersExt  PRE         Continuous
+  )
+ELSE()
+  TRIBITS_PROJECT_DEFINE_EXTRA_REPOSITORIES(
```
bkochuna: Why is FakeRepo needed?

TPLsList.cmake
```diff
@@ -7,14 +7,31 @@
...
+    #Rythmos       TRIBITS_PKG   SS
+    CTeuchos      TRIBITS_PKG   SS
+    ForTeuchos    TRIBITS_PKG   SS
+    )
+ELSE()
+  SET(TRILINOS_PACKAGE_TPLS_LIST)
+ENDIF()
+
+TRIBITS_REPOSITORY_DEFINE_TPLS(
+  MPI      "${${PROJECT_NAME}_TRIBITS_DIR}/core/std_tpls/" SS
```
bkochuna: Should these be ST instead of SS also?

.travis.yml
```diff
@@ -5,14 +5,14 @@ services:
 
 before_install:
  - echo $PWD
- - docker pull dalg24/vera_tpls
+ - docker pull bcollins/futility_tpl
```
bkochuna: Documentation for this docker image would be helpful.

bscollin: Yeah, I'm working on automating the process and then adding the script to the repo but its basically the vera_tpls docker with Futility+Trilinos built and installed.

cmake/ExtraRepositoriesList.cmake
bscollin: So, I need to follow up with @bartlettroscoe about this.  If the file ExtraRepositoriesList is included we must call `TRIBITS_PROJECT_DEFINE_EXTRA_REPOSITORIES` but that must have at least one repo listed.  I realize that this isn't idea but the other option is to prevent Futility to include the ExtraRepositoriesList file which I figured would add more confusion if we ever add a new extra repo which we don't prebuild.

stimpsonsg:
Are we able to revisit this now?

bscollin:
A few things need to be made... 1... gcc 5.4 needs to be default for CASL... we would need to rebuild the docker to have 5.4 and all the TPLS... that would need to be passing... so... it will take a bit.

stimpsonsg:
Do we want to keep this open?  It's been around for a very long time.

While we still plan to move to trilinos as a tpl, still seems like a bit before that will happen.

bscollin:
I'd like to keep this around until we Trilinos updated.  Sounds like there's some issues with DTK wrappers  that we need to work through.

PR #70, opened 2017-11-28
Assign reals

djabaay:


bkochuna:
So does this fix Scott’s issue with null characters?

djabaay:
@bkochuna It does not.  This is functionality I wanted for making the error message in setVars in the state object simpler.  I want to print the min and max valid values for a given state variable, and I'd rather not write reals to characters to strings while just writing integers to strings.

djabaay:
@bkochuna If you have some time to kill, feel free.

djabaay:
@bscollin I know, I was thinking the same thing previously.  The issue at hand is that we are overloading the assignment operator, so there is no place for the optional format argument of which I was thinking.  If you're worried about discrepancies between systems and builds producing a different 15th digit, we can change the default to something less than 15.  What do you think would be an acceptable accuracy that is build and system consistent?

djabaay:
@zdodson94 Nice test Zack, glad it works for that case too!  It could be worth it for the TINY operation, just to cover all the bases.

bkochuna:
I also want stress, that this is relatively low priority for compared to other problems.

bscollin:
@djabaay, I think you're right to do full digits by default.  I was thinking about maybe having a editor object which stores the number of digits for outputs but that's way outside of the scope of this ticket.

bscollin:
Whats the hold up on merging this?

bscollin:
@djabaay @bkochuna can we merge this?

djabaay:
I am ok merging it.

bscollin:
done.  it was sitting there

PR #71, opened 2017-11-29
Checking if the string variable is len_trim > 0 before writing.

djabaay:
Description:
In an attepmt to eliminate erroneous characters being printed for a
null/empty string type on a parameter list, the parameter list value is
checked to make sure it has a length > 0 before printing.  If it is ==
0, an empty string is printed (i.e. '') in its place.

CASL Ticket # - PHI #5241

### DIFF
--- a/src/ParameterLists.f90
+++ b/src/ParameterLists.f90
@@ -4705,7 +4705,7 @@ MODULE ParameterLists
       CHARACTER(LEN=PARAM_MAX_DAT_LEN) dtype
       CHARACTER(LEN=12) :: fmt
       INTEGER(SIK) :: i
-      TYPE(StringType) :: sprefix,sdtype
+      TYPE(StringType) :: sprefix,sdtype,tmpstr
 
       i=1
       IF(PRESENT(indent)) i=i+indent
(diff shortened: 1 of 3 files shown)


src/ParameterLists.f90
```diff
@@ -4718,12 +4718,14 @@ SUBROUTINE edit_ParamType_STR(thisParam,funit,indent,prefix,paddtw)
         ENDIF
       ENDIF
       WRITE(fmt,'(i12)') i; fmt=ADJUSTL(fmt)
+      tmpstr=''
+      IF(LEN_TRIM(thisParam%val) > 0) tmpstr=thisParam%val
       IF(LEN_TRIM(thisParam%description) == 0) THEN
         WRITE(UNIT=funit,FMT='('//TRIM(fmt)//'x,a)') sprefix// &
-          sdtype//' :: '//thisParam%name//'='//thisParam%val
+          sdtype//' :: '//thisParam%name//'='//tmpstr
```
bkochuna: why write ```tmpstr```?

Just write ```sdtype//' :: '//thisParam%name//'='```

src/ParameterLists.f90
```diff
@@ -4718,12 +4718,14 @@ SUBROUTINE edit_ParamType_STR(thisParam,funit,indent,prefix,paddtw)
         ENDIF
       ENDIF
       WRITE(fmt,'(i12)') i; fmt=ADJUSTL(fmt)
+      tmpstr=''
+      IF(LEN_TRIM(thisParam%val) > 0) tmpstr=thisParam%val
```
bkochuna: I think there are some other parameter types for arrays of strings that also need to get modified.

bkochuna:
Another issue is can we reproduce? And that's the biggest one. Need to find a machine that gives those null characters.

src/ParameterLists.f90
djabaay: I didn't want to have a nested set of IF-blocks.  Because we already check if the description is LEN_TRIM=0, now you have 4 logical cases, and it looks ugly.  I can simplify it differently if you'd prefer.  I may do it anyways to get rid of the IF-ELSE block and just have a single write statement.

djabaay:
I'll ask what machine and build.

bkochuna:
This looks fine to me.

One comment for the future: I think the refactoring of the test variables should have been made its own commit.

Will merge once the CI gives the green light.

PR #72, opened 2017-12-08
Adds .XOR. operator for logicals to IntrTypes

aarograh:


djabaay:
As an fyi, this intrinsic already exists in Fortran as .NEQV. which is the updated standard for .XOR.  

.XOR. is for f77 compatibility.

https://gcc.gnu.org/onlinedocs/gfortran/XOR.html

src/IntrType.f90
```diff
@@ -1000,6 +1008,19 @@ ELEMENTAL FUNCTION notequalto_logical(l1,l2) RESULT(bool)
     ENDFUNCTION notequalto_logical
 !
```
djabaay: xor_logical is the same function as notequalto_logical.

aarograh: Ah, good call... totally didn't know this was in there.  Do we want to revert the change, or just leave both?

aarograh:
Dan, thanks for clarifying.  I didn't realize the .NEQV. operator existed.  I've always checked for .XOR. (since that is the standard name for this operator in other languages) with no luck.

If we want to revert this change I can just switch to using .NEQV. or the overloaded /= operators where appropriate.

djabaay:
Yea, I know the feeling, I've looked for it in the past as well since that's what I was familiar with.  I'd say revert the changes, since we can use the /= operator which is a little shorter and cleaner.

PR #75, opened 2017-12-11
Simplify usage of DBC

bkochuna:
Description:
For client code that will use DBC, it would have to add two lines:
  #include "DBC.h"
  USE DBC

I've modified "DBC.h" to to include the USE statement when
FUTILITY_DBC is defined. Now client code wanting to use DBC should
just do:

  MODULE client
  #include "DBC.h"
  ...

Also:
  * made indentation consistent with other source files
  * changed variable definitions to avoid unnecessary implied save
    attribute

CASL Ticket # - #5200

### DIFF
--- a/src/DBC.f90
+++ b/src/DBC.f90
@@ -41,7 +41,6 @@ MODULE DBC
 #endif
 
   PRIVATE
-
 !
 ! List of Public items
   PUBLIC :: DBC_FAIL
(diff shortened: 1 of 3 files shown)


bkochuna:
Trying to start using DBC. Found this obvious improvement.

bscollin:
Not opposed to this but it just hides more than I thought it benefited.  Same change could be made to the unit test harness.

bkochuna:
Should we pose this question to a larger group of dev's and go with the majority?

bscollin:
@CASL/futility-developers please up vote/down vote this pull request to address the inclusion of the `USE` statement in the `#include` file.  We'll go with majority as of 9 am on 12/12.

djabaay:
If this module is eventually going to be used in every file of the code, I think it'll be understood that the include statement has the USE statement.

Syntactically similar to the way we include other header files, they don't have an accompanying USE statement.

bscollin:
@djabaay what other code actually has an include file that has a `USE` statement?

djabaay:
@bscollin That is sort of my point.  They don't.  `#include <finclude/petsc.h>`  No USE statement

djabaay:
Or to ask it another question, would there be necessary circumstances where a developer needs to `#include` the header and **not** USE it?

bscollin:
@djabaay well... petsc has two methods actually.  They have the include file methods (the old F77 way) and then they have the option to use the `USE` statement (and not an include file).  We use the include file because its portable.  In this case, its fairly different.  Like I said, I don't really care either way.  Just wanted to make sure we make a conscious decision that is applied uniformly across the entire project (which means we should change unit test harness to add the use file).

@bkochuna Just remembered why we did unit test helper the other way.  By having it in the include file, it implies that someone put it with the rest of the `USE` statements.  In principle, an include file can go anywhere in the code but a `USE` statement only goes at the top.  If everyone knows that that include file has a `USE` statement, then they will place it correctly but its a little hidden.

djabaay:
@bscollin Ah, the explanation makes more sense now, didn't think about include files being put anywhere.

bscollin:
@djabaay but again... I could go either way.  Its not a big deal to me and based on the survey... you have 2 yes votes and 4 eh votes.  I'm good to accept when the indentation thing is fixed.

bscollin:
I will say, I care a lot more that people are starting to use it!!

bkochuna:
Amended commit for backtrace indentation and pushed.

Regarding the points about the ```USE DBC``` my thoughts are:
* You would not have a client with just ```USE DBC``` or just ```#include "DBC.h"```. They are always used together.
* The location of the ```USE DBC``` is limited, and the ```#include "DBC.h"``` basically has to be before using any of the preprocessor macros for whatever scope in which its used.
* I would think it would be more confusing to write code that uses ```USE DBC``` and ```#include "DBC.h"``` and NOT have these lines adjacent to one another. Or in other words my expectation is these two lines will always appear next to each other as is.
  * Therefore just combine them to avoid errors where you have to have 2 lines, and maybe you forget one.

That makes more sense to me then leaving them separate.

clarnokt:
Wish i'd seen this earlier so i could comment.  "DBC.h" is a terrible name
for an include file.  Do you really think Futility is the only code out
there with DBC.h?  Make it: FUTILITY_DBC.h

SCALE and Exnihilo get away with it because of where they put their include
structure, but MPACT does not:

   - Exnihilo: #include "Nemesis/harness/DBC.hh"
   - SCALE:    #include "ScaleDBCF/DBCF.h"
   - MPACT:    #include "DBC.h"

bkochuna:
That's a good point Kevin. We'll make a new pull request for that. (I want to get HDF5 stuff for CASL ticket 5200 merged first.)

bkochuna:
So I found the real reason why we CAN'T do this.

By putting the ```USE``` statement in preprocessed header, the module dependency is missed by CMake/Make. So build errors may arise when compiling in parallel if modules are compiled out of order.

I've removed the ```USE``` from the header in #76

PR #78, opened 2017-12-15
Memprof update

bscollin:
Some updates to memprof:
  * can turn edits on and off
  * can set a threshold for the minimum change that will edit the memory
 * has a verbose mode which will edit all of the data for each process.

### DIFF
--- a/src/MemProf.f90
+++ b/src/MemProf.f90
@@ -97,7 +97,6 @@ MODULE MemProf
 
       INTEGER(C_LONG_LONG) :: tmpL1, tmpL2
       CHARACTER(LEN=16) :: filename
-      CHARACTER(LEN=45)  :: tmpchar
 
       IF(ASSOCIATED(myLog)) THEN
         thisMP%pe=>pe
--- a/src/MemProf.f90
+++ b/src/MemProf.f90
@@ -21,6 +21,7 @@ MODULE MemProf
   USE ISO_C_BINDING
   USE IntrType
   USE Strings
+  USE FileType_Fortran
   USE FileType_Log
   USE ExceptionHandler
   USE ParameterLists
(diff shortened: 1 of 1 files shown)


bscollin:
Leaving until noon tomorrow for others to comment

bkochuna:
When I was looking at this the other day I had some thoughts about some other potential improvements, which need not be addressed now, just want to pass them along.
* one was doing file edits by process, so glad to see that
* I was thinking it might be useful in the summary output to include the rank of the process with the max use (reduce with MPI_MAX_LOC)
* For the dmem, it may be useful to have a different unit (KB instead of MB), especially for small thresholds
* I think the dependencies on the init interface could be reduced.
  * You only use pe%world, so you don't actually need the PE type, you could just reference MPI_COMM_WORLD (PE_COMM_WORLD) inside init.
  * You don't necessarily need to pass a log file either. You could just pass an integer unit number, and set up a log file internally
  * although I think it would get a little tricky because if it is open (as it would be for root) you would have to set the open status without calling fopen.

Another challenge I saw with this is that it is effectively MPI_COMM_WORLD collective, and for the way its implemented that's fine, but if you want to profile other parts of the code more finely, by just adding these calls willy nilly whereever, that is a little difficult to do with the current init dependencies.

Given that it might make sense to implement a singleton (public module global variable) for this since you are typically going to use it to profile a whole executable program, and not necessarily different subsections to different output files.

As for avoiding deadlocks if it gets called in branches that don't have all processors, I'm not sure how to get around this... I spent some time looking to see if it was possible or there were any MPI interfaces that just let you create a new communicator based all processes that called it, but that's not possible, and when thinking more about the MPI model I can see why that would not be possible, so there's probably not a way around it (other than having the verbose mode function independently of the edits with the reduce calls).

So some of that I think certainly lends itself to more use cases (which can facilitate debugging. But these changes are fine and certainly improve the functionality.

If you don't want to handle any of these now, then I would suggest just adding them as TODO comments.

bscollin:
@bkochuna I'd rather add these as an issue and track them there instead of in the code.  There are some things I agree with, others I think need more discussion, which the issue would facilitate.  I'm going to go ahead and add the max_loc because its an easy change.  The dmem units, I wanted to keep them all consistent since there isn't an existing banner to tell the user the different units but it occurs to me that I can do that during init... or keep track of the first time ```%edit``` is called.

bscollin:
I added in labels, scaled the dmem to be MB instead of GB, and recorded the process with the maximum change.  The remainder of the issues mentioned by @bkochuna I"ll move to an issue to be discussed/addressed later.

PR #81, opened 2017-12-30
Mesh transfer

bscollin:


src/MeshTransfer.f90
```diff
@@ -0,0 +1,1570 @@
...
+!>  transfer during initialization.  A mesh transfer matrix or linear system
+!>  will be used to transfer input data onto the output mesh.  The call to
+!>  transfer will transform the input data onto the output mesh using the transfer
+!>  matrix and/or linear system which was defined during init.  As long as the
+!>  incoming and outgoing mesh/moments are the same, transfer can be called
+!>  an arbitrary number of times very efficiently.  If the mesh is different,
+!>  a new transfer object will be needed.
+!>
+!> @par Module Dependencies
+!>  - @ref IntrType "IntrType": @copybrief IntrType
```
stimpsonsg: probably either need to add the rest of the module dependencies or delete this one

bscollin: yawn...

stimpsonsg: agreed...but if you're going to put them in...

src/MeshTransfer.f90
```diff
@@ -0,0 +1,1570 @@
...
+    INTEGER(SIK) :: nmesh_out=0
+    !>
+    INTEGER(SIK) :: poly_transfer=-1
+    !>
+    REAL(SRK),ALLOCATABLE :: mesh_in(:)
+    !>
+    REAL(SRK),ALLOCATABLE :: mesh_out(:)
+    !>
+    REAL(SRK),ALLOCATABLE :: transferMatrix(:,:)
+    !>
```
stimpsonsg: need descriptions here

src/MeshTransfer.f90
```diff
@@ -0,0 +1,1570 @@
...
+      CALL pList%get('map_out',tmpstr)
+      SELECTCASE(CHAR(tmpstr))
+        CASE('CONTINUOUS')
+          this%MapType_out=MapType_CONTINUOUS
+        CASE('POINT')
+          this%MapType_out=MapType_POINT
+        CASE('VOLUME')
+          this%MapType_out=MapType_VOLUME
+        CASE DEFAULT
+          WRITE(*,*) "invalid outgoing map type"
```
stimpsonsg: exception handler here?

src/MeshTransfer.f90
```diff
@@ -0,0 +1,1570 @@
...
+      CALL pList%get('map_in',tmpstr)
+      SELECTCASE(CHAR(tmpstr))
+        CASE('CONTINUOUS')
+          this%MapType_in=MapType_CONTINUOUS
+        CASE('POINT')
+          this%MapType_in=MapType_POINT
+        CASE('VOLUME')
+          this%MapType_in=MapType_VOLUME
+        CASE DEFAULT
+          WRITE(*,*) "invalid incomming map type"
```
stimpsonsg: and here

src/MeshTransfer.f90
```diff
@@ -0,0 +1,1570 @@
...
+      CALL myPL%add('LinearSolverType->timerName','testTimer')
+      CALL myPL%add('LinearSolverType->matType',DENSESQUARE)
+      CALL myPL%add('LinearSolverType->A->MatrixType->isSym',.FALSE.)
+      CALL myPL%add('LinearSolverType->A->MatrixType->n',nin)
+      CALL myPL%add('LinearSolverType->x->VectorType->n',nin)
+      CALL myPL%add('LinearSolverType->b->VectorType->n',nin)
+      CALL TLS%init(myPL)
+      CALL myPL%clear()
+
+      ALLOCATE(x(nin))
```
stimpsonsg: in quite a few of these, you allocate x.  It might automatically get cleaned up since it's a local real, but might be good to deallocate it/them for completeness

unit_tests/testMeshTransfer/testMeshTransfer.f90
```diff
@@ -0,0 +1,1040 @@
...
+      CALL testPL%add('MeshTransfer->map_in','VOLUME')
+      CALL testPL%add('MeshTransfer->map_out','POINT')
+      CALL testPL%add('MeshTransfer->volumemesh_in',mesh)
+      CALL testPL%add('MeshTransfer->pointmesh_out',mesh_out)
+      CALL testPL%add('MeshTransfer->poly_transfer',4)
+
+      COMPONENT_TEST('%init_base')
+      CALL testMT%init_base(testPL)
+
+      ASSERT_EQ(testMT%MapType_in,3,'map_in')
```
stimpsonsg: should we use the enumerations when checking MapType_in and MapType_out?

unit_tests/testMeshTransfer/testMeshTransfer.f90
```diff
@@ -0,0 +1,1040 @@
...
+
+      ALLOCATE(c(8))
+      DO i=1,8
+        c(i)=1.0_SRK/REAL(i,SRK)
+      ENDDO
+
+      ALLOCATE(sol(11))
+
+      ALLOCATE(ref(11))
+      ref(1)=0.6345238095238095238095238095238095238095238095238095238095238095238_SRK
```
stimpsonsg: the amount of precision here is always hilarious

bscollin: So more?!?

src/MeshTransfer.f90
```diff
@@ -0,0 +1,202 @@
...
+!> @param coef  The coefficients of the Legendre expansion
+!> @param globX The value of x in the global coordinate
+!> @param h     The height of the node, OPTIONAL (if not provided a height of 2 is assumed)
+!> @param nodeX The center of the node, OPTIONAL (if not provided a position of 0 is assumed)
+!>
+!> This routine returns the value of an arbitrary order Legendre expansion at a
+!> specific point for NEM and SANM nodal methods.
+!>
+    FUNCTION LPPointVal(coef,globX,h,nodeX) RESULT(y)
+      REAL(SRK),DIMENSION(:),INTENT(IN) :: coef
```
bkochuna: Less characters and better column orientation if you do:
```REAL(SRK),INTENT(IN) :: coef(:)```

File: src/MeshTransfer.f90 [hunk 9]
```diff
@@ -14,34 +14,848 @@
...
+      CALL pList%get('map_in',tmpstr)
+      SELECTCASE(CHAR(tmpstr))
+        CASE('CONTINUOUS')
+          this%MapType_in=MapType_CONTINUOUS
+        CASE('POINT')
+          this%MapType_in=MapType_POINT
+        CASE('VOLUME')
+          this%MapType_in=MapType_VOLUME
+        CASE DEFAULT
+          WRITE(*,*) "invalid incomming map type"
```
bkochuna: Change to DBC, no ```WRITE(*,*)```

src/MeshTransfer.f90
bkochuna: I think DBC would be better (shouldn't happen in production, only development).

src/MeshTransfer.f90
```diff
@@ -14,34 +14,848 @@
...
+      this%isInit=.FALSE.
+    ENDSUBROUTINE clear_1Dbase
+!
+!-------------------------------------------------------------------------------
+!> @brief
+!> @param this  The mesh transfer object
+!>
+!> This routine
+!>
+    SUBROUTINE transfer_1Dbase(this,data_in,data_out)
```
bkochuna: For type bound procedures, I'm generally not a fan of the passed object being called ```this```. It is not descriptive enough. In particular when you have a single file defining multiple concrete types, if the routine is long enough that you can be looking in the middle of it and not see the encapsulating construct or definition of ```this```, then it's not always obvious what type ```this``` is.

I think it's much better to have ```thisXXX``` where ```XXX``` is some acronym for the type. In this case (1DB, 1DC or 1DR).

src/MeshTransfer.f90
```diff
@@ -416,14 +433,28 @@ SUBROUTINE setupP2P(TM,mesh_in,mesh_out)
...
       ALLOCATE(TM(SIZE(mesh_out),SIZE(mesh_in)))
+      TM(:,:)=ZERO
+
+      DO i=1,SIZE(mesh_out)
+        j=findIndex(mesh_in,mesh_out(i),.FALSE.,incl=1)
+        IF(j>0) THEN
+          dx=mesh_in(j+1)-mesh_in(j)
+          TM(i,j)=(mesh_in(j+1)-mesh_out(i))/dx
+          TM(i,j+1)=(mesh_out(i)-mesh_in(j))/dx
+        ELSEIF(j==-1) THEN
```
bkochuna: Spacing on relational operators ```(arg1 == arg2)```

src/MeshTransfer.f90
```diff
@@ -6,7 +6,54 @@
...
+!>
+!>  In all cases, the user specifies the input mesh or moments and output mesh
+!>  or moments and to use the native implementations or a polynomial projection
+!>  transfer during initialization.  A mesh transfer matrix or linear system
+!>  will be used to transfer input data onto the output mesh.  The call to
+!>  transfer will transform the input data onto the output mesh using the transfer
+!>  matrix and/or linear system which was defined during init.  As long as the
+!>  incoming and outgoing mesh/moments are the same, transfer can be called
+!>  an arbitrary number of times very efficiently.  If the mesh is different,
+!>  a new transfer object will be needed.
```
bkochuna: It'd be cool to insert latex-only blocks for the actual mathematical formulations. See XenonSamarium.f90 for an example.

This is something I'd like to see us do more of, because it gives us the mathematical "requirements" in the comments.

src/MeshTransfer.f90
bscollin: Done

src/MeshTransfer.f90
bscollin: Done :(

src/MeshTransfer.f90
bscollin: Doner

src/MeshTransfer.f90
bkochuna: Clarify or add a todo, that the object does NOT assume the mesh is distributed (e.g. there's no MPI in this).

src/MeshTransfer.f90
bscollin: Done... lots of them

unit_tests/testMeshTransfer/testMeshTransfer.f90
bscollin: I didn't because I hand't made them public.  This was mostly because the test is the only reason to.

src/Constants_Conversion.f90
```diff
@@ -33,6 +33,7 @@ MODULE Constants_Conversion
   REAL(SRK),PUBLIC,PARAMETER ::    ZERO=0.000000000000000_SRK !16 digits
   REAL(SRK),PUBLIC,PARAMETER ::     ONE=1.000000000000000_SRK !16 digits
   REAL(SRK),PUBLIC,PARAMETER ::     TWO=2.000000000000000_SRK !16 digits
+  REAL(SRK),PUBLIC,PARAMETER ::    HALF=0.500000000000000_SRK !16 digits
```
bkochuna: One other comment on this based on experience.

At one point I thought it'd be useful to put lots of common numbers like this in a module (I think I did this for TRIAD and DeCART) and after some time, I found it REALLY annoying to move code around have this module. It was easier to just type out ```0.0d0``` or ```0.0_SRK``` because this prevents a dependency. Normally though I'll go with approaches that require the least amount of typing.

src/MeshTransfer.f90
bscollin: I added the exception handler.  For initiialization, I think we should give clear output errors.  Everything after init, I think DBC is good enough.  That's just my opinion but I think its a good balance of user friendliness and robust testing.  At least this is the standard I was going to propose.  Thoughts?

src/MeshTransfer.f90
bkochuna: On a side note, I've been thinking about how we can automate whitespace formatting to give the code a consistent look without needing to have all the developers remember to fix these little quirks.

src/Constants_Conversion.f90
bscollin: yeah, I go back and forth.  I needed Pi anyway, so it was already loaded.  I added HALF because it saves some room.  In nodal, I added a lot of local parameters for other numbers 4, 7, etc. but there wasn't a lot of reason to add those here since they are fairly uncommon.  I think common numbers here makes sense.

src/MeshTransfer.f90
bscollin: This was here for a long time.  I just moved the code.  I'll change it to your way because I like it more.

src/MeshTransfer.f90
bscollin: Probably more concerning is how inconsistent I was about this.  Fixed now.

src/MeshTransfer.f90
bscollin: Added this:
All routines implemented here only handle mesh transfers in serial.  The base class does not prevent an extension for parallel maps if someone wishes to do so in the future.

bscollin: I'm not entirely sure how to represent this mathematically outside of saying ```out=MT*in``` or ```LS%A*out = in```.  Do you have something specific that you think would add value to the description?

bscollin:
Addressed all of @stimpsonsg comments and most of @bkochuna, but there were a few questions I had

src/MeshTransfer.f90
bkochuna: For me the question has always boiled down to:

Do I expect this branch that is an error to happen during normal/production execution, perhaps as a corner case? Or as something the user can get to by modifying the input?

Or is this error really for a developer (e.g. a lot of the ```IF(.NOT.isInit)``` or ```unknown modular ray tracing enumeration```) indicating their usage of the provider is incorrect/not supported.

exception handler is fine.

bscollin: So I'm thinking that I could imagine the input, specifically through decorators, the transfer type could be defined by a user.  You're probably right that these errors were a stretch but I'd rather error on the side of caution at give a better error message than just crash the code immediately.

bscollin:
@bkochuna Anything else that needs to happen to get this merged?

src/MeshTransfer.f90
```diff
@@ -212,6 +212,9 @@ SUBROUTINE mt_trnsfr_absintfc(this,data_in,data_out)
       PROCEDURE,PASS :: init => init_1DCyl
   ENDTYPE MeshTransfer_1DCyl
 
+  !> Name of module
+  CHARACTER(LEN=*),PARAMETER :: modName='MESTTRANSFER'
```
bkochuna: Typo ```MESHTRANSFER```

bkochuna:
This looks good to merge now.

PR #85, opened 2018-01-03
Adds 3D capability to the ABBoxType.  All 3D arguments and funcitonality

aarograh:
are optional, preserving any pre-existing uses of this type.

### DIFF
--- a/src/Geom_Box.f90
+++ b/src/Geom_Box.f90
@@ -576,16 +576,14 @@ MODULE Geom_Box
       LOGICAL(SBK) :: inside
 
       inside=.FALSE.
-      IF(p%dim == 2 .OR. (p%dim == 3 .AND. .NOT.thisABB%is3D)) THEN
+      IF(p%dim > 1) THEN
         inside=((p%coord(1) > thisABB%xMin-fuzz) .AND. &
           (p%coord(1) < thisABB%xMax+fuzz) .AND. &
           (p%coord(2) > thisABB%yMin-fuzz) .AND. &
           (p%coord(2) < thisABB%yMax+fuzz))
-      ELSEIF(p%dim == 3 .AND. thisABB%is3D) THEN
-        inside=((p%coord(1) > thisABB%xMax-fuzz) .AND. &
-          (p%coord(1) < thisABB%xMax+fuzz) .AND. &
-          (p%coord(2) > thisABB%yMin-fuzz) .AND. &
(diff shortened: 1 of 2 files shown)


src/Geom_Box.f90
```diff
@@ -555,11 +574,22 @@ ELEMENTAL FUNCTION inside_ABBoxType(thisABB,p) RESULT(inside)
...
+          (p%coord(2) < thisABB%yMax+fuzz))
+      ELSEIF(p%dim == 3 .AND. thisABB%is3D) THEN
+        inside=((p%coord(1) > thisABB%xMax-fuzz) .AND. &
+          (p%coord(1) < thisABB%xMax+fuzz) .AND. &
+          (p%coord(2) > thisABB%yMin-fuzz) .AND. &
+          (p%coord(2) < thisABB%yMax+fuzz) .AND. &
+          (p%coord(3) > thisABB%zMin-fuzz) .AND. &
+          (p%coord(3) < thisABB%zMax+fuzz))
+      ENDIF
+
```
bscollin: So I would do this a little differently:
```
IF(p%dim > 1) THEN
  inside=((p%coord(1) > thisABB%xMin-fuzz) .AND. &
         (p%coord(1) < thisABB%xMax+fuzz) .AND. &
         (p%coord(2) > thisABB%yMin-fuzz) .AND. &
         (p%coord(2) < thisABB%yMax+fuzz))
 IF(p%dim == 3 .AND. thisBB%is3D) inside=(inside .AND. &
          (p%coord(3) > thisABB%zMin-fuzz) .AND. &
           (p%coord(3) < thisABB%zMax+fuzz))
ENDIF
```

It just cuts down duplicating a few lines of code.  Just a suggestion though.  I'm fine with the changes as made.

bkochuna:
What if the ABBox type was an extended type?

It would just overload the 2D routines, then call them, then operate on the z dimension?

If this is how the other objects are setup, just go with it, but maybe we think about refactoring 3D to new types and make that a separate issue...

aarograh:
I addressed Ben's comment on the source and Brendan's comment on the commit message.  I also added a test for the ABBoxType.

Brendan, there are several improvements that could be made to these types, including having a more sensible class hierarchy.  I think they serve the necessary purpose right now, though, so I plan to leave them be until we actually need those changes.

aarograh:
@bkochuna Are you good with this?

bkochuna:
Yeah.

PR #87, opened 2018-01-10
Adds 3 new conversions to Constants_Conversion

aarograh:
Added Bar2MPA, kg2g, and m2cm.


djabaay:
Just a thought Aaron, but if we were going to add a unit conversion to this module that involved squared units, would the numerical "2" be potentially confusing?  Would it be worthwhile to just modify the conversions to use the literal "To" instead?

bscollin:
I kind of disagree with @djabaay on this one.  There isn't a good reason for squared units, just square the conversion factor `m2cm**2`.  I also think if you add To, its harder to read.

djabaay:
Fair point @bscollin about the `**2`

PR #88, opened 2018-01-17
Rst compress

bkochuna:


src/FileType_HDF5.f90
```diff
@@ -109,6 +109,8 @@ MODULE FileType_HDF5
     LOGICAL(SBK) :: isInit=.FALSE.
     !> Whether or not the file uses compression for writing
     LOGICAL(SBK) :: hasCompression=.FALSE.
+    !> Option for gzip compression
+    INTEGER(SIK),PRIVATE :: zlibOpt=-1
```
aarograh: Might be good to expand the comment on this to explain the [-1,9] range of allowed values.  Also, I assume those values are defined by the HDF5 specs?  That would be good to note as well.

bkochuna: Thanks Aaron, I'll add those comments.

PR #92, opened 2018-02-01
PartitionGraph - Fixed RMS to actually be RMS rather than 2-norm.

apfitzge:
Description:
RMS did not have 1/n inside the square root, thus it was more of a
2-norm. This has been fixed by adding the 1/n.

CASL Ticket # - #5014


apfitzge:
This was something that should have been merged months ago. RMS needs a 1/n otherwise it'll give uselss metric.

PR #94, opened 2018-02-03
Add overwrite status to HDF5 files

rks171:
This is to address Phase 1 of Issue 74.  I ran the MPACT and CTF Nightly's and all continue to pass.  This allows Filetype_HDF5 to be used for reading and writing all CTF HDF5 files.


Futility-CopyrightAndLicense.txt
```diff
@@ -0,0 +1,26 @@
+Copyright (c) 2016, Oak Ridge National Laboratories
```
bscollin: What is this and how'd it get here?

header
```diff
@@ -0,0 +1,9 @@
+!+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
```
bscollin: This too

bscollin:
So it looks like you started from casl-dev... which is an issue.  We should just cherry-pick your existing commit onto the head of github master.  (then add the test)

rks171:
I cloned this repo and put the changes in there.  Pushed to a new branch called issue_74_full, which includes the unit test.

PR #95, opened 2018-02-04
Add new file status to allow overwriting HDF5 datasets

rks171:
This includes the fixes for issue_74 merge request, which was not built on the correct version of the repo.

### DIFF
--- a/unit_tests/testHDF5FileType/testHDF5FileType.f90
+++ b/unit_tests/testHDF5FileType/testHDF5FileType.f90
@@ -1113,6 +1113,40 @@ PROGRAM testHDF5
     SUBROUTINE testHDF5FileTypeOverwrite()
       TYPE(HDF5FileType) :: h5
       CHARACTER(LEN=EXCEPTION_MAX_MESG_LENGTH) :: msg,refmsg
+      REAL(SDK),ALLOCATABLE :: testD1(:),testD2(:,:),testD3(:,:,:),testD4(:,:,:,:)
+      REAL(SDK),ALLOCATABLE :: testD5(:,:,:,:,:),testD6(:,:,:,:,:,:),testD7(:,:,:,:,:,:,:)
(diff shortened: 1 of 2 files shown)


unit_tests/testHDF5FileType/testHDF5FileType.f90
```diff
@@ -1107,6 +1108,560 @@ SUBROUTINE testHDF5FileTypeWrite()
...
+      CALL h5%fwrite('groupCNCHAR->CNCHAR2',refCNCHAR2,SHAPE(refCNCHAR2))
+      CALL h5%fwrite('groupCNCHAR->CNCHAR3',refCNCHAR3,SHAPE(refCNCHAR3))
+      CALL h5%mkdir('groupST')
+      CALL h5%fwrite('groupST->memST0',refST0)
+      CALL h5%fwrite('groupST->memST1',refST1,SHAPE(refST1))
+      CALL h5%fwrite('groupST->memST2',refST2,SHAPE(refST2))
+      CALL h5%fwrite('groupST->memST3',refST3,SHAPE(refST3))
+      CALL h5%mkdir('groupC')
+      CALL h5%fwrite('groupC->memC1',refC1,LEN(refC1))
+
```
bscollin: It doesn't appear that there are any asserts to ensure it actually changes the data.  I'd suggest writing some basic data (say 0's, then overwrite with the correct data and assert it actually changes them.

rks171:
Okay, made the test more robust by reading the data back in and making sure it is the same.  Note that I discovered a potential bug with the ability to write/read long integers.  The existing test doesn't seem to be using a big enough integer for the array types, so it's not catching this.  I think we can get away with using just the int64 h5 type rather than converting to double and back.  This seems to be leading to a loss of precision as well.  The bug comes in where a short int data type is being used to read in the long integer.  For now, I just used the same short ints in my test to get this feature pushed in and we can open another issue for the bug. 

Also, I did not add a test on the variables called refCNCHAR0, refCNCHAR1, etc.  I'm not sure what those are testing that are different from the other string types, refSTR0, refSTR1, etc.

PR #96, opened 2018-02-18
Added boundCheck function to ArrayUtils, which checks if an index of a

zdodson94:
1D polymorphic array is accessible.

### DIFF
--- a/src/ArrayUtils.f90
+++ b/src/ArrayUtils.f90
@@ -986,7 +986,7 @@ MODULE ArrayUtils
       LOGICAL(SBK) :: res
 
       res=.FALSE.
-      IF(LBOUND(array,DIM=1) <= idx.AND.idx <= UBOUND(array,DIM=1)) res=.TRUE.
+      IF(LBOUND(array,DIM=1) <= idx .AND. idx <= UBOUND(array,DIM=1)) res=.TRUE.
     ENDFUNCTION boundCheck
 !
 ENDMODULE ArrayUtils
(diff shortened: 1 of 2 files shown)


src/ArrayUtils.f90
```diff
@@ -971,4 +972,21 @@ PURE FUNCTION findUpBound_1DReal(r,pos,delta,incl,tol) RESULT(val)
...
 !
+!-------------------------------------------------------------------------------
+!> @brief This function checks if an index is within the allocated domain of a
+!>        1D array of unspecified type. Currently can only be used for 1D arrays
+!>        until assumed-shape polymorphic arrays are supported.
+!> @param array the 1D array to be checked
+!> @param idx the index to check
+!>
+    PURE FUNCTION boundCheck(array,idx) RESULT(res)
+      CLASS(*),INTENT(IN) :: array(:)
```
bkochuna: Did not realize you could use this...

I figured this would get implemented as a type bound generic.

src/ArrayUtils.f90
```diff
@@ -971,4 +972,21 @@ PURE FUNCTION findUpBound_1DReal(r,pos,delta,incl,tol) RESULT(val)
...
+      INTEGER(SIK),INTENT(IN) :: idx
+      !
+      LOGICAL(SBK) :: res
+
+      res=.FALSE.
+      IF(LBOUND(array,DIM=1) <= idx.AND.idx <= UBOUND(array,DIM=1)) res=.TRUE.
```
bkochuna: whitespace around ```.AND.```

zdodson94:
The purpose is to have the ability to check if the scram timeline array is allocated to include the current perturbation (since the core could be scrammed, but the delay could result in the first control rod movement occurring in the next perturbation). It just shortens the source code, and it seems like it could come in handy for other situations.

PR #100, opened 2018-04-10
Renaming DBC to Futility_DBC

bkochuna:
Description:
Also updating dependencies. This is done to avoid
potential namespace collisions when integrating with other
repos/packages since "DBC" may be a common header or module.

CASL Ticket # - #5114


stimpsonsg:
Do you have a comparable branch ready for the rest of MPACT?

bkochuna:
I have a corresponding branch for MPACT here on our gitlab that should be merged at the same time as this, so MPACT don't break.

If this is used in CTF or MAMBA they would need to be updated also.

stimpsonsg:
Yeah, okay...good.  Check with Bob and Ben first...then we'll proceed.

stimpsonsg:
Ben expressed that he'll address it once it's merged; Bob isn't using it...so we're good to go.

I'll approve, but let you merge it so you can time it how you want.

bscollin:
3...2...1... go!!

PR #102, opened 2018-05-01
Split convTol

stimpsonsg:
Splitting convTol into relative and absolute.  Native solvers use absolute, but TPLs (like PETSc) can use both.

### DIFF
--- a/src/LinearSolverTypes.f90
+++ b/src/LinearSolverTypes.f90
@@ -212,8 +212,10 @@ MODULE LinearSolverTypes
     INTEGER(SIK) :: nRestart=30_SIK
     !> Actual iterations performed
     INTEGER(SIK) :: iters=0_SIK
-    !> Tolerance for successful convergence
-    REAL(SRK) :: convTol=1.0E-5_SRK
+    !> Relative tolerance for successful convergence
+    REAL(SRK) :: relConvTol=1.0E-5_SRK
+    !> Absolute tolerance for successful convergence
+    REAL(SRK) :: absConvtol=1.0E-5_SRK
     !> Actual residual converged to
     REAL(SRK) :: residual=0._SRK
     !> Preconditioner to be used by LinearSolverTyope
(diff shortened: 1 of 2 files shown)


unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -1591,20 +1591,20 @@ SUBROUTINE testIterativeOthers()
...
       SELECTTYPE(thisLS); TYPE IS (LinearSolverType_Iterative)
-        CALL thisLS%setConv(-2,-0.1_SRK,-1,-1)
-        CALL thisLS%setConv(-2,1.1_SRK,-1,-1)
+        CALL thisLS%setConv(-2,-0.1_SRK,-0.1_SRK,-1,-1)
+        CALL thisLS%setConv(-2,1.1_SRK,1.1_SRK,-1,-1)
         !Check if default value is used
         bool = thisLS%maxIters == 1000_SIK .AND. thisLS%normType == 2_SIK &
-          .AND. thisLS%convTol == 0.001_SRK .AND. thisLS%nRestart == 30_SIK
+          .AND. thisLS%absConvTol == 0.001_SRK .AND. thisLS%nRestart == 30_SIK
         ASSERT(bool, 'Iterative%setConv(...)')
```
aarograh: Should probably assert that relConvTol got set correctly (and maybe pass in a value different from the absConvTol just to make sure)

src/LinearSolverTypes.f90
```diff
@@ -1515,7 +1527,7 @@ SUBROUTINE solveBiCGSTAB(solver)
         !get L_norm
         CALL LNorm(vr%b,solver%normType,solver%residual)
         !check convergence
-        IF(solver%residual <= solver%convTol) EXIT
+        IF(solver%residual <= solver%absConvTol) EXIT
```
aarograh: So does relConvTol actually get used anywhere?

unit_tests/testLinearSolver/testLinearSolver.f90
aarograh: Already addressed

src/LinearSolverTypes.f90
aarograh: Talked about offline.  Not used yet.

aarograh:
Talked with shane offline.  He has MPACT changes locally to address these changes, but need to check CTF, Mongoose, etc.

The code for this merge looks good.  I'll approve and Shane can merge when the other packages are ready.

stimpsonsg:
Talked with Ben and Bob...we're good to go.

PR #104, opened 2018-05-11
Implement a PETSc-based multigrid solver in Futility

benyee:
In these commits, I have implemented a PETSc-based multigrid solver in Futility.

Summary of changes:
* New LinearSolverTypes_Multigrid.f90 module for the PETSc-based multigrid linear solver
* New MultigridMesh.f90 module.  Filling out a MultigridMeshStructureType object is the easiest way to tell LinearSolverTypes_Multigrid.f90 how to set-up the interpolation and grid structure.
* Added unit tests corresponding to these two modules.
* New pcmg_supplement.c file to create a "SoftReset" function that PETSc was missing for its multigrid linear solver
* I tried to make a custom smoother using PCSHELL in PETSc, but it didn't work very well, so I scrapped it.  However, the commits required for it can be found below in case we ever want to revisit it.
* Some relatively minor changes/improvements were made to a few other files in Futility as well

Although I developed it with MPACT in mind, I tried to make it as general as possible.  Basically, the steps for using LinearSolverTypes_Multigrid are as follows:

1. Call %init as you would with a normal LSTypes
2. Fill out a MultigridMeshStructureType based on your problem's geometric information and your desired interpolation scheme.
3. Call %fillInterpMats, which allocates the necessary PETSc matrices for the interpolation operators and fills them out based on the information you have stored in MultigridMeshStructureType.
4.  Call %setupPETScMG, which defines the smoother on each multigrid level, makes sure each level gets assigned the appropriate interpolation operator, and sets other multigrid-related parameters from a parameter list

I wasn't sure who to request to review this, so if you would like to review it, please go for it!  Thanks!


bkochuna:
Ben, the CI failed for this. Looks like you might need to ifdef some of the code as it relates to PETSc.

```
/mnt/Futility/src/LinearSolverTypes_Multigrid.f90:951:27:
       TYPE(PETScMatrixType),POINTER,INTENT(IN) :: interpMats(:)
                           1
Error: Derived type 'petscmatrixtype' at (1) is being used before it is defined
/mnt/Futility/src/LinearSolverTypes_Multigrid.f90:948:69:
     SUBROUTINE setInterp_LinearSolverType_Multigrid(solver,interpMats)
                                                                     1
Error: Symbol 'interpmats' at (1) has no IMPLICIT type
src/CMakeFiles/Utils.dir/build.make:1430: recipe for target 'src/CMakeFiles/Utils.dir/LinearSolverTypes_Multigrid.f90.o' failed
make[2]: *** [src/CMakeFiles/Utils.dir/LinearSolverTypes_Multigrid.f90.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

benyee:
I just pushed a commit to fix it... it should hopefully work now.

unit_tests/testMultigridMesh/testMultigridMesh.f90
```diff
@@ -0,0 +1,57 @@
...
+  FINALIZE_TEST()
+!
+!===============================================================================
+  CONTAINS
+!
+!-------------------------------------------------------------------------------
+  SUBROUTINE testClear
+    TYPE(MultigridMeshStructureType) :: myMMeshes
+
+    myMMeshes%nLevels=5
```
bkochuna: Should clear also clear the other attributes and should this be checked? e.g. interp matrix?

unit_tests/testMultigridMesh/testMultigridMesh.f90
```diff
@@ -0,0 +1,57 @@
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
```
bkochuna: This is technically sufficient.

However, I'm wondering if it would be worthwhile to provide an example setup of the mesh in here? Maybe a simple structured grid with the Galerkin triple product?

src/LinearSolverTypes_Multigrid.f90
```diff
@@ -0,0 +1,972 @@
...
+!petscisdef.h defines the keyword IS, and it needs to be reset
+#undef IS
+#endif
+!
+! List of public members
+  PUBLIC :: LinearSolverType_Multigrid
+
+  INTEGER(SIK),PARAMETER,PUBLIC :: MAX_MG_LEVELS=8
+
+  !!Public enumerations from LSTypes
```
bkochuna: Extra "!"

unit_tests/testLinearSolver_Multigrid/testLinearSolver_Multigrid.f90
```diff
@@ -0,0 +1,1031 @@
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
```
bkochuna: Does updatedA get its own test?

unit_tests/testLinearSolver_Multigrid/testLinearSolver_Multigrid.f90
```diff
@@ -0,0 +1,1031 @@
...
+  CALL vecPList%add('VectorType -> n',2)
+  CALL vecPList%add('VectorType -> MPI_Comm_ID',mpiTestEnv%comm)
+
+  !Configure exception handler for test
+  CALL e%setStopOnError(.FALSE.)
+  CALL e%setQuietMode(.TRUE.)
+  CALL eParams%addSurrogate(e)
+  CALL eLinearSolverType%addSurrogate(e)
+
+  CREATE_TEST('Test Linear Solvers')
```
bkochuna: Change name here to include multigrid

bkochuna:
@stimpsonsg or @aarograh do either of you want to review before we merge?

It would be good to get a second set of eyes, but Ben Yee leaves next Tuesday, so not a lot of time for him to address comments.

unit_tests/testMultigridMesh/testMultigridMesh.f90
benyee: There is an example of it in testLinearSolverTypes_Multigrid.f90 under SUBROUTINE testFillInterpMats().  I'll add a note about this in testMultigridMesh.f90

src/LinearSolverTypes_Multigrid.f90
benyee: Changed the comment to be more clear

unit_tests/testLinearSolver_Multigrid/testLinearSolver_Multigrid.f90
benyee: Done

unit_tests/testLinearSolver_Multigrid/testLinearSolver_Multigrid.f90
benyee: updatedA is tested in "testSoftReset"; the only important part of the updatedA routine in LS_MG.f90 is the call to the softreset function in pcmg_supplement.c.  The parent updatedA is tested in testUpdatedA of testLinearSolverTypes.f90

unit_tests/testMultigridMesh/testMultigridMesh.f90
benyee: The interpolation matrices are not stored on MultigridMesh.f90.  I added another clear routine to check the clearing on the mesh level rather than the full structure level.

benyee:
I'll have 2 new commits to address the still-failing CI build and Brendan's comments shortly.  I just want to make sure it compiles and runs locally first.

src/LinearSolverTypes.f90
```diff
@@ -1348,6 +1371,105 @@ SUBROUTINE setConv_LinearSolverType_Iterative(solver,normType_in,  &
...
+!> @param rTol Relative convergence criteria
+!> @param absTol Absolute convergence criteria
+!> @param dTol Divergence criteria
+!> @param maxIters The maximum number of iterations to perform
+!> @param nRestart The number of iterations before GMRES restarts
+!>
+!> This subroutine sets the convergence criterion for the iterative solver.
+!>
+    SUBROUTINE setConvPETSc(solver,normType_in,rTol_in,maxIters_in,nRestart_in, &
+                                   absTol_in,dTol_in)
```
aarograh: I could be missing something, but it looks like this interface is basically identical to setConv_LinearSolverType_Iterative and doing a lot of similar things.  Do they really need to be 2 separate methods on the class?  If they do, I'd prefer not having a method that includes a specific TPL name (i.e., replace the PETSC part of setConvPETSC with something that might help help describe why this routine is different from the one immediately above).

My preference would be to just fold this functionality into the old setConv method if you can.

src/LinearSolverTypes_Multigrid.f90
```diff
@@ -0,0 +1,973 @@
...
+!>  - @ref ParameterLists "ParameterLists": @copybrief ParameterLists
+!>  - @ref ParallelEnv "ParallelEnv": @copybrief ParallelEnv
+!>  - @ref VectorTypes "VectorTypes": @copybrief VectorTypes
+!>  - @ref MatrixTypes "MatrixTypes": @copybrief MatrixTypes
+!>  - @ref LinearSolverTypes "LinearSolverTypes": @copybrief LinearSolverTypes
+!>
+!> @par EXAMPLES
+!> @code
+!>
+!> @endcode
```
aarograh: Either fill in or just delete this block.  It's not really accomplishing anything as is.

src/MatrixTypes_PETSc.f90
```diff
@@ -51,6 +51,9 @@ MODULE MatrixTypes_PETSc
 
   TYPE,EXTENDS(DistributedMatrixType) :: PETScMatrixType
     Mat :: A
+
+    !> Number of rows for nonsquare matrices:
+    INTEGER(SIK) :: m
```
aarograh: You call `m` The number of rows here, but you refer to it as the number of columns at lines 101 and 141.  Make those consistent (I assume columns is what you meant)

src/MultigridMesh.f90
```diff
@@ -0,0 +1,418 @@
...
+    !>  operator has no injections.  This is not needed for fillInterpMatrices
+    !>  in LinearSolverTypes_Multigrid, it is only here for the user's
+    !>  convenience and using/setting this variable is optional.
+    INTEGER(SIK) :: finestGridIndex=-1
+    !> For degree 0 points, childIndices is a size-1 array containing the
+    !>  **child** mesh index of the point that it injects into.
+    !> For degree >0 points, childIndices is a size >1 array containing the
+    !>  **current** mesh indices of the points from which the value on this point
+    !>  is interpolated.
+    !> For MPACT, ordering should be WSENBT on finest mesh and WESNBT on
```
aarograh: I think this comment would probably be more appropriate somewhere in MPACT where this is set up, since a Futility use may not know or care anything about MPACT.

Also, was this numbering scheme chosen for a particular purpose?  In MPACT, we enumerate WNESTB.  May not be a big deal, but make sure you're clear wherever this gets used in MPACT or people will break it eventually.

src/MultigridMesh.f90
```diff
@@ -0,0 +1,418 @@
...
+    !>  2 = between 4 1st degree points, should only exist for 2D/3D problems
+    !>  3 = between 6 2nd degree points, should only exist for 3D problems
+    !>  size of mmData(i)%childIndices or mmData(i)%childWeights is
+    !>    MAX(1,2*interpDegrees(i))
+    !>  indices of interpDegrees should be istt:istp
+    !> Ideally, this would be stored in MultigridMeshElementType, but having
+    !>  it stored as an array is really handy for PETSc's dnnz arguments.
+    INTEGER(SIK),ALLOCATABLE :: interpDegrees(:)
+    !TODO Generalize this logic beyond structured Cartesian grids.
+    !  Actually, I think it should work as is for unstructured grids.
```
aarograh: Pick one?  I'm fine with the TODO as long as it's clear what your expectations are.

src/MultigridMesh.f90
```diff
@@ -0,0 +1,418 @@
...
+      myWts%num_eqns=num_eqns
+
+      !TODO turn on option for flat weights to save on memory cost
+      !ALLOCATE(myWts%wts_point(meshes%istt:meshes%istp))
+      !DO i=meshes%istt,meshes%istp
+      !  IF(meshes%interpDegrees(i) > 0) THEN
+      !    ALLOCATE(myWts%wts_point(i)%wts(num_eqns,meshes%interpDegrees(i)*2))
+      !    myWts%wts_point(i)%wts=1.0_SRK/(2*meshes%interpDegrees(i))
+      !  ENDIF
+      !ENDDO
```
aarograh: Do we need to save this code?

benyee:
CI is still not liking something about the #ifdef's with PETSc... I'll look into this a bit more.  (This shouldn't stop any code review though.)

src/pcmg_supplement.c
```diff
@@ -0,0 +1,102 @@
...
+#define PetscToPointer(a) (*(PetscFortranAddr *)(a))
+#define PetscFromPointer(a) (PetscFortranAddr)(a)
+#define PetscRmPointer(a)
+#endif
+
+#if ((PETSC_VERSION_MAJOR<=3) && (PETSC_VERSION_MINOR<=5))
+//This is taken from petsc-3.5/src/ksp/pc/impls/mg/mgimpl.h
+//There is no version of petsc/private/pcmgimpl.h that is includable in older
+//  versions of PETSc, so we just have to manually redefine the typdefs defined
+//  in mgimpl.h here.
```
aarograh: This is more of a question for @bkochuna, @bscollin, or @stimpsonsg, but would it be more appropriate to just not support some of these features if built with an old version of petsc?  It doesn't seem unreasonable to ask people to use a somewhat recent version of libraries if they want the latest features we have.  And duplicating petsc type definitions insides futility isn't ideal.

src/LinearSolverTypes_Multigrid.f90
```diff
@@ -0,0 +1,973 @@
...
+  CONTAINS
+!
+!-------------------------------------------------------------------------------
+!> @brief Initializes the multigrid Linear Solver Type with a parameter list
+!>
+!> @param solver The linear solver to act on
+!> @param Params the parameter list
+!> @param A The A operator in Ax=b
+!>
+    SUBROUTINE init_LinearSolverType_Multigrid(solver,Params,A)
```
aarograh: Seems like some duplicate code here compared with init_LinearSolverType_Base.  Might be worth exploring moving the similar code into a true initBase method.  Probably doesn't have to be done right now though.  I'm not sure of the extent of the duplication.

src/LinearSolverTypes.f90
benyee: That makes sense; I'll merge them into the same subroutine.  When I made this code, it was before Shane's recent commit that changed the standard setConv so that the rel and abs convergence tolerances were split.  Now they're practically the same.

src/LinearSolverTypes_Multigrid.f90
benyee: Good catch, thanks!

src/MatrixTypes_PETSc.f90
benyee: Good catch, thanks!  I did mean columns there

src/MultigridMesh.f90
benyee: I'll move this comment to MPACT.  It was more of a note to myself.  In general, it is arbitrary as long as you are consistent.  From the multigrid linear solver's perspective, it is irrelevant even if you are inconsistent.  Maintaining consistency is mostly to help you avoid errors when filling out multigridmesh levels.

Regarding the convention MPACT, my "positive" x and y directions are EAST and NORTH; I wanted the "negative" directions listed first where possible.  This is really only relevant for CMFD_Std::fillMultigridMeshes.

src/MultigridMesh.f90
benyee: Good catch -- I clarified the comment to describe the situations in which it should work.

src/MultigridMesh.f90
aarograh: Ok.  The reason for the convention was more of a curiosity on my part.  But yeah, moving the comment to MPACT so others know what the convention is and why it was chosen is good.  No need to actually change that convention.

src/MultigridMesh.f90
benyee: I had intended to try using non-uniform interpolation weights, but found them to be unnecessary for the reactor problems we were trying.  My intention was that it should be easy for someone to try to properly re-implement it should the need arise, so I left it there commented out.  However, I suppose there is a log of it in the commits, so we could just delete it.

Would you guys prefer that I delete it, or should I just leave it there?

src/pcmg_supplement.c
benyee: Personally, I think it would be great to stop supporting PETSc 3.5 :)

src/LinearSolverTypes_Multigrid.f90
benyee: There is quite a bit of duplicate code, but I couldn't think of a better way to implement it in the standard LS types without putting in a bunch of long IF-blocks.  The downside is that if someone wants to change something common to both inits, they would have to do it twice.  I probably won't have time to implement it in a more optimal manner before I leave, but, if someone thinks of a good way to refactor this, that would be good!

bkochuna:
@aarograh are you satisfied to merge (pending a successful CI)

aarograh:
I'm good with this once the CI passes.

PR #106, opened 2018-05-21
Adds trueForAny and trueForAll array support to parallelEnv

aarograh:


aarograh:
I fully expect this build to fail with "Error: Found no matching specific binding for the call to the GENERIC 'trueforany'" in testParallelEnv.  I'm probably missing something very simple, but I can't for the life of me spot my mistake in the GENERIC.  If someone could give me a second set of eyes (@stimpsonsg, @bkochuna, @bscollin) that would be helpful.

bkochuna:
@aarograh The problem might be in the call statements. You may need to do:
```
CALL testMPI%trueForAny(10,bool2d(:,1))
```
You're passing a 2D array to a 1D array, got to make the call pass a 1D array.

aarograh:
@bkochuna You're supposed to be able to do that when the dummy argument is declared with (*) instead of (:), right?  We do that all the time for the allReduce calls

bkochuna:
The allreduce methods are not generic interfaces. The dimensionality/rank of the arguments is part of the routine signature, could be why you're running into this with a GENERIC?

aarograh:
@bkochuna I finally found the problem.  You were on the right track.  When resolving a GENERIC interface to determine the specific procedure, each variable's type, kind, and rank is checked.  When assumed-size dummy arguments declared with (*) don't resolve to a particular rank, so you can't use them in generics.  I'll have to rewrite this to have a specific version for 1D, 2D, etc. :/

Thanks for the fresh set of eyes.

aarograh:
Ok @stimpsonsg, this is actually ready for review now.

aarograh:
Ok, just added more tests to cover all the interfaces.

Also made the test parallel... because it apparently wasn't before?  Required a few minor changes to get it passing, but nothing too crazy.

unit_tests/testParallelEnv/testParallelEnv.f90
```diff
@@ -336,13 +392,18 @@ SUBROUTINE testPE_Env()
       ASSERT(.NOT.testPE%world%isInit(),'%isInit()')
 
       COMPONENT_TEST('%initialize(...)')
-      CALL testPE%initialize(PE_COMM_SELF,1,1,1,1)
+#ifdef HAVE_MPI
+      CALL testPE%initialize(PE_COMM_WORLD,2,1,1,1)
+
```
stimpsonsg: Extra line here

PR #107, opened 2018-06-04
Fixes Valgrind memory leak in testCMFDGroupByGroup

ew4:


stimpsonsg:
We're in a holding pattern on merging new things until the RC is tagged, but this will go through once the tag is complete.

PR #109, opened 2018-06-23
Refac reinitmat

shenqicang:
Add `zeroentries` in `Matrixtype` to set all the elements in matrix to be zeros


src/MatrixTypes_Native.f90
```diff
@@ -801,5 +813,40 @@ SUBROUTINE transpose_SparseMatrixType(matrix)
...
     ENDSUBROUTINE transpose_SparseMatrixType
-    
+!
+!-------------------------------------------------------------------------------
+!> @brief zero the matrix
+!> @param matrix declare the matrix type to act on
+!>
+!>
+    SUBROUTINE  zeroentries_SparseMatrixType(matrix)
+      CLASS(SparseMatrixType),INTENT(INOUT) :: matrix
```
bkochuna: Implement and test

src/MatrixTypes_Native.f90
```diff
@@ -801,5 +813,40 @@ SUBROUTINE transpose_SparseMatrixType(matrix)
...
+      CLASS(TriDiagMatrixType),INTENT(INOUT) :: matrix
+    ENDSUBROUTINE zeroentries_TriDiagMatrixType
+!
+!-------------------------------------------------------------------------------
+!> @brief zero the matrix
+!> @param matrix declare the matrix type to act on
+!>
+!>
+    SUBROUTINE  zeroentries_DenseRectMatrixType(matrix)
+      CLASS(DenseRectMatrixType),INTENT(INOUT) :: matrix
```
bkochuna: Implement and test

src/MatrixTypes_Native.f90
```diff
@@ -801,5 +813,40 @@ SUBROUTINE transpose_SparseMatrixType(matrix)
...
+      CLASS(DenseSquareMatrixType),INTENT(INOUT) :: matrix
+    ENDSUBROUTINE zeroentries_DenseSquareMatrixType
+!
+!-------------------------------------------------------------------------------
+!> @brief zero the matrix
+!> @param matrix declare the matrix type to act on
+!>
+!>
+    SUBROUTINE  zeroentries_TriDiagMatrixType(matrix)
+      CLASS(TriDiagMatrixType),INTENT(INOUT) :: matrix
```
bkochuna: Implement and test

src/MatrixTypes_Native.f90
```diff
@@ -801,5 +813,40 @@ SUBROUTINE transpose_SparseMatrixType(matrix)
...
+      CLASS(SparseMatrixType),INTENT(INOUT) :: matrix
+    ENDSUBROUTINE zeroentries_SparseMatrixType
+!
+!-------------------------------------------------------------------------------
+!> @brief zero the matrix
+!> @param matrix declare the matrix type to act on
+!>
+!>
+    SUBROUTINE  zeroentries_DenseSquareMatrixType(matrix)
+      CLASS(DenseSquareMatrixType),INTENT(INOUT) :: matrix
```
bkochuna: Implement and test

src/MatrixTypes_PETSc.f90
```diff
@@ -565,6 +568,20 @@ SUBROUTINE matmult_PETSc(A,B,C,alpha,beta,transA,transB)
       DEALLOCATE(tmpB)
       DEALLOCATE(tmpC)
     ENDSUBROUTINE matmult_PETSc
+!
+!-------------------------------------------------------------------------------
+!> @brief set all the entries of the matrix to be zero
+!> @param matrix declare the matrix type to act on
+!>
+!>
+    SUBROUTINE zeroentries_PETScMatrixType(matrix)
```
bkochuna: Test this routine

src/MatrixTypes_PETSc.f90
```diff
@@ -565,6 +568,20 @@ SUBROUTINE matmult_PETSc(A,B,C,alpha,beta,transA,transB)
...
+      CHARACTER(LEN=*),PARAMETER :: myName='zeroentries_PETScMatrixType'
+      CLASS(PETScMatrixType),INTENT(INOUT) :: matrix
+      PetscErrorCode  :: ierr
+      IF(matrix%isInit) THEN
```
bkochuna: Change to a DBC REQUIRE

Are there any other requirements on the state of the matrix for this? E.g. prealloced, assembled, etc. If so, those need to be checked.

src/MatrixTypes_PETSc.f90
```diff
@@ -565,6 +568,20 @@ SUBROUTINE matmult_PETSc(A,B,C,alpha,beta,transA,transB)
...
+        CALL MatZeroEntries(matrix%a,ierr)
```
bkochuna: The call to this routine should be added to the `testTPLPETSc` unit test to insure the version of petsc that we are using supports this interface.

src/MatrixTypes_Trilinos.f90
```diff
@@ -397,7 +400,15 @@ SUBROUTINE matvec_TrilinosVector(thisMatrix,trans,alpha,x,beta,y,uplo,diag,incx_
...
     ENDSUBROUTINE matvec_TrilinosVector
-    
+!
+!-------------------------------------------------------------------------------
+!> @brief zero all the elements of the matrix
+!> @param matrix declare the matrix type to act on
+!>
+!>
+    SUBROUTINE zeroentries_TrilinosMatrixType(matrix)
+      CLASS(TrilinosMatrixType),INTENT(INOUT) :: matrix
```
bkochuna: Implement and test

stimpsonsg:
Correct me if I'm wrong.  But MatZeroEntries will:
1) Prevent us from needing to reinitialize the matrix/linear system
but will
2) Incur additional linear system memory by keeping all of the nonzeros in the preallocated matrix?

Even if so, it's probably fine, I just want to make sure.  We preallocate it anyway so we can certainly accommodate it, but I think it does trim some memory during the assemble phase.

src/MatrixTypes_Trilinos.f90
shenqicang: I don't believe for now there's a need for implementing this. Since the problem is only with PETSC.

src/MatrixTypes_Native.f90
shenqicang: I don't believe for now there's a need for implementing this. Since the problem is only with PETSC.

src/MatrixTypes_Native.f90
shenqicang: I don't believe for now there's a need for implementing this. Since the problem is only with PETSC.

src/MatrixTypes_Native.f90
shenqicang: I don't believe for now there's a need for implementing this. Since the problem is only with PETSC.

src/MatrixTypes_PETSc.f90
shenqicang: I believe as long as the matrix type is preallocation. Since the initiation of PETSC type finished everything, so there's no need to check preallocation.

src/MatrixTypes_Native.f90
bkochuna: They should all be implemented and tested so there's consistent behavior for clients of the package. They may want to switch matrix types, and if they do, their code won't work, and it will not be obvious.

Implementing and testing these is really important to the overall quality of the package.

src/MatrixTypes_Trilinos.f90
shenqicang: No need for this, trilinos reset the matrix when the matrix is assembled.

src/MatrixTypes_Native.f90
```diff
@@ -696,7 +710,7 @@ SUBROUTINE transpose_DenseSquareMatrixType(matrix)
 !>
 !>
     SUBROUTINE transpose_DenseRectMatrixType(matrix)
-      CHARACTER(LEN=*),PARAMETER :: myName='transpose_DenseSquareMatrixType'
+      CHARACTER(LEN=*),PARAMETER :: myName='transpose_DenseRecMatrixType'
```
bkochuna: Thanks for taking the time to clean up other other code!

bkochuna:
I pushed some minor whitespace changes. Otherwise this looks good to me. Thanks Qicang!

src/MatrixTypes_Trilinos.f90
```diff
@@ -397,7 +400,19 @@ SUBROUTINE matvec_TrilinosVector(thisMatrix,trans,alpha,x,beta,y,uplo,diag,incx_
...
+      CHARACTER(LEN=*),PARAMETER :: myName='matvec_MatrixTypeVectorType'
+      ! No need
+      CALL eMatrixType%raiseFatalError(modName//'::'//myName// &
```
bkochuna: So if Trilinos already zeros the matrix... perhaps we just leave a REQUIRE statement in here so the "contract" for this routine is consistent across all implementations?

src/MatrixTypes_Native.f90
```diff
@@ -801,5 +815,53 @@ SUBROUTINE transpose_SparseMatrixType(matrix)
...
+!> @brief zero the matrix
+!> @param matrix declare the matrix type to act on
+!>
+!>
+    SUBROUTINE  zeroentries_DenseRectMatrixType(matrix)
+      CHARACTER(LEN=*),PARAMETER :: myName='zeroentries_DenseRectMatrixType'
+      CLASS(DenseRectMatrixType),INTENT(INOUT) :: matrix
+      REQUIRE(matrix%isInit)
+      matrix%a=0.0_SRK
+    ENDSUBROUTINE zeroentries_DenseRectMatrixType
```
bkochuna: Question for @stimpsonsg and @aarograh 

Do you think we should include an `ENSURE(ALL(matrix%a == 0.0_SRK))` at the end of these routines?

It seems like it might be kind of redundant for testing, but at the same time, it defines a more explicit contract. I think having more explicit contracts is generally a better thing. Thoughts?

aarograh: I've been sort of indifferent about that kind of thing for the exact combination of reasons you just stated.  I would say that it's probably better to just put the ENSURE there to be completely explicit about the contract (even though I admittedly haven't put similar ENSURE statements in other places), but I'm open to either.

stimpsonsg: Agreed.  It's also a bit more cumbersome to check it from PETSc since it's not as simple as ENSURE(...%a...).  I'd probably leave it off as redundant.

bkochuna: Thanks for the comments guys. Sounds like we're pretty much on the same page.

bkochuna:
I force pushed an update to this branch to remove the debugging commits

PR #110, opened 2018-06-23
Elements isotopes

bscollin:
So there's a stray commit in here basically changing `TYPE` to `CLASS` in linear solvers.

Other than that, there's a new class that allows a user to go between ZAID and the name.  Also allows user to get atomic and mass numbers, and convert between atomic number and element name.


stimpsonsg:
Looks good...that unit test looks so slick.

PR #111, opened 2018-06-29
RELEASE_DEBUG Options

stimpsonsg:
Adding configuration options for release_debug (RELWITHDEBINFO per TriBITS).


cmake/Futility_Configurations.cmake
```diff
@@ -37,6 +37,7 @@ ENDIF()
 SET(CONFIG_TYPES
     DEBUG
     RELEASE
+    RELWITHDEBINFO
```
bscollin: can we just follow CASL RELEASE_DEBUG syntax so I don't have to remember 2 crazy names?

cmake/Futility_Configurations.cmake
```diff
@@ -276,6 +277,11 @@ ELSEIF(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
     SET(Fortran_FLAGS_RELEASE
         ${CSYM}O3
        )
+
```
bscollin: What about other compilers... could you just add ${Fortran_FLAGS_DEBUG} ${Fortran_FLAGS_RELEASE} instead of calling them out... also... debug has bounds_check on??

cmake/Futility_Configurations.cmake
stimpsonsg: Tribit specs

cmake/Futility_Configurations.cmake
stimpsonsg: debug does bounds checking (fcheck=all)

stimpsonsg: added for other compilers

cmake/Futility_Configurations.cmake
bscollin: I'll allow it...

cmake/Futility_Configurations.cmake
bscollin: very nice... note that fbounds-check is a depreciated feature to be replaced with fcheck=bounds.

PR #112, opened 2018-07-05
Bug fix in nFields

bkochuna:
Description:
Quoted strings were truncating the number of fields by 1.

Test updated to test quoted strings. This had been commented out. Quoted
strings also tested for multiples.

### DIFF
--- a/src/IO_Strings.f90
+++ b/src/IO_Strings.f90
@@ -657,13 +657,12 @@ MODULE IO_Strings
         IF(aline(i:i) == "'" .OR. aline(i:i) == '"') THEN
           IF(inQuotes) THEN
             inQuotes=.FALSE.
-            n=n+1
+            n=n+2
             CYCLE
           ELSE
             inQuotes=.TRUE.
           ENDIF
         ENDIF
-
         !Process the spaces and multiplier characters if not in a quoted string
         IF(.NOT.inQuotes) THEN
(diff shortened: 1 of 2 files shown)


bkochuna:
ping @stimpsonsg or @aarograh this has just been sitting for awhile. should be a pretty quick review. If either of you do not have time though I can assign to someone else.

src/IO_Strings.f90
```diff
@@ -684,6 +683,14 @@ PURE FUNCTION nFields_char(aline)  RESULT(nfields)
             READ(aline(i+1:multidcol-1),*,IOSTAT=ioerr) nmult
             IF(ioerr /= 0) nmult=1
             n=n+(nmult-1)*2
+
+            !If we are multiplying a quoted string need to subtract 1.
```
stimpsonsg: When you saying "multiplying a quoted string"...do you mean like 2*"some string"?

bkochuna: yes. exactly.

bkochuna: IF you look farther down near line 229 this test already existed (didn't need to add it). The problem was that the previously when that was made to work it broke the cases that were commented out.

I found an alternative way of implementing the routine such that both sets of ASSERTS could pass.

Make sense?

stimpsonsg: I see...makes sense.

PR #113, opened 2018-07-05
Modification to MemProf to use better edit values

bkochuna:
Description:
Using /proc/self/status VmRSS for the resident set size (RSS) versus
the VmSize parameter. The VmSize parameter will include memory from
shared libraries not explicitly allocated to physical memory or by
the program. This can lead to artificially high values that obfuscate
the actual memory usage.

The maximum memory (high water mark for the resident set size) is
also now utilized in the MemProf.f90 module. Previously it was just
the maximum value of VmSize.

Units are also now correct. Raw measurements in getSysProcInfo
are returned as "kilobytes" for peakUsage and "kB" for the RSS.

Generally:
kilobytes = 1000 bytes (or 1024 bytes, it is ambiguous)
kB        = 1000 bytes (may be used erroneously)
KB        = 1024 bytes (may be used erroneously)
KiB       = 1024 bytes (unambiguous, but not widely used)

The peakUsage is assumed to be 1024 bytes, and the VmRSS is assumed
to be 1000 bytes.

### DIFF
--- a/src/MemProf.f90
+++ b/src/MemProf.f90
@@ -163,12 +163,12 @@ MODULE MemProf
           thisMP%print_banner=.FALSE.
         ENDIF
         CALL thisMP%pe%world%barrier()
-        CALL getProcMemInfo(tmpL1,tmpL2)
-        thisMP%mem_current=REAL(tmpL1,SRK)/(1024.0_SRK*1024.0_SRK)
+        CALL getProcMemInfo(tmpL1,tmpL2) !tmpL1 in kB, and tmpL2 in bytes
+        thisMP%mem_current=REAL(tmpL1,SRK)*1000.0_SRK/(1024.0_SRK*1024.0_SRK*1024.0_SRK)
         loc(1)=thisMP%pe%world%rank
 
         mem=thisMP%mem_current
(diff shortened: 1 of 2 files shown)


src/MemProf.f90
```diff
@@ -163,12 +163,12 @@ SUBROUTINE edit_MemProf(thisMP,name)
           thisMP%print_banner=.FALSE.
         ENDIF
         CALL thisMP%pe%world%barrier()
-        CALL getProcMemInfo(tmpL1,tmpL2)
-        thisMP%mem_current=REAL(tmpL1,SRK)/(1024.0_SRK*1024.0_SRK)
+        CALL getProcMemInfo(tmpL1,tmpL2) !tmpL1 in kB, and tmpL2 in bytes
+        thisMP%mem_current=REAL(tmpL1,SRK)*1000.0_SRK/(1024.0_SRK*1024.0_SRK*1024.0_SRK)
```
bscollin: why is there a factor of 1000/1024?

bkochuna: See the comments in the commit message.
kB = 1000 bytes. I'm converting to KB.

PR #117, opened 2018-07-23
Avogadro's number is changed to 6.022140857e+23 from Source NIST.

jipuwang:
Description:
Avogadro's number is changed to 6.022140857e+23 from web page:
[https://physics.nist.gov/cgi-bin/cuu/Value?na](https://physics.nist.gov/cgi-bin/cuu/Value?na)

CASL Ticket # - #5478

### DIFF
--- a/src/Constants_Conversion.f90
+++ b/src/Constants_Conversion.f90
@@ -42,8 +42,8 @@ MODULE Constants_Conversion
   REAL(SRK),PUBLIC,PARAMETER ::     RPI=ONE/PI
   REAL(SRK),PUBLIC,PARAMETER ::  RTWOPI=ONE/TWOPI
   REAL(SRK),PUBLIC,PARAMETER :: RFOURPI=ONE/FOURPI
-  !> Avogadro's Number
-  REAL(SRK),PUBLIC,PARAMETER :: Na=6.02214129000E+23_SRK !16 digits
+  !> Avogadro's Number; source: https://physics.nist.gov/cgi-bin/cuu/Value?na
+  REAL(SRK),PUBLIC,PARAMETER :: Na=6.022140857E+23_SRK
 
   ! Area conversion
   !> Conversion factor from barns to cm**2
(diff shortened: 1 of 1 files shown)


djabaay:
Thanks Jipu.  We'll probably need to rebaseline all of the MPACT/VERA tests because of this, so it will take a while to merge in.

src/Constants_Conversion.f90
```diff
@@ -42,8 +42,8 @@ MODULE Constants_Conversion
   REAL(SRK),PUBLIC,PARAMETER ::     RPI=ONE/PI
   REAL(SRK),PUBLIC,PARAMETER ::  RTWOPI=ONE/TWOPI
   REAL(SRK),PUBLIC,PARAMETER :: RFOURPI=ONE/FOURPI
-  !> Avogadro's Number
-  REAL(SRK),PUBLIC,PARAMETER :: Na=6.02214129000E+23_SRK !16 digits
+  !> Avogadro's Number; source: https://physics.nist.gov/cgi-bin/cuu/Value?na
+  REAL(SRK),PUBLIC,PARAMETER :: Na=6.022140857E+23_SRK
```
bkochuna: Change to all caps `NA`

bkochuna:
@aarograh and @stimpsonsg 

What are your guys thoughts about adding unit tests for the constants? I'm on the fence about it... guess i'm leaning towards add a unit test for this module, that way if anyone changes one of these values, it becomes immediately clear that things are failing from a change to these values (where otherwise this would not be as clear).

stimpsonsg:
I'm not sure adding a unit test will do much good, honestly.  Devs will just end up changing it in two places and still potentially fail downstream tests.

bkochuna:
Yeah, makes sense... Ok, Jipu, just change capitalization, and this is good to merge.

jipuwang:
Na is all caps now.  @bkochuna

PR #118, opened 2018-07-26
Speedup Driver Compilation

stimpsonsg:
Changes to farm out USE statements to individual subpackages, which should speed up the overall build time, and definitely improves Driver build time.

### DIFF
--- a/src/Futility.f90
+++ b/src/Futility.f90
@@ -1,0 +1,31 @@
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!  This manuscript has been authored by UT-Battelle, LLC, under Contract       !
+!  No. DE-AC0500OR22725 with the U.S. Department of Energy. The United States  !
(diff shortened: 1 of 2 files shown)


src/Futility.f90
```diff
@@ -0,0 +1,31 @@
...
+  USE MemProf
+  USE IO_Strings
+  USE IOutil
+  USE BLAS
+  USE FileType_Fortran
+  USE FileTYpe_Input
+  USE FileType_Log
+  USE ExpTables
+  IMPLICIT NONE
+!
```
aarograh: Not sure we need this ! here.

bscollin: So I'm a little confused on why you picked these... other than they are in driver.  Just doesn't seem very intuitive.

aarograh: Actually, ben's right... I saw modules like BLAS that are already lumped together and thought you covered everything... you should probably list all the fortran modules

stimpsonsg: Okay...I just included the ones Driver used...yep

PR #119, opened 2018-07-27
Add exceptions to MatrixType

rks171:
Integrating the native solvers for sparse matrices would have went more smoothly if Futility would have warned me that I needed to call setShape before setting data and that the j values need to be in ascending order.  Added a few other exceptions that should help prevent unintended behavior.

### DIFF
--- a/src/MatrixTypes_Native.f90
+++ b/src/MatrixTypes_Native.f90
@@ -458,7 +458,16 @@ MODULE MatrixTypes_Native
             ENDIF
           ENDDO
           IF(found_ja) matrix%a(ja_index)=setval
+        ELSE
+          IF(matrix%jCount==0) THEN
+            CALL eMatrixType%raiseError(modName//'::'//myName//' - Matrix entries not defined.  '//&
+               'Ensure that setShape has been called.')
+          ELSEIF(i>matrix%n.or.i<0.or.j<0) THEN
+            CALL eMatrixType%raiseError(modName//'::'//myName//' - Passed row/col out of bounds.')
+          ENDIF
         ENDIF
+      ELSE
+         CALL eMatrixType%raiseError(modName//'::'//myName//' - Matrix not initialized.')
       ENDIF
     ENDSUBROUTINE set_SparseMatrixtype
 !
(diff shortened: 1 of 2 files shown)


rks171:
Just realized I forgot to throw an error when 'get' is called with indices out of bounds.  I'll add and push.

bscollin:
So I'm struggling to decide if these should be DBC or exceptions.  In my view... DBC makes a bit more sense because its a developer error, not a user error.  General thoughts @stimpsonsg @bkochuna ?

stimpsonsg:
So the rest of MatrixType_Native is using exception handling (no DBC yet).  I think you're probably right in that DBC would be more appropriate, but would we rather the change be consistent for now and transition DBC in one push?  I can see value in either approach.

bscollin:
The only reason it wasn't used before is it didn't exist.  I'm more of the opinion is that we should make the correct decision when we refactor and maybe clean some stuff up periodically.

stimpsonsg:
fair enough

rks171:
How do I unit test them if I switch to dbc?  Is there an example somewhere else?

rks171:
#120 added the capability to shut off the stop in DBC, but I don't think I'm going to be able to test most of the DBC.  In cases where I'm REQUIRE'ing the matrix to be initialized or that indices are in bounds, the code is just going to segfault when it tries to access the array data out of bounds after the failed REQUIRE, which then fails the whole unit test.  I'll only be able to test REQUIREs that still let the procedure run through and exit.  With the exceptions, an else block was being triggered, so it wasn't actually hitting the main code in the procedures.

rks171:
It's also useful to note that the developer needs to wrap any unit tests that are testing DBC with #ifdef FUTILITY_DBC so the tests do not get added when DBC is not enabled.  Otherwise the tests would fail because no DBC statements will get hit, but the unit test would still be ASSERTing on the DBC counter.

rks171:
Okay this is ready to be reviewed again.

src/MatrixTypes_Native.f90
```diff
@@ -663,7 +659,7 @@ SUBROUTINE get_DenseSquareMatrixType(matrix,i,j,getval)
 !>
 !> This routine gets the values of the sparse matrix.  If an (i,j) value
 !> is not present, then 0.0 is returned.  If the (i,j) location is out of
-!> bounds, then -1051.0 is returned (-1051.0 is an arbitrarily chosen key).
+!> bounds, then an exception is raised.
```
stimpsonsg: might need to update this comment since exceptions were removed

rks171: Say code crashes if DBC enabled... or just remove the whole sentence and expect the dev not to use out of bounds?

stimpsonsg: I personally prefer the first

rks171: Okay fixed.

stimpsonsg:
So this is failing a bunch of stuff:

image

stimpsonsg:
image

stimpsonsg:
I'm going to back it out for now until it's resolved.

rks171:
Okay I’ll look into it.

On Aug 16, 2018, at 10:07 PM, stimpsonsg <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm going to back it out for now until it's resolved.

PR #123, opened 2018-08-17
Update header doc to clarify a return value

rks171:
Is this okay to make this guarantee?  If I know it's negative 1 when the file wasn't opened, I can use iostat to allow the code to catch and ignore errors when I don't want to write to the file.


stimpsonsg:
I'm fine with that guarantee

PR #127, opened 2018-08-22
Adds deviceID component to ParallelEnvtype

aarograh:
This is intended to store a device ID for offloading calculations
to and accelerator.  The immediate use will be for the GPU system
ID in conjuction with Kokkos, but there could be other uses as well.


aarograh:
Test added.

PR #128, opened 2018-08-31
Have Futility_DBC check if MPI is intialized

rks171:
Description:
Skip calling MPI_get_rank and MPI_get_nproc if not so it doesn't cause a
misleading error.

CASL Ticket # - N/A

### DIFF
--- a/src/Futility_DBC.f90
+++ b/src/Futility_DBC.f90
@@ -74,7 +74,6 @@ MODULE Futility_DBC
     INTEGER :: mpierr
     LOGICAL :: mpiInit
     CALL MPI_Initialized(mpiInit,mpierr)
-    print *, mpiInit
     IF(mpiInit) THEN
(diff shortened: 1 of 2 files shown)


src/Futility_DBC.f90
```diff
@@ -72,8 +72,16 @@ SUBROUTINE DBC_FAIL(test_char,mod_name,line)
     INTEGER :: rank,nproc
 #ifdef HAVE_MPI
     INTEGER :: mpierr
-    CALL MPI_Comm_rank(MPI_COMM_WORLD,rank,mpierr)
-    CALL MPI_Comm_size(MPI_COMM_WORLD,nproc,mpierr)
+    LOGICAL :: mpiInit
+    CALL MPI_Initialized(mpiInit,mpierr)
+    print *, mpiInit
```
bscollin: delete this.

rks171: fixed

PR #131, opened 2018-09-06
Fixing unused variable causing MPACT CI to fail

bscollin:


bscollin:
Talked this over with Shane.  He is okay merging but not by machine.

PR #136, opened 2018-09-14
Adds the ComputingEnvironment module and object

aarograh:
CASL Ticket # - 5607


aarograh:
I'm also expecting that if we leave `timers` on this object, there would probably be `addTimer`, `getTimer`, and `clearTimers` methods.  I didn't bother adding those before we agreed that timers should actually be on this object and we get this merged in.

src/ComputingEnvironment.f90
```diff
@@ -0,0 +1,40 @@
...
+!                          Futility Development Group                          !
+!                             All rights reserved.                             !
+!                                                                              !
+! Futility is a jointly-maintained, open-source project between the University !
+! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
+! can be found in LICENSE.txt in the head directory of this repository.        !
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief Defines the FutilityComputingEnvironment extended type
+!>
+!> This tyep is a simple struct-like object that bundles together commonly used
```
stimpsonsg: type*

src/ComputingEnvironment.f90
```diff
@@ -0,0 +1,40 @@
...
+  USE Times
+  USE FileType_Input
+  USE FileType_XML
+
+  IMPLICIT NONE
+  PRIVATE
+
+  PUBLIC :: FutilityComputingEnvironment
+
+  TYPE :: FutilityComputingEnvironment
```
stimpsonsg: should we call this FutilityComputingEnvironment? or just ComputingEnvironment, as the filename suggests?

src/ComputingEnvironment.f90
```diff
@@ -0,0 +1,40 @@
...
+    CLASS(ParallelEnvType),POINTER :: parEnv => NULL()
+    CLASS(ExceptionHandlerType),POINTER :: exceptHandler => NULL()
+    CLASS(LogFileType),POINTER :: logFile => NULL()
+    CLASS(InputFileType),POINTER :: inputFile => NULL()
+    CLASS(XMLFileType),POINTER :: xmlFile => NULL()
+    CLASS(HDF5FileType),POINTER :: h5outFile => NULL()
+    CLASS(Memory_Profiler),POINTER :: memProf => NULL()
+    CLASS(TimerType),POINTER :: timers(:) => NULL()
```
stimpsonsg: yeah, I'm good with timers being on here

src/ComputingEnvironment.f90
```diff
@@ -0,0 +1,40 @@
...
+
+  PUBLIC :: FutilityComputingEnvironment
+
+  TYPE :: FutilityComputingEnvironment
+    CLASS(ParallelEnvType),POINTER :: parEnv => NULL()
+    CLASS(ExceptionHandlerType),POINTER :: exceptHandler => NULL()
+    CLASS(LogFileType),POINTER :: logFile => NULL()
+    CLASS(InputFileType),POINTER :: inputFile => NULL()
+    CLASS(XMLFileType),POINTER :: xmlFile => NULL()
+    CLASS(HDF5FileType),POINTER :: h5outFile => NULL()
```
stimpsonsg: we need to make it way easier to get stuff to the HDF5 file too...having this here should/could help that

src/ComputingEnvironment.f90
```diff
@@ -0,0 +1,40 @@
...
+  USE FileType_XML
+
+  IMPLICIT NONE
+  PRIVATE
+
+  PUBLIC :: FutilityComputingEnvironment
+
+  TYPE :: FutilityComputingEnvironment
+    CLASS(ParallelEnvType),POINTER :: parEnv => NULL()
+    CLASS(ExceptionHandlerType),POINTER :: exceptHandler => NULL()
```
stimpsonsg: DELETE ALL THE SURROGATES!!!!

stimpsonsg:
I'm thining we add some stuff to this before merging

src/ComputingEnvironment.f90
aarograh: I added Futility to the names of the module and type since they are the names that are usable in higher level code.  Adding this will prevent possible name collisions, like Brendan suggested in the ticket.

If you really want them consistent, I would change the file name over the module/type name.  I personally don't think it matters, but I'm open to changing it if that's what others want.

src/ComputingEnvironment.f90
aarograh: Yes.  This.  Forever.

src/ComputingEnvironment.f90
stimpsonsg: sure...it seems clunky, but I can understand that benefit.  I would really prefer that the filename matched, but that's really mostly a preference.

aarograh:
Talked to ben offline a bit about this and he also suggested that the environment own an array of pointers to other "sub" computing environment objects.  This will allow mpact to set one up for ctf (and eventually other codes, maybe) and pass it off, since we now find ourselves giving ctf several futility objects anyways.

src/ComputingEnvironment.f90
aarograh: Alright, I'll go ahead and change the filename when I address other comments.

src/ComputingEnvironment.f90
aarograh: Fixed

src/ComputingEnvironment.f90
aarograh: Changed

src/ComputingEnvironment.f90
aarograh: I'll add a couple methods to handle adding/removing them as well and push that up

aarograh:
Ok, I pushed updates to address comments so far.  I also went ahead and filled in some get/add/remove/clear methods and added a unit test for all of them.

stimpsonsg:
Well, except that it's failing :-/

PR #138, opened 2018-09-28
Add new getAll procedures for parallel vectors

rks171:
Description:
The original getAll procedure will not work for a parallel vector because it
will cause an out of bounds error in PETSc.  To get all values at once, the user
needs to pass in the global indices of the values to get and they can only
pass  in indices that belong to the proc.  I added a new procedure, 'getSelected',
which can be used to do this, clarified that getAll should not be used for parallel
vectors, added tests for getSelected, and then added a new set of parallel vector
tests.

### DIFF
--- a/src/VectorTypes_Base.f90
+++ b/src/VectorTypes_Base.f90
@@ -85,9 +85,11 @@ MODULE VectorTypes_Base
       PROCEDURE(vector_getOne_absintfc),DEFERRED,PASS :: getOne
       !> Deferred routine for getting all vector values
       PROCEDURE(vector_getAll_absintfc),DEFERRED,PASS :: getAll
+      !> Deferred routine for getting all vector values, entire vector at once
+      PROCEDURE(vector_getSelected_absintfc),DEFERRED,PASS :: getSelected
       !> Deferred routine for getting a range of vector values
       PROCEDURE(vector_getRange_absintfc),DEFERRED,PASS :: getRange
-      GENERIC :: get => getOne,getAll,getRange
+      GENERIC :: get => getOne,getAll,getSelected,getRange
   ENDTYPE VectorType
 !
 !List of Abstract Interfaces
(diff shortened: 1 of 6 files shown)


stimpsonsg:
Looks good.  We should probably add a check to prevent parallel vectors from trying getAll, or have it internally call getSelected and maybe gather the data.

rks171:
I can get an error in there, but won't be able to call getSelected because as far as I know, it needs the global indices of the local data, which the VectorType has no knowledge of.

rks171:
This is something I'm a little confused on with PETSc.  Seems like it is in charge of defining the global indices and it does so by using the process rank in a way that rank 0 owns 1->nlocal, rank 1 owns nlocal+1->2*nlocal, etc. (assuming nlocal are all the same).  If the user sets nlocal=-1 and lets petsc decide, how would you ever know what proc owns what elements so you don't call with a proc that doesn't own the data and cause a seg fault?  Assuming nlocal is set, then Vector types can use knowledge of the MPI proc to figure out the global indexing and do this, but the caller still needs to know the global indices anyway so it knows what to do with the data.

rks171:
For the error, are we sticking with returning error codes or switching to DBC statements?

stimpsonsg:
Yeah, if nlocal isn't set, it would make it nearly impossible to for the vector to know.  Let's pass on that.  I don't think we're using getAll for anything substantial anyway.

Let's add new errors as DBC.

rks171:
Okay we can actually get the global indices using VecGetOwnershipRange.  Then call getRange with this range.  The user needs to ensure they allocated their get array to nlocal for this to work.

stimpsonsg:
Ah, good...makes sense.  I'm good with deferring, but if you want to pursue it...by all means ;-)

rks171:
It's already done in last push

stimpsonsg:
sweet

stimpsonsg:
Some unused variables:

image

stimpsonsg:
Still one more left :-/

image

rks171:
done

stimpsonsg:
We may hold onto this until Monday unless you need it urgently.

rks171:
Monday is fine, but if you're waiting for the 4.0 tag, Ben cleared that already.

stimpsonsg:
Yeah, just want to give it some clearance/breathing room.  Should be fine to merge it now, but Monday doesn't hurt anything, I don't think.

rks171:
Added a setSelected to go with the getSelected.

PR #141, opened 2018-10-06
Hotfix Trilinos Vector

stimpsonsg:
Added set and getSelected for Trilinos vector type (just encounters error).


stimpsonsg:
Yeah, I actually looked to the setAll and getAll routines that were already there, and they just threw errors, so figured I'd be consistent for now.

PR #143, opened 2018-10-19
Update MatrixTypes_Trilinos.f90

rks171:
Fix broken DBC statement.

### DIFF
--- a/src/MatrixTypes_Trilinos.f90
+++ b/src/MatrixTypes_Trilinos.f90
@@ -262,7 +262,7 @@ MODULE MatrixTypes_Trilinos
       REAL(SRK),INTENT(IN) :: setval(:)
 
       REQUIRE(matrix%isInit)
-      REQUIRE(((j <= matrix%n) .AND. (i <= matrix%n)) .AND. ((j > 0) .AND. (i > 0)))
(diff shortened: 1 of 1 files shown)


src/MatrixTypes_Trilinos.f90
```diff
@@ -262,7 +262,7 @@ SUBROUTINE setRow_TrilinosMatrixType(matrix,i,j,setval)
       REAL(SRK),INTENT(IN) :: setval(:)
 
       REQUIRE(matrix%isInit)
-      REQUIRE(((j <= matrix%n) .AND. (i <= matrix%n)) .AND. ((j > 0) .AND. (i > 0)))
+      REQUIRE(((ALL(j <= matrix%n)) .AND. (ALL(i <= matrix%n))) .AND. ((ALL(j > 0)) .AND. (ALL(i > 0))))
```
stimpsonsg: so I think the ALL additions around the i components will be a problem since it's just a scalar value:

```
Error: 'mask' argument of 'all' intrinsic at (1) must be a logical array
```

The ones around the j terms should be fine.  It should roughly match the DBC check in setRow_SparseMatrixType.

rks171: Yep, i is just a scalar.  I fixed it.

PR #147, opened 2018-11-01
Fix valgrind defects in FileType_HDF5

aarograh:


stimpsonsg:
We're going to revert this for now.  There's an issue in overwriting data.  The changes made are good, but it uncovers a potential deficiency in that HDF5 datasets are not resized appropriately when overwritten.  Aaron is going to dig deeper on this before pushing on it again.

In related news, we really need to fix the second build (with HDF5) on the Travis CI.

PR #155, opened 2019-01-11
Remove some "Unused PRIVATE module variable" build warnings

aarograh:


src/MultigridMesh.f90
```diff
@@ -193,9 +193,6 @@ MODULE MultigridMesh
 
   ENDTYPE InterpWeightsStructureType
 
-  !> Exception Handler for use in MatrixTypes
-  TYPE(ExceptionHandlerType),SAVE :: eMultigridMesh
```
bscollin: shouldn't we be more concerned the exception handler is never used?

aarograh: Actually the module uses the `eLinearSolverType` variable a bit.  I expected to agree with you, but turns out there is some error checking, just using a different handler.

PR #157, opened 2019-01-22
Hotfix: GMRES restarting

mattkab2:
While working on a separate branch, we discovered the native GMRES solver does not restart properly. The unit tests did not require a restart and so the issue was never noticed.

This branch updates the unit tests for GMRES, making them require a restart to solve, and also correctly restarts the GMRES algorithm.

### DIFF
--- a/src/LinearSolverTypes.f90
+++ b/src/LinearSolverTypes.f90
@@ -1254,7 +1254,7 @@ MODULE LinearSolverTypes
 !>
     SUBROUTINE setX0_LinearSolverType_Iterative(solver,X0)
       CLASS(LinearSolverType_Iterative),INTENT(INOUT) :: solver
-      REAL(SRK),POINTER,INTENT(IN) :: X0(:)
+      REAL(SRK),INTENT(IN) :: X0(:)
       INTEGER(SIK) :: i
 
       IF(solver%isInit) THEN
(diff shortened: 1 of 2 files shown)


bkochuna:
Review Comments (based on checklist)
1. Tests are present and adequate
2. No stray debug (that was not already there)
3. No comments. I think it could use some.
4. Formatting is good
5. No DBC, but I don't think this change needs any
6. Just a minor low-level fix, no HL design changes
7. I think we want to test this change against MPACT to make sure nothing bad happens. To test it should be a SERIAL build w/o PETSc
8. Commit looks fine
9. Commits will be squashed onto master
10. No other competing changes

src/LinearSolverTypes.f90
```diff
@@ -1598,66 +1598,79 @@ SUBROUTINE solveGMRES_nopc(solver)
...
-        y(:)=0._SRK
-        tol=solver%absConvTol*beta
-
-        v(:,1)=-u%b/beta
-        h=beta
-        phibar=beta
+
+        tol=solver%absConvTol*beta ! Is this correct?
+
+        DO itOuter=1,solver%maxIters
```
bkochuna: Clarify this is the GMRES iterations

src/LinearSolverTypes.f90
```diff
@@ -1576,7 +1576,7 @@ SUBROUTINE solveGMRES_nopc(solver)
       REAL(SRK)  :: beta,h,t,phibar,temp,tol
```
bkochuna: Clarify in routine documentation that subsequent calls restart the GMRES iters.

PR #158, opened 2019-01-24
Adds -Wno-surprising flag to GNU Fortran_FLAGS

aarograh:
This will remove all the spurious warnings from using FINAL procedures.
The warnings that -W-surprising normally covers are for pretty strange,
unusual situations, so we're not really losing any useful information by
disabling this.


aarograh:
This will help clean up the warnings from Bob's merge request, though it looks like there might be some other issues with that one as well.

PR #163, opened 2019-02-01
Implements utility to set a string to lower case

HendersonSC:
Adds the complement function of toUPPER which casts a letter characters of a string to lower case

### DIFF
--- a/unit_tests/testIOutil/testIOutil.f90
+++ b/unit_tests/testIOutil/testIOutil.f90
@@ -362,6 +362,14 @@ PROGRAM testIOutil
       CALL toUPPER(tmpStr)
       ASSERT(TRIM(tmpStr) == '239P84UYQH;JNDF:JKDFH./','gibberish (string)')
 
+      COMPONENT_TEST('toLower')
+      string='239p84uyqh;jndf:JKDFH./'
+      CALL toLower(string)
+      ASSERT(TRIM(string) == '239p84uyqh;jndf:jkdfh./','gibberish (char)')
+      tmpStr='239p84uyqh;jndf:JKDFH./'
+      CALL toLower(tmpStr)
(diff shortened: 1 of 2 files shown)


unit_tests/testIOutil/testIOutil.f90
```diff
@@ -362,6 +362,14 @@ SUBROUTINE testIO_Strings()
...
       ASSERT(TRIM(tmpStr) == '239P84UYQH;JNDF:JKDFH./','gibberish (string)')
 
+      COMPONENT_TEST('toLower')
+      string='239p84uyqh;jndf:JKDFH./'
+      CALL toLower(string)
+      ASSERT(TRIM(string) == '239p84uyqh;jndf:jkdfh./','gibberish (char)')
+      tmpStr='239p84uyqh;jndf:JKDFH./'
+      CALL toLower(tmpStr)
+      ASSERT(TRIM(tmpstr) == '239p84uyqh;jndf:jkdfh./','gibberish (string)')
+
```
aarograh: We really should be using `ASSERT_EQ`... If you want to be my hero you would just do some quick search and replace in this whole test.

Also I don't like that this stuff is in `testIOUtil`, but I see that for some reason there is no `testIO_Strings`, so don't worry about it for the time being.  I might move it later.

HendersonSC: So swapped ASSERT for ASSERT_EQ everywhere that it made ready sense.

stimpsonsg: What inspired this gibberish?

HendersonSC: Ha copy/paste. Looks like @bkochuna inspired the original

PR #169, opened 2019-02-28
Implement Non-linear Solver Class

aarograh:
Adds a new NonLinearSolver module that implements a class for
solving systems of equations.  Currently Newton's method using
native vector, matrix, and linear system objects is all the only supported
functionality, but this can be extended to use other methods or TPLs.

This capability can be used as a simple root-finding procedure or it
could be used to solve non-linear problems.  The client code implements
an extension of the NonLinearSolverInterface_Base object, which provides
methods to evaluate a function and evaluate (or approximate) its
Jacobian.  This function and its Jacobian can be arbitrary so long as they
follow the defined interface.


src/NonLinearSolver.f90
```diff
@@ -0,0 +1,429 @@
...
+  !
+  REAL(SRK),ALLOCATABLE :: bounds(:)
+
+  REQUIRE(.NOT.this%isInit)
+  REQUIRE(ASSOCIATED(f))
+  REQUIRE(plist%has('NonLinearSolver -> n'))
+  REQUIRE(plist%has('NonLinearSolver -> method'))
+  REQUIRE(plist%has('NonLinearSolver -> tolerance'))
+  REQUIRE(plist%has('NonLinearSolver -> bounds'))
+  CALL plist%get('NonLinearSolver -> bounds',bounds)
```
bscollin: Should bounds be required?

src/NonLinearSolver.f90
```diff
@@ -0,0 +1,429 @@
...
+  CLASS(NonLinearSolver_Native),INTENT(INOUT) :: this
+
+  CALL this%clearBase()
+
+ENDSUBROUTINE clear_NonLinearSolverNative
+!
+!-------------------------------------------------------------------------------
+!> @brief Solves a @c NonLinearSolver_Native object
+!> @param this the @c NonLinearSolver_Native object to solve
+!> @param x the @c VectorType containing the starting guess for the solve
```
bscollin: x is also the final solution?

src/ODESolverTypes.f90
```diff
@@ -6,36 +6,14 @@
...
-!>  - @ref VectorTypes "VectorTypes": @copybrief VectorTypes
-!>  - @ref MatrixTypes "MatrixTypes": @copybrief MatrixTypes
-!>  - @ref LinearSolverTypes "LinearSolverTypes": @copybrief LinearSolverTypes
-!>
-!> @par EXAMPLES
-!> @code
-!>
-!> @endcode
-!>
-!> @author Ben Collins
```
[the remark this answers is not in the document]
aarograh: Yes, per the style guide.  Just the Futility header and the description of the module.

unit_tests/testNonLinearSolver/testNonLinearSolver.f90
```diff
@@ -0,0 +1,175 @@
...
+
+CONTAINS
+SUBROUTINE eval_test(this,x,y)
+  CLASS(NonLinearSolverInterface_Test),INTENT(IN) :: this
+  CLASS(VectorType),INTENT(INOUT) :: x
+  CLASS(VectorType),INTENT(INOUT) :: y
+  !
+  REAL(SRK) :: xreal(2)
+
+  CALL x%getAll(xreal)
```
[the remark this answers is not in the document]
aarograh: Good catch.  I'll fix that

src/NonLinearSolver.f90
```diff
@@ -0,0 +1,429 @@
...
+!> this class for its specific applications.
+TYPE,ABSTRACT :: NonLinearSolverInterface_Base
+  PRIVATE
+  CONTAINS
+    !> @copybrief NonLinearSolverModule::clearInterface_Base
+    !> @copydetails NonLinearSolverModule::clearInterface_Base
+    PROCEDURE,PASS :: clear => clearInterface_Base
+    !> @copybrief NonLinearSolverModule::nonlinearsolver_eval_absintfc
+    !> @copydetails NonLinearSolverModule::nonlinearsolver_eval_absintfc
+    PROCEDURE(nonlinearsolver_eval_absintfc),DEFERRED,PASS :: eval
```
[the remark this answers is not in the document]
aarograh: I copied that from `ODESolverTypes` where you used `eval` ;)

src/NonLinearSolver.f90
```diff
@@ -0,0 +1,429 @@
...
+TYPE,ABSTRACT :: NonLinearSolver_Base
+  PRIVATE
+  LOGICAL(SBK),PUBLIC :: isInit=.FALSE.
+  INTEGER(SIK) :: solverMethod=-1
+  INTEGER(SIK) :: TPLType=-1
+  INTEGER(SIK) :: n=-1
+  INTEGER(SIK) :: iterations=-1
+  REAL(SRK) :: tol=-1.0_SRK
+  REAL(SRK) :: bounds(2)=(/HUGE(1.0_SRK),-HUGE(1.0_SRK)/)
+  TYPE(LinearSolverType_Direct) :: linSys
```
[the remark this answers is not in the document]
aarograh: Yeah, you're right.  I'll fix it.  I did that to prototype and forgot to undo it.

src/NonLinearSolver.f90
```diff
@@ -0,0 +1,429 @@
...
+  CLASS(NonLinearSolverInterface_Base),POINTER :: func => NULL()
+  TYPE(FutilityComputingEnvironment),POINTER :: ce => NULL()
+  CONTAINS
+    !> @copybrief NonLinearSolverModule::init_NonLinearSolverBase
+    !> @copydetails NonLinearSolverModule::init_NonLinearSolverBase
+    PROCEDURE,PASS :: init => init_NonLinearSolverBase
+    !> @copybrief NonLinearSolverModule::init_NonLinearSolverBase
+    !> @copydetails NonLinearSolverModule::init_NonLinearSolverBase
+    PROCEDURE,PASS,NON_OVERRIDABLE :: initBase => init_NonLinearSolverBase
```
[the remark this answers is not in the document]
aarograh: Yeah, I think we should use that a little more for abstract classes so people don't do weird things.

src/NonLinearSolver.f90
aarograh: Your suggestion about the function object enforcing the bounds makes a lot of sense, so I'll do that instead.  That will nullify this comment.

src/NonLinearSolver.f90
aarograh: I'll update the comment to reflect that.

aarograh: done

src/NonLinearSolver.f90
aarograh: done

src/NonLinearSolver.f90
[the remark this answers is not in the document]
aarograh: done.  The native solver currently still assumes a direct solver, but it allocates it that way during initialization.  We can add some logic later to allow for iterative solvers if necessary.

unit_tests/testNonLinearSolver/testNonLinearSolver.f90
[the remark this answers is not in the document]
aarograh: done

src/NonLinearSolver.f90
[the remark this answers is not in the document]
bscollin: Do as I say... not as I do...

src/NonLinearSolver.f90
```diff
@@ -0,0 +1,417 @@
...
+  CONTAINS
+    !> @copybrief NonLinearSolverModule::clearInterface_Base
+    !> @copydetails NonLinearSolverModule::clearInterface_Base
+    PROCEDURE,PASS :: clear => clearInterface_Base
+    !> @copybrief NonLinearSolverModule::nonlinearsolver_eval_absintfc
+    !> @copydetails NonLinearSolverModule::nonlinearsolver_eval_absintfc
+    PROCEDURE(nonlinearsolver_eval_absintfc),DEFERRED,PASS :: eval
+    !> @copybrief NonLinearSolverModule::nonlinearsolver_jacobian_absintfc
+    !> @copydetails NonLinearSolverModule::nonlinearsolver_jacobian_absintfc
+    PROCEDURE(nonlinearsolver_jacobian_absintfc),DEFERRED,PASS :: jacobian
```
bscollin: We had discussed writing a base jacobian that estimated it.

aarograh: Moved from Nodal to the base type

src/NonLinearSolver.f90
```diff
@@ -0,0 +1,478 @@
...
+    CLASS(VectorType),INTENT(INOUT) :: y
+  ENDSUBROUTINE nonlinearsolver_eval_absintfc
+
+  !> @brief Interface for the @c jacobian method of a @c NonLinearSolverInterface_Base
+  !>        extension
+  !> @param this the @c NonLinearSolverInterface_Base object
+  !> @param x the @c VectorType to evaluate the Jacobian at
+  !> @param J the @c MatrixType containing the evaluation of the Jacobian at @c x
+  !>
+  SUBROUTINE nonlinearsolver_jacobian_absintfc(this,x,J)
```
bscollin: you don't need the abstract interface anymore

PR #171, opened 2019-03-05
Implement float utility for counting digits left of decimal

HendersonSC:
Looks at the leading digits of a float and determines how many there are. The projected use case is as a helper determining the format required for outputting a given value.


src/IO_Strings.f90
```diff
@@ -1726,4 +1726,37 @@ FUNCTION str_SDK_nDecimal(r,nDecimal) RESULT(string)
...
     ENDFUNCTION str_SDK_nDecimal
 !
+!-------------------------------------------------------------------------------
+!> @brief Returns a character array by repeating a provided sample
+!> 
+  FUNCTION charBanner(charSamp,reps) RESULT(tmpChar)
+    CHARACTER(LEN=*),INTENT(IN) :: charSamp
+    INTEGER(SIK),INTENT(IN) :: reps
+    CHARACTER(LEN=reps*LEN(TRIM(charSamp))) :: tmpChar
+    tmpChar = REPEAT(TRIM(charSamp),reps)
```
aarograh: Definitely had no idea that was a function

aarograh: In fact, what does `charBanner` really do for us that `REPEAT` didn't already accomplish?  Looks like it's basically the exact same interface and functionality...

src/IO_Strings.f90
```diff
@@ -1726,4 +1726,37 @@ FUNCTION str_SDK_nDecimal(r,nDecimal) RESULT(string)
 
     ENDFUNCTION str_SDK_nDecimal
 !
+!-------------------------------------------------------------------------------
+!> @brief Returns a character array by repeating a provided sample
```
aarograh: If we keep this function (see below), need to fill in this banner more with `@param` and `@returns` lines

src/IO_Strings.f90
```diff
@@ -1726,4 +1726,37 @@ FUNCTION str_SDK_nDecimal(r,nDecimal) RESULT(string)
...
+  FUNCTION charBanner(charSamp,reps) RESULT(tmpChar)
+    CHARACTER(LEN=*),INTENT(IN) :: charSamp
+    INTEGER(SIK),INTENT(IN) :: reps
+    CHARACTER(LEN=reps*LEN(TRIM(charSamp))) :: tmpChar
+    tmpChar = REPEAT(TRIM(charSamp),reps)
+  ENDFUNCTION charBanner
+!
+!-------------------------------------------------------------------------------
+!> @brief takes in a real, returns the number of digits to the left of the
+!> decimal.
```
aarograh: `@param` and `@returns` lines

src/IO_Strings.f90
```diff
@@ -1726,4 +1726,37 @@ FUNCTION str_SDK_nDecimal(r,nDecimal) RESULT(string)
...
+!> 
+! Primary intent of this is for writing floats properly, for that reason negative
+! numbers automatically count "-" in the returned length. However a leading zero
+! is not counted.
+  FUNCTION getFloatFormat(inputData) RESULT(lengthOut)
```
aarograh: So the goal is to basically return the number of places left of the decimal, right?  So shouldn't we really try to resolve the leading 0 too?

src/IO_Strings.f90
HendersonSC: Yeah, turns out they are.

src/IO_Strings.f90
HendersonSC: Just getting rid of it. We can use the intrinsic

src/IO_Strings.f90
HendersonSC: Added that.

src/IO_Strings.f90
HendersonSC: added those

unit_tests/testIOutil/testIOutil.f90
```diff
@@ -551,21 +551,21 @@ SUBROUTINE testIO_Strings()
       ASSERT_EQ(str(-5000.0_SDK,5),'-5.00000E+03','str(SDK,nDecimal)')
 
       COMPONENT_TEST('getFloatFormat')
-      ASSERT_EQ(getFloatFormat(0.0_SRK),0_SIK,'zero')
-      ASSERT_EQ(getFloatFormat(-0.0_SRK),0_SIK,'zero')
-      ASSERT_EQ(getFloatFormat(0.1_SRK),0_SIK,'zero')
-      ASSERT_EQ(getFloatFormat(-0.1_SRK),1_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(0.0_SRK),1_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(-0.0_SRK),1_SIK,'zero')
```
aarograh: Hmm, this seems like it could be a problem.  I'm not sure this value would print properly if you told it to use 1 digit left of the decimal.  But I'm not sure the best way of handling it since `-0.0` does show up from time to time...

unit_tests/testIOutil/testIOutil.f90
```diff
@@ -550,6 +550,30 @@ SUBROUTINE testIO_Strings()
...
+      ASSERT_EQ(getFloatFormat(-0.01_SRK),2_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(0.123_SRK),1_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(-0.123_SRK),2_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(1.1234_SRK),1_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(-1.1342_SRK),2_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(12.12_SRK),2_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(-21.12_SRK),3_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(123.12_SRK),3_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(-213.12_SRK),4_SIK,'zero')
+      ASSERT_EQ(getFloatFormat(4213.12_SRK),4_SIK,'zero')
```
aarograh: Also, you have `zero` as the message for all of these... I don't care to much about the message but using `zero` everywhere is wrong.  You could just use `float format` as the message in all of them.  The `ASSERT_EQ` will already print the information we care about

PR #175, opened 2019-03-08
Fixing the intersection between the line type and circle type.

djabaay:
Description:
The bug is when a line segment is near a circle, is non-tangent to the
circle, but if extended would be.  The intersection routine would
produce a point, but ultimately, the intersection point needed more logic
to determine if the intersection point is outside the circle.

CASL Ticket # - PHI 5083

### DIFF
--- a/unit_tests/testGeom_CircCyl/testGeom_CircCyl.f90
+++ b/unit_tests/testGeom_CircCyl/testGeom_CircCyl.f90
@@ -94,6 +94,14 @@ PROGRAM testGeom_CircCyl
       ASSERT(bool, 'circle1%intersectLine(...) (disjoint 1)')
       CALL line1%clear()
 
+      !Test disjoint although extension tagent
+      CALL line1%p1%init(DIM=2,X=0.4225_SRK,Y=0.4_SRK)
+      CALL line1%p2%init(DIM=2,X=0.4225_SRK,Y=0.6_SRK)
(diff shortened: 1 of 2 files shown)


djabaay:
We are working on adding a unit test for this fix.

bkochuna:
Dan incorporated the wrong patch from @shenqicang. So updated the branch (and also updated with master).

I've also asked @shenqicang to add a unit test.

The patch looks good for me (except for lack of unit test)

bkochuna:
Also need to make sure we run against the MPACT test suite to make sure there are no downstream tests affected by this. I'm doing this now, and so far no issues.

djabaay:
Incorporated one of the earlier patches, there were several :)

unit_tests/testGeom_CircCyl/testGeom_CircCyl.f90
```diff
@@ -102,6 +110,7 @@ SUBROUTINE TestCircle_and_Cylinder
       ASSERT(bool, 'circle1%intersectLine(...) (tangent)')
       CALL line1%clear()
 
+
```
bkochuna: Lets fix this. No need to add extra whitespace.

I ran all the heavies under MPACT. 2q failed, but it appeared to be spurious. I reran the test by itself through ctest and it passed. 2q is the case with the grid spacer.

PR #177, opened 2019-03-20
Revert "Revert "Implement float utility for counting digits left of decimal""

stimpsonsg:
Reverts CASL/Futility#173

### DIFF
--- a/src/IO_Strings.f90
+++ b/src/IO_Strings.f90
@@ -107,6 +107,7 @@ MODULE IO_Strings
   PUBLIC :: isChar
   PUBLIC :: isCharCap
   PUBLIC :: isCharLow
+  PUBLIC :: getFloatFormat
   !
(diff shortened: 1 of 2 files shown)


aarograh:
@stimpsonsg @HendersonSC Not sure what this was going to be used for or why it was originally reverted.  Could one of you update?  If it's not needed anymore, let's just close this PR.

HendersonSC:
I don't really recall what was driving this. My guess would be it was a first step toward finding formats for reals in anticipation of a "table writer". Probably only got reverted because we were pausing changes for a release candidate or something (also a guess).

In the grand scheme of things nothing implemented by the original commit is difficult so if it turns out we do need it for some later goal, it will not be hard to re-implement. Unless someone has a use for it right now, I would say just close.

stimpsonsg:
yeah, we reverted because of the tag

PR #178, opened 2019-04-02
Req generator

bkochuna:


bkochuna:
Hey Bob, passed to you for review.

rks171:
We will need to get some tests on this at some point, but we've both been using it so it would be good to get it merged into master.  I'm also not sure the cmake directory is the best place for it to live.

PR #180, opened 2019-05-01
Pl verify list

djabaay:
Adding a modified verify subroutine for development use.  The original verify was only able to be used for unit tests.

### DIFF
--- a/src/ParameterLists.f90
+++ b/src/ParameterLists.f90
@@ -115,6 +115,11 @@ MODULE ParameterLists
   CHARACTER(LEN=*),PARAMETER :: modName='PARAMETERLISTS'
   INTEGER(SIK),PARAMETER :: MAX_1D_LEN=10
 
+  !> Verification enumerations
+  INTEGER(SIK),PARAMETER :: VALIDTYPE_VALIDATE=0
+  INTEGER(SIK),PARAMETER :: VALIDTYPE_VERIFYTEST=1
+  INTEGER(SIK),PARAMETER :: VALIDTYPE_VERIFYLIST=2
+
   !> Exception handler for the module
   TYPE(ExceptionHandlerType),SAVE :: eParams
 
@@ -482,9 +487,12 @@ MODULE ParameterLists
       !> @copybrief ParameterLists::validate_ParamType
       !> @copydoc ParameterLists::validate_ParamType
       PROCEDURE,PASS :: validate => validate_ParamType
-      !> @copybrief ParameterLists::verify_ParamType
-      !> @copydoc ParameterLists::verify_ParamType
-      PROCEDURE,PASS :: verify => verify_ParamType
+      !> @copybrief ParameterLists::verifyTest_ParamType
+      !> @copydoc ParameterLists::verifyTest_ParamType
+      PROCEDURE,PASS :: verify => verifyTest_ParamType
+      !> @copybrief ParameterLists::verifyTest_ParamType
+      !> @copydoc ParameterLists::verifyTest_ParamType
+      PROCEDURE,PASS :: verifyList => verifyList_ParamType
       !> @copybrief ParameterLists::edit_ParamType
       !> @copydoc ParameterLists::edit_ParamType
       PROCEDURE,PASS :: edit => edit_ParamType
@@ -2395,14 +2403,15 @@ MODULE ParameterLists
 !> @returns isValid logical indicating that all the required parameters exist
 !>          in @c thisParam and are of the correct type.
 !>
-    RECURSIVE FUNCTION validateReq_ParamType(thisParam,reqParams,prefix,isMatch) &
+    RECURSIVE FUNCTION validateReq_ParamType(thisParam,reqParams,prefix,validType,isMatch) &
       RESULT(isValid)
       CHARACTER(LEN=*),PARAMETER :: myName='validateReq_ParamType'
       CLASS(ParamType),INTENT(INOUT) :: thisParam
       CLASS(ParamType),INTENT(IN) :: reqParams
       CHARACTER(LEN=*),INTENT(IN) :: prefix
-      LOGICAL(SBK),INTENT(OUT),OPTIONAL :: isMatch
-      LOGICAL(SBK) :: isValid
+      INTEGER(SIK),INTENT(IN) :: validType
+      LOGICAL(SBK),INTENT(OUT) :: isMatch
+      LOGICAL(SBK) :: isValid,tmpbool
       INTEGER(SIK) :: i,ntrue
       CLASS(ParamType),POINTER :: tmpParam
 
@@ -2413,40 +2422,55 @@ MODULE ParameterLists
       SELECTTYPE(p=>reqParams)
         TYPE IS(ParamType)
           !Call validate on the required parameter's value
-          IF(PRESENT(isMatch)) THEN
+          IF((validType == VALIDTYPE_VERIFYTEST) .OR. &
+              (validType == VALIDTYPE_VERIFYLIST)) THEN
             IF(ASSOCIATED(p%pdat)) &
-              isValid=validateReq_ParamType(thisParam,p%pdat,prefix,isMatch)
+              isValid=validateReq_ParamType(thisParam,p%pdat,prefix,validType,isMatch)
           ELSE
             IF(ASSOCIATED(p%pdat)) &
-              isValid=validateReq_ParamType(thisParam,p%pdat,prefix)
+              isValid=validateReq_ParamType(thisParam,p%pdat,prefix,validType,tmpbool)
           ENDIF
         TYPE IS(ParamType_List)
           !Loop over all parameters in the list and check each
           IF(ALLOCATED(p%pList)) THEN
             ntrue=0
             DO i=1,SIZE(p%pList)
-              IF(PRESENT(isMatch)) THEN
+              IF((validType == VALIDTYPE_VERIFYTEST) .OR. &
+                  (validType == VALIDTYPE_VERIFYLIST)) THEN
                 IF(validateReq_ParamType(thisParam,p%pList(i), &
-                  prefix//p%name//'->',isMatch)) ntrue=ntrue+1
+                  prefix//p%name//'->',validType,isMatch)) ntrue=ntrue+1
               ELSE
                 IF(validateReq_ParamType(thisParam,p%pList(i), &
-                  prefix//p%name//'->')) ntrue=ntrue+1
+                  prefix//p%name//'->',validType,tmpbool)) ntrue=ntrue+1
               ENDIF
             ENDDO
-            IF(ntrue == SIZE(p%pList)) isValid=.TRUE.
+            isValid=(ntrue == SIZE(p%pList))
           ELSE
             !The required list is not allocated, which means we do not
             !check any of it's possible subparameters, but we must at least
             !check that the list exists
             CALL thisParam%getParam(prefix//p%name,tmpParam)
             IF(.NOT.ASSOCIATED(tmpParam)) THEN
-              CALL eParams%raiseError(modName//'::'//myName// &
-                ' - Failed to locate required parameter "'//prefix// &
-                  p%name//'"!')
+              SELECTCASE(validType)
+                CASE(VALIDTYPE_VERIFYLIST,VALIDTYPE_VERIFYTEST)
+                  isMatch=.FALSE.
+                  CALL eParams%raiseError(modName//'::'//myName// &
+                      ' - When verifying that parameters are equal, the parameter "'// &
+                      prefix//p%name//'" was not found on both lists!')
+                CASE DEFAULT
+                  CALL eParams%raiseError(modName//'::'//myName// &
+                      ' - Failed to locate required parameter "'//prefix// &
+                      p%name//'"!')
+              ENDSELECT
             ELSE
               IF(SAME_TYPE_AS(tmpParam,p)) THEN
                 isValid=.TRUE.
-                IF(PRESENT(isMatch)) isMatch=match_ParamType(tmpParam,p,prefix)
+                SELECTCASE(validType)
+                  CASE(VALIDTYPE_VERIFYTEST)
+                    isMatch=isMatch .AND. matchTest_ParamType(tmpParam,p,prefix)
+                  CASE(VALIDTYPE_VERIFYLIST)
+                    isMatch=isMatch .AND. matchList_ParamType(tmpParam,p,prefix)
+                ENDSELECT
               ELSE
                 CALL eParams%raiseError(modName//'::'//myName// &
                   ' - Required parameter "'//prefix//p%name//'" has type "'// &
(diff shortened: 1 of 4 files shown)


djabaay:
Don't approve this just yet since it will break testXMLProc and testInputProc.  I can make a merge request to modify just those tests anticipating this merge.

src/ParameterLists.f90
```diff
@@ -2413,40 +2422,55 @@ RECURSIVE FUNCTION validateReq_ParamType(thisParam,reqParams,prefix,isMatch) &
...
+              ENDSELECT
             ELSE
               IF(SAME_TYPE_AS(tmpParam,p)) THEN
                 isValid=.TRUE.
-                IF(PRESENT(isMatch)) isMatch=match_ParamType(tmpParam,p,prefix)
+                SELECTCASE(validType)
+                  CASE(VALIDTYPE_VERIFYTEST)
+                    isMatch=isMatch .AND. matchTest_ParamType(tmpParam,p,prefix)
+                  CASE(VALIDTYPE_VERIFYLIST)
+                    isMatch=isMatch .AND. matchList_ParamType(tmpParam,p,prefix)
```
bkochuna: This can be refactored as:
```
IF(validType == VALIDTYPE_VERIFYTEST .OR. validType == VALIDTYPE_VERIFYLIST) THEN
```
I don't think  you need a select construct here
?

src/ParameterLists.f90
```diff
@@ -3062,7 +3127,319 @@ FUNCTION match_ParamType(thisParam,thatParam,prefix) RESULT(bool)
         CLASS DEFAULT
           CONTINUE
       ENDSELECT
-    ENDFUNCTION match_ParamType
+    ENDFUNCTION matchTest_ParamType
+!
+!-------------------------------------------------------------------------------
+!> @brief This function assumes that thisParam and thatParam are of the same
```
bkochuna: Split this into a more brief description and then `Implementation Notes` where you should list assumptions about the routine.

src/ParameterLists.f90
```diff
@@ -2688,13 +2726,39 @@ SUBROUTINE verify_Paramtype(thisParam,reqParams,isMatch)
...
-        isValid=validateReq_ParamType(thisParam,reqParams,'',isMatch)
+        isValid=validateReq_ParamType(thisParam,reqParams,'',VALIDTYPE_VERIFYTEST,isMatch)
       ELSE
         isMatch=.NOT.ASSOCIATED(thisParam%pdat)
       ENDIF
-    ENDSUBROUTINE verify_Paramtype
+    ENDSUBROUTINE verifyTest_Paramtype
+!
+!-------------------------------------------------------------------------------
+!> @brief Verify should only be used in a unit test setting.  It is for checking
```
bkochuna: Again split this comment.

bkochuna: Clarify if this routine checks for consistent entries or consistent entries _and_ values of data for each entry

src/ParameterLists.f90
bkochuna: So this one checks list structure and values? Not just structure?

src/ParameterLists.f90
djabaay: The selecttype is there because we need to call a different `match` subroutine depending on whether we are in a unit test (_VERIFYTEST) or in normal code execution (_VERIFYLIST).

src/ParameterLists.f90
djabaay: I'll make those mods.  Yes, structure and values are checked.

src/ParameterLists.f90
djabaay: Roger.

djabaay: Updated.

src/ParameterLists.f90
djabaay: Updated.

djabaay:
Pinging @bkochuna to resolve the discussions when he gets a chance.

src/ParameterLists.f90
```diff
@@ -3062,7 +3129,321 @@ FUNCTION match_ParamType(thisParam,thatParam,prefix) RESULT(bool)
...
+!> @param thatParam  The parameter list being checked against
+!> @param bool The logical result of the checked parameters.
+!>
+!> The assumptions of this routine are that the parameters passed in are the 
+!> same extended ParamType.  It also assumes that there is a "gettable" value
+!> that is of thisParam%name on the ParamType.  This function determines
+!> the extended type, then "gets" the appropriate parameter from both lists,
+!> then checks their equivalence.  If they are equal or approximately equal, 
+!> the function results in true.  If not, false.  An error is reported if the 
+!> comparison fails.
```
bkochuna: Do we want this to throw an error or is it sufficient to just return false? 

Seems like redundant behavior. If we ALWAYS want this to throw an error on a non-matching condition, then it should not have a return value. If we want it to throw an error sometimes then that is the prerogative of the client code, and that should not be in this routine.

djabaay: I see your point.  One reason I'd argue for keeping both is that this function is called on the parameter level.  If we are comparing a list of parameters, you'd probably like to know all the ones that failed to match on the list at once.  The alternative would be fixing each parameter in serial, with successive reruns of the PL.

 Would it be better to temporarily disable the stop on error for the PL exception handler, and re-enable upon completion of the `verifyList_Paramtype` call?  This more closely matches the `verifyTest_Paramtype`, where all of the parameters in the list are checked with assertions for two given PLs.

bkochuna: I would not modify the exception handler because this dissociates the surrogate. (This is the problem with messages getting reported to the log file, or screen or both in a way that is inconsistent).

But sounds like motivation for the error messaging is for debugging only. In that case, you might ifdef or create an optional input arg to modify the exception handler to print all the messages, when a developer wants to do debugging. 

I think the expectation is that `thisParam` and `thatParam` are going to be sane structures, but in testing maybe we want the values to be different. If we are making changes that affect the list structure then we want a strong error so the programmer knows something is up. Can you DBC-ify it? I think that's what we're looking for here.

djabaay: I'll try to explain the functionality better, because I think I'm not doing it well.  Take 2!

The previous `verify` procedure was only to be called for unit tests.  It would compare one parameter list to a reference list and ensure that the structure **and** the values were the same.  At every point, there were assertions to check these values and flag the errors for missing parameters or unequal values.

This procedure was attempted to be used during normal code execution, not a unit test.  It did not work, but the same functionality was needed to compare a parameter structure **and** the values.  If I just needed to compare the structure of a parameter list, I would have used the `validate` procedure.

The desire for having the `verifyList` procedure report _all_ the mismatches is to save the user time updating a single parameter to pass, only to have another fail in the same parameter list.  This behavior is analogous to what we do in XMLProc.  We report all of the errors, and let the user fix all of them, so they do not need to successively rerun cases to fix all of the errors.  The specific case for this is when the input assembly geometry does not match the assembly geometry in the restart file, and there are multiple values that are incorrect.  Does that part make sense?

If we don't want to toggle the `%stoponerror` flag, what would be another good way to report all of the mismatches when calling the `verifyList` procedure?  I'm open to other suggestions if that option screws up the surrogate exception association.

djabaay: Ok, so talked with Brendan about how to handle this, and we came to the conclusion to pass in an exception handler to the `%verifyList` routine.  The problem I ran into when doing this is that when I reassign the original conditions of `eParams` back, the counters all get reset in the unit test, which caused assertions to fail.  There also isn't a way to explicitly set the `%counter` attribute.

The ultimate issue is that there isn't a way to assert that the number of errors matches what was expected in the parameterlist unit test without adding another feature on the exception handler.  The overloaded assignment operation copies the `%counter` array.  We either need a `%setCounter` function, or a special type of assignment function, which I think is unnecessary.  Currently, with my patch to futility and MPACT, I can get the desired behavior in XMLProc, because the `eXML` doesn't stop on errors, and I just change the verbosity on the `eParams` exception handler outside of the calls to `%verifyList`.  The actual feature I need to use in XMLProc is just suppress the parameter list error messages for a call, which I already can do.

djabaay: Updated the code @bkochuna.  Let me know your thoughts.

djabaay: Pinging @bkochuna again.

src/ParameterLists.f90
```diff
@@ -2743,21 +2743,29 @@ SUBROUTINE verifyTest_Paramtype(thisParam,reqParams,isMatch)
...
+      CLASS(ExceptionHandlerType),INTENT(IN) :: e
       LOGICAL(SBK),INTENT(OUT) :: isMatch
       LOGICAL(SBK) :: isValid
+      INTEGER(SIK) :: tmpcount(EXCEPTION_SIZE)
+      TYPE(ExceptionHandlerType) :: tmpe
 
       !Assume the list is valid, check it only if the required parameter
       !list is not empty.
+      tmpe=eParams
+      eParams=e
```
bkochuna: I don't think we should modify `eParams` anywhere in this routine.

src/ParameterLists.f90
```diff
@@ -2743,21 +2743,29 @@ SUBROUTINE verifyTest_Paramtype(thisParam,reqParams,isMatch)
 !> @param reqParams
 !> @param isMatch
 !>
-    SUBROUTINE verifyList_Paramtype(thisParam,reqParams,isMatch)
+    SUBROUTINE verifyList_Paramtype(thisParam,reqParams,e,isMatch)
       CLASS(ParamType),INTENT(INOUT) :: thisParam
       CLASS(ParamType),INTENT(IN) :: reqParams
+      CLASS(ExceptionHandlerType),INTENT(IN) :: e
```
bkochuna: Make this inout

src/ParameterLists.f90
```diff
@@ -2743,21 +2743,29 @@ SUBROUTINE verifyTest_Paramtype(thisParam,reqParams,isMatch)
...
       LOGICAL(SBK),INTENT(OUT) :: isMatch
       LOGICAL(SBK) :: isValid
+      INTEGER(SIK) :: tmpcount(EXCEPTION_SIZE)
+      TYPE(ExceptionHandlerType) :: tmpe
```
bkochuna: Don't think you need this if you modify the dummy arg `e`

bkochuna:
The client/calling routine should define or use the necessary error handler for the `verifyList` routine.

`verifyList` then modifies the passed `ExceptionHandler`.

The client code then checks the passed the state of the passed `ExceptionHandler` upon return from `verifyList`.

This eliminates the need for `setCounter`. Sorry I didn't realize you didn't have `INOUT` on the argument yesterday.

src/ParameterLists.f90
djabaay: Then why are we passing in an exception handler?

bkochuna: So we don't have to modify `eParams` and we can still count/track errors. `e` should be properly setup outside the call to `verifyLists` by the client.

djabaay:
I want to take a step back a second, because we seem to be making this more complicated than it should be.

Why are we adding the exception handler in the first place?  We already have one for the module that is public and usable, so why is accessing that one insufficient for what we're doing here?

I have verified that using the module exception handler works with the downstream code I have written in XMLProc for processing the restart and input geometry parameter list `%verifyList` call.  So, if that works, why are we complicating the interface?

bkochuna:
My understanding from our previous discussions was that you had to
customize the behavior of the exception handler to be used by the
verifyLists routine so that all errors were reported and output to the
appropriate destination(s).

To do this, the options were modify eParams or modify eXML. These are
global variables and therefore have a weak dependency. So if we modify
either we risk unintended consequences because other parts of the code
depend on their state.

My proposed solution was to define a new exception handler in the client
with the desired configuration (some modification to eParams or eXML) and
pass this to verifyList so it could register errors safely.

If you get the desired behavior by passing eXML to verifyLists that’s
great. The unit test could also pass eParams.

Make sense?

djabaay:
Got it, thanks for a better clarification and bearing with me.

djabaay:
Updated the interface and the test.

djabaay:
@bkochuna When the MPACT HOTFIX_verifytests goes in, you can pull the trigger on this merge if you're satisfied with the updates.

src/ParameterLists.f90
```diff
@@ -2739,33 +2776,28 @@ SUBROUTINE verifyTest_Paramtype(thisParam,reqParams,isMatch)
...
+!> @param thisParam The parameter list on which to verify the values
+!> @param reqParams The reference parameter list and values
+!> @param e The exception handler to pass
+!> @param isMatch The logical if all parameter names and values are the same
 !>
     SUBROUTINE verifyList_Paramtype(thisParam,reqParams,e,isMatch)
       CLASS(ParamType),INTENT(INOUT) :: thisParam
       CLASS(ParamType),INTENT(IN) :: reqParams
-      CLASS(ExceptionHandlerType),INTENT(IN) :: e
+      CLASS(ExceptionHandlerType),INTENT(INOUT) :: e
```
bkochuna: I'm bothered that the order of `e` and `isMatch` are reversed between `verifyList_Paramtype` and `validateReq_ParamType` seems like one more way the programmer can trip over their own feet.

src/ParameterLists.f90
```diff
@@ -2403,14 +2403,15 @@ RECURSIVE SUBROUTINE clear_ParamType(thisParam)
 !> @returns isValid logical indicating that all the required parameters exist
 !>          in @c thisParam and are of the correct type.
 !>
-    RECURSIVE FUNCTION validateReq_ParamType(thisParam,reqParams,prefix,validType,isMatch) &
-      RESULT(isValid)
+    RECURSIVE FUNCTION validateReq_ParamType(thisParam,reqParams,prefix,validType, &
+        isMatch,e) RESULT(isValid)
```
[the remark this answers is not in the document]
djabaay: Nope, I can do that.

src/ParameterLists.f90
```diff
@@ -2454,13 +2455,25 @@ RECURSIVE FUNCTION validateReq_ParamType(thisParam,reqParams,prefix,validType,is
...
                   isMatch=.FALSE.
-                  CALL eParams%raiseError(modName//'::'//myName// &
-                      ' - When verifying that parameters are equal, the parameter "'// &
-                      prefix//p%name//'" was not found on both lists!')
+                  IF(PRESENT(e)) THEN
+                    CALL e%raiseError(modName//'::'//myName// &
+                        ' - When verifying that parameters are equal, the parameter "'// &
+                        prefix//p%name//'" was not found on both lists!')
+                  ELSE
+                    CALL eParams%raiseError(modName//'::'//myName// &
```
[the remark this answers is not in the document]
djabaay: So it is ultimately because the `validateReq` subroutine is used in various fashion by 3 different wrapping subroutines.  One validates the parameter lists like how we do for the various lists setup in the Reactor subpackage.  One is for verifying lists in unit tests, and this one is for verifying lists in executable code.  I suppose I can enforce the exception handler to be passed through the `validateReq` calls and just have it pass in `eParams` for those other two wrapping subroutines I mentioned.  I'll do that a second.

src/ParameterLists.f90
djabaay: Done.  Good point.

djabaay:
Ok, addressed all the other points.

PR #181, opened 2019-05-09
Added %isSector, %getInnerRadius, and %getOuterRadius to PolygonType.

zdodson94:


zdodson94:
Anymore changes needed here?

djabaay:
Superficial change, but move the getRadius function down to the getInnerRadius and getOuterRadius.

src/Geom_Poly.f90
```diff
@@ -362,19 +374,140 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
...
+        !   ENDDO
+        ! ENDIF
+      ENDIF
+    ENDFUNCTION isSector_PolygonType
+!
+!-------------------------------------------------------------------------------
+!> @brief Determines whether a polygon meets the criteria for being an circle
+!>        section. In short, the polygon must have exactly 1 quadratic edge
+!>        and exactly two non-quadratic edges. A full circle is a section.
+!> @param thisPoly The polygon type to query
```
djabaay: Change the comment to be "two non-quadratic edges, or be a full circle".  A little clearer.

zdodson94: Done.

src/Geom_Poly.f90
```diff
@@ -361,20 +373,141 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
...
+    ELEMENTAL FUNCTION getOuterRadius_PolygonType(thisPoly) RESULT(r)
+      CLASS(PolygonType),INTENT(IN) :: thisPoly
+      REAL(SRK) :: r
+      INTEGER(SIK) :: i
+      r=0.0_SRK
+      IF(thisPoly%isSection()) THEN
+        r=thisPoly%getRadius()
+      ELSEIF(thisPoly%isSector()) THEN
+        DO i=1,thisPoly%nQuadEdge
+          IF(thisPoly%quadEdge(3,i) > r) r=thisPoly%quadEdge(3,i)
```
bkochuna: Exit?

src/Geom_Poly.f90
```diff
@@ -361,20 +373,141 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
...
+      CLASS(PolygonType),INTENT(IN) :: thisPoly
+      REAL(SRK) :: r
+      INTEGER(SIK) :: i
+      r=0.0_SRK
+      IF(thisPoly%isSection()) THEN
+        r=0.0_SRK
+      ELSEIF(thisPoly%isSector()) THEN
+        r=HUGE(1.0_SRK)
+        DO i=1,thisPoly%nQuadEdge
+          IF(thisPoly%quadEdge(3,i) < r) r=thisPoly%quadEdge(3,i)
```
bkochuna: Could you use an EXIT here?

src/Geom_Poly.f90
```diff
@@ -361,20 +373,141 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
...
+!> @param thisPoly The polygon type to query
+!> @param bool The logical result of the operation
+!>
+    ELEMENTAL FUNCTION isSection_PolygonType(thisPoly) RESULT(bool)
+      CLASS(PolygonType),INTENT(IN) :: thisPoly
+      LOGICAL(SBK) :: bool
+      INTEGER(SIK) :: nNonQuadEdges!,i
+      ! REAL(SRK) :: dist,dist_old
+      bool=.FALSE.
+      IF(thisPoly%isinit) THEN
```
bkochuna: DBC?

src/Geom_Poly.f90
```diff
@@ -361,20 +373,141 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
...
+        !         dist_old=dist
+        !       ENDIF
+        !     ENDIF
+        !   ENDDO
+        ! ENDIF
+      ENDIF
+    ENDFUNCTION isSector_PolygonType
+!
+!-------------------------------------------------------------------------------
+!> @brief Determines whether a polygon meets the criteria for being an circle
```
bkochuna: a circle

src/Geom_Poly.f90
```diff
@@ -361,20 +373,141 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
...
+    ELEMENTAL FUNCTION isSector_PolygonType(thisPoly) RESULT(bool)
+      CLASS(PolygonType),INTENT(IN) :: thisPoly
+      LOGICAL(SBK) :: bool
+      INTEGER(SIK) :: nNonQuadEdges!,i
+      ! REAL(SRK) :: dist,dist_old
+      bool=.FALSE.
+      IF(thisPoly%isinit) THEN
+        !nVert should be equal to the total number of edges
+        nNonQuadEdges=thisPoly%nVert-thisPoly%nQuadEdge
+        bool=(thisPoly%nQuadEdge == 2) .AND. (nNonQuadEdges == 2)
```
bkochuna: If above is DBC will this execute w/o segfault if the object is not initialized

src/Geom_Poly.f90
```diff
@@ -361,20 +373,141 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
...
+!> @param thisPoly The polygon type to query
+!> @param bool The logical result of the operation
+!>
+    ELEMENTAL FUNCTION isSector_PolygonType(thisPoly) RESULT(bool)
+      CLASS(PolygonType),INTENT(IN) :: thisPoly
+      LOGICAL(SBK) :: bool
+      INTEGER(SIK) :: nNonQuadEdges!,i
+      ! REAL(SRK) :: dist,dist_old
+      bool=.FALSE.
+      IF(thisPoly%isinit) THEN
```
bkochuna: Can this be a DBC?

src/Geom_Poly.f90
```diff
@@ -361,20 +373,141 @@ ELEMENTAL FUNCTION isCircle_PolygonType(thisPoly) RESULT(bool)
     ENDFUNCTION isCircle_PolygonType
 !
 !-------------------------------------------------------------------------------
+!> @brief Determines whether a polygon meets the criteria for being an arc
+!>        sector. In short, the polygon must have exactly two quadratic edges
+!>        and exactly two non-quadratic edges. A circle is not an arc sector.
+!> @param thisPoly The polygon type to query
+!> @param bool The logical result of the operation
+!>
+    ELEMENTAL FUNCTION isSector_PolygonType(thisPoly) RESULT(bool)
```
bkochuna: Why is this elemental?

bkochuna:
One final general comment but are there any corner cases to be considered?

src/Geom_Poly.f90
bkochuna: Just want to raise the point that this is a comparison of floats. I think with the `HUGE(1.0_SRK)` you should be ok though.

HendersonSC: So to me it looks like the intent is to get the minimum of `%quadEdge(3,:)` if this `%isSector` is true...If that is the case, couldn't this just be...
``` 
  if(this%isSector()) then
    r=minval(this%quadEdge(3,:))
  else
    r=0.0_SRK
  endif
```
or I think we could even go nuts and do...
```
r = merge(minval(this%quadedge(3,:)),0.0_SRK,this%isSector())
```

src/Geom_Poly.f90
zdodson94: I'm changing this to use `MAXVAL`, but I'm finding the max so I can't `EXIT`.

src/Geom_Poly.f90
zdodson94: Great suggestion, I will use it instead.

src/Geom_Poly.f90
zdodson94: Yes, I will try `REQUIRE`ing that the poly is initialized and hopefully it doesn't break other things. Other checking routines like `%isCircle` did not have it so I didn't originally have DBC.

src/Geom_Poly.f90
zdodson94: Yes, it should execute but I will try adding DBC because it should be a requirement that it is initialized.

src/Geom_Poly.f90
zdodson94: See above.

src/Geom_Poly.f90
zdodson94: Other checking routines were `ELEMENTAL`, like `%isCircle`, so I was just sticking with the same convention.

src/Geom_Poly.f90
zdodson94: Good point, but I'm changing to use `MAXVAL` anyway.

zdodson94:
There are corner cases in which the sector or section might be deformed. This might show up down the line if the shape of the mesh is allowed to change. But in that case the future programmer could do something like check if the shape is "ideal" or "deformed". I added some commented code that would provide a start on this by checking if the two sides of the section or sector are the same length.

zdodson94:
Updating to leave out DBC because it can't be used in an `ELEMENTAL`, and thus `PURE`, routine.

zdodson94:
Is this ok to merge?

bkochuna:
I'm good with merging. @HendersonSC or @stimpsonsg any concerns with this?

stimpsonsg:
I'd prefer to wait till after the tag, though I acknowledge this should be pretty safe.

bkochuna:
Ok, lets wait for tag. That's kind of what my feeling was also.

djabaay:
Hey @stimpsonsg, since I think all necessary changes to Futility have been made and we can tag a previous commit if necessary, can we get this merged?  Zack needs this for getting other MR's merged into MPACT.

stimpsonsg:
Yeah, that's fine ;-)

djabaay:
Excellent, thanks! :)

PR #183, opened 2019-05-13
Adds isValidElemName to ElementsIsotopes

aarograh:
Also adds a new assignment interface for stringTypes to assign
a 1D array of length-1 characters to a single stringType to make
a few things simpler.

CASL Ticket # - 5907


aarograh:
This isn't urgent if we want to wait until after the tag.  It just supports some species transport work in MPACT.

PR #186, opened 2019-07-11
Add getZZZAAAI method to ElementsIsotopes

aarograh:


src/ElementsIsotopes.f90
```diff
@@ -105,6 +105,9 @@ MODULE ElementsIsotopes
       !> @copybrief ElementsIsotopes::getZAID_ElemIso
       !> @copydetails ElementsIsotopes::getZAID_ElemIso
       PROCEDURE,PASS :: getZAID => getZAID_ElemIso
```
stimpsonsg: is getZAID still used?

aarograh: For now, yes.  Most of them are actually calculating the `ZZZAAAI` from the `ZAID` so they could get replaced, but it's quite a few locations so I intend to leave it for now.  And having this function doesn't hurt anyways.

stimpsonsg: yeah

PR #189, opened 2019-09-03
Fix for very long warning messages

rks171:
Description:
Prior to this, if a message passed to raiseWarning exceeded 512 characters,
it would crash the code due to a IO error.  This fix will truncate the end
of the message in this case.

CASL Ticket # - 6064

### DIFF
--- a/src/ExceptionHandler.f90
+++ b/src/ExceptionHandler.f90
@@ -60,7 +60,7 @@
 !>   CALL e%setLogFileUnit(23)
 !>   OPEN(UNIT=e%getLogFileUnit(),FILE='Exception.log', &
 !>        ACCESS='SEQUENTIAL',FORM='FORMATTED')
-!>   CALL e%setLogFileActive(.TRUE.)
+!>   CALL e%setLogActive(.TRUE.)
 !>
 !>   !Suppress reporting of exceptions to standard error
 !>   CALL e%setQuietMode(.TRUE.)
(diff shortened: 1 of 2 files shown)


rks171:
I ran into this issue when pulling this into CTF.  Evidently I have some very long warning messages.  Ideally they should be shorter, but Futility was actually crashing on an IO error and not printing the message, which is something we definitely do not want for a warning message.

src/ExceptionHandler.f90
```diff
@@ -1035,7 +1035,8 @@ SUBROUTINE exceptionMessage(eCode,isQuiet,isLogActive,logUnit,mesg)
       ENDIF
 
       !Set the message to be included as one line back to exception object
-      WRITE(mesg,'(a)') TRIM(prefix)//' - '//TRIM(mesg)
+      prefix = TRIM(prefix)//' - '
+      WRITE(mesg,'(a)') TRIM(prefix)//TRIM(mesg(1:EXCEPTION_MAX_MESG_LENGTH-LEN(TRIM(prefix))))
```
aarograh: So it looks like this fixes the issue just fine, but I guess I wonder why we aren't just using StringType objects so that length isn't an issue at all?  Then we wouldn't need to worry about truncation of potentially important information either.

Dummy arguments for routines called from outside the module should be `CHARACTER(LEN=*)`, then converted to StringTypes if they need to be passed around a lot inside the module.

If that's going to be time consuming, we can defer for now.  But it would be good to at least take a quick look to see how time consuming that change would be.

rks171: I think that makes sense.  I was going for the least invasive fix here, but it might not take much more to fix it the right way.  BTW, do you know anything about the CI failure and failing to find HDF5?

aarograh: Yeah.  Changing just this routine and leaving everything else would be fine too.  My fear is that there are going to be a bunch of random unit tests in MPACT that will need to be updated.  In some places we check the previous error message (which I don't think we should, but that's a different debate).  If the Futility side seems really easy, I can spin up an MPACT build with this Futility branch and see if it's going to be a mess.

As for the CI failure, I don't know what's going on there.  It's only in the docker build and has been that way for awhile.  We really need to fix that but I don't think anyone has time to look into it.  The docker build's failure actually doesn't cause a failure in the CI either, which is scary.  You have 3 failing tests (testExceptionHandler, testParameterLists, and testPartitionGraph) in the other build that's causing the CI failure.

PR #194, opened 2019-09-30
Implement string library like functionality

HendersonSC:
Implements string library like functionality similar to other modern languages.


stimpsonsg:
Waiting for MPACT MR to by de-conflicted before merging.

PR #195, opened 2019-09-30
Improve vtk efficiency

aarograh:
Adds changes to improve the efficiency of generating large FSR VTK files.  Namely, the act of repeatedly adding VTK meshes together can be very slow.  This change allows client code to add an array of them together, which is much more efficient.

A couple other minor cleanup items are also added: improved error messages for files and a fix to a bad comment in Times.f90.

### DIFF
--- a/src/VTKFiles.f90
+++ b/src/VTKFiles.f90
@@ -89,6 +89,7 @@ MODULE VTKFiles
   PUBLIC :: VTKLegFileType
   PUBLIC :: eVTK
   PUBLIC :: OPERATOR(+)
+  PUBLIC :: sumMeshes_VTKMesh
 
   !>Interface for mesh addition
   INTERFACE OPERATOR(+)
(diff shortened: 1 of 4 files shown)


src/VTKFiles.f90
```diff
@@ -1421,4 +1422,58 @@ FUNCTION addMesh_VTKMesh(mesh1,mesh2) RESULT(newMesh)
                                                   mesh1conv%numPoints
   ENDFUNCTION addMesh_VTKMesh
 !
+!-------------------------------------------------------------------------------
+!> @brief Adds VTK mesh objects together so that a single mesh is formed
+!> @param meshes the array of meshes to add
+!> @returns mesh the resulting mesh object
+!>
+  FUNCTION sumMeshes_VTKMesh(meshes) RESULT(mesh)
```
stimpsonsg: any unit test?

aarograh: Added

unit_tests/testVTKFiles/testVTKFiles.f90
```diff
@@ -303,32 +351,112 @@ PROGRAM testVTKFiles
...
+    ASSERT_EQ(testVTKMesh%cellList(i),cellRef(i),'%cellList')
+  ENDDO !i
+
   DEALLOCATE(xref)
   DEALLOCATE(yref)
   DEALLOCATE(zref)
+  DEALLOCATE(cellRef)
+  DEALLOCATE(nodeRef)
+
+  COMPONENT_TEST('Array')
```
aarograh: This COMPONENT_TEST is for the new routine I added.  Everything else was just rearranging to use the test harness better and update to follow the new code standard.

PR #196, opened 2019-10-01
Partition graph metrics update

djabaay:
Updating the partition graph type to have a weighting factor and the %metrics routine to return the max fraction of processor dependent source regions divided by the total number.

### DIFF
--- a/src/PartitionGraph.f90
+++ b/src/PartitionGraph.f90
@@ -36,6 +36,8 @@
 !>
 !++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
 MODULE PartitionGraph
+#include "Futility_DBC.h"
+  USE Futility_DBC
   USE IntrType
   USE Allocs
   USE Strings
@@ -1822,7 +1824,7 @@ MODULE PartitionGraph
       ENDIF
 
       ENSURE((maxnsr > 0.0_SRK) .AND. (maxnsr <= 1.0_SRK))
-      ENSURE((mmr > 0.0_SRK)
+      ENSURE(mmr > 0.0_SRK)
       ENSURE((srms > 0.0_SRK) .AND. (srms <= 100.0_SRK))
       ENSURE(ecut >= 0.0_SRK)
       ENSURE(comm >= 0.0_SRK)
--- a/src/PartitionGraph.f90
+++ b/src/PartitionGraph.f90
@@ -1720,7 +1720,7 @@ MODULE PartitionGraph
 !-------------------------------------------------------------------------------
 !> @brief Calculate metrics relevant to the quality of partition
 !> @param thisGraph the partitioned graph to check metrics of
-!> @param maxnsr the maximum number of source regions
+!> @param maxnsr the maximum ratio of the number of source regions
 !> @param mmr the ratio of maximum sized(weighted) group to minimum
 !> @param srms the rms of the group size(weighted) difference from optimal
 !> @param ecut the total weight of edges cut
@@ -1820,6 +1820,12 @@ MODULE PartitionGraph
         CALL ePartitionGraph%raiseError(modName//'::'//myName// &
           ' - graph is not partitioned!')
       ENDIF
+
+      ENSURE((maxnsr > 0.0_SRK) .AND. (maxnsr <= 1.0_SRK))
+      ENSURE((mmr > 0.0_SRK)
+      ENSURE((srms > 0.0_SRK) .AND. (srms <= 100.0_SRK))
+      ENSURE(ecut >= 0.0_SRK)
+      ENSURE(comm >= 0.0_SRK)
     ENDSUBROUTINE calcDecompMetrics_PartitionGraph
 !
 !-------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/PartitionGraph.f90
```diff
@@ -111,6 +111,8 @@ MODULE PartitionGraph
     INTEGER(SIK) :: nGroups=0
     !> Number of partitioning methods
     INTEGER(SIK) :: nPart=0
+    !> Weighting factor used when normalizing computational weights
+    REAL(SRK) :: wtfactor=0.0_SRK
```
apfitzge: Could this be initialized to 1.0, since that should be the default behavior?

src/PartitionGraph.f90
```diff
@@ -1703,28 +1720,42 @@ SUBROUTINE SpatialKernighanLin_PartitionGraph(thisGraph,L1,L2)
 !-------------------------------------------------------------------------------
 !> @brief Calculate metrics relevant to the quality of partition
 !> @param thisGraph the partitioned graph to check metrics of
+!> @param maxnsr the maximum number of source regions
```
apfitzge: Possibly reading the code incorrectly, but this returns the max **fraction** of source regions, right? If so, we should let this comment reflect that.

unit_tests/testPartitionGraph/testPartitionGraph.f90
```diff
@@ -397,10 +412,11 @@ SUBROUTINE testClear()
       CALL testPG%clear()
       CALL params%clear()
       ASSERT(testPG%nvert == 0,'%clear(...)%nvert')
-      ASSERT(testPG%dim == 0,'%clear(...)%nvert')
-      ASSERT(testPG%maxneigh == 0,'%clear(...)%nvert')
-      ASSERT(testPG%nGroups == 0,'%clear(...)%nvert')
-      ASSERT(testPG%nPart == 0,'%clear(...)%nvert')
```
apfitzge: Good fix. Copypasta from my original implementation I guess!

src/PartitionGraph.f90
```diff
@@ -1703,28 +1720,42 @@ SUBROUTINE SpatialKernighanLin_PartitionGraph(thisGraph,L1,L2)
...
+          igstt=thisGraph%groupIdx(ig)
+          igstp=thisGraph%groupIdx(ig+1)-1
+          gwt=0.0_SRK
+          DO iv=igstt,igstp
+            gwt=gwt+thisGraph%wts(thisGraph%groupList(iv))
+          ENDDO
+          maxnsr=MAX(maxnsr,REAL(gwt,SRK))
+        ENDDO
+        maxnsr=maxnsr/SUM(thisGraph%wts)
+
```
apfitzge: Might be reasonable to put a DBC statement at the end to verify that 
`(maxnsr > 0_SRK) .AND. (maxnsr < 1.0_SRK)`

src/PartitionGraph.f90
djabaay: I thought about it, but changed it to `0.0` to reflect the uninitialized state.  When the graphtype is called, if there is no parameter present, it will get set to `1.0` by default.  That make sense and sound ok?

src/PartitionGraph.f90
djabaay: Good call.

djabaay: While I'm doing the one, are these reasonable for the others?
```
      ENSURE((maxnsr > 0.0_SRK) .AND. (maxnsr <= 1.0_SRK))                                  
      ENSURE((mmr > 0.0_SRK)                                                                                                                                                             
      ENSURE((srms > 0.0_SRK) .AND. (srms <= 100.0_SRK))                                    
      ENSURE(ecut >= 0.0_SRK)                                                               
      ENSURE(comm >= 0.0_SRK)            
```

src/PartitionGraph.f90
apfitzge: Yeah, it makes sense. This sort of thing is done a lot in Futility & MPACT. I think my mind always just defaults to the c++ behavior where you construct objects, and these initial values are respected i.e shouldn't have uninitialized states.

djabaay: Gotcha.

djabaay:
All items have been updated.  Awaiting your reply @apfitzge.

djabaay:
Currently just running MPACT master with the branch checked out to verify the OPTIONAL argument is compatible.  When the tests finish, I'll post results.

djabaay:
```
100% tests passed, 0 tests failed out of 472

Subproject Time Summary:
CTeuchos         =   1.39 sec*proc (4 tests)
ForTeuchos       =   2.46 sec*proc (4 tests)
Futility         =  48.15 sec*proc (60 tests)
MPACT_API        =  44.02 sec*proc (3 tests)
MPACT_Drivers    =   2.33 sec*proc (4 tests)
MPACT_exe        = 16000.34 sec*proc (218 tests)
MPACT_libs       = 1102.36 sec*proc (179 tests)

Label Time Summary:
BASIC         =  48.15 sec*proc (60 tests)
CONTINUOUS    =  48.15 sec*proc (60 tests)
HEAVY         =  48.15 sec*proc (60 tests)
NIGHTLY       =  48.15 sec*proc (60 tests)

Total Test time (real) = 1265.74 sec

djabaay@nightfort:~/MPACT/Futility$ git status
# On branch PartitionGraphMetricsUpdate
```

No failures, safe to merge in with the optional argument.

djabaay:
ping @bkochuna

PR #197, opened 2019-10-01
Pl get string update

djabaay:
Change the %getString interface to allow the writing of multidimensional parameter to a scalar string.


src/ParameterLists.f90
```diff
@@ -2031,20 +2039,168 @@ SUBROUTINE getString_ParamType_scalar(thisParam,name,string,sskfmt,sdkfmt)
...
+              WRITE(tmpchar,TRIM(sdkfmtDef)) param%val(i)
+              string=string//delimDef//TRIM(ADJUSTL(tmpchar))
+            ENDDO
+          TYPE IS(ParamType_SNK_a1)
+            WRITE(tmpchar,'(i0)') param%val(1)
+            string=TRIM(ADJUSTL(tmpchar))
+            DO i=2,SIZE(param%val)
+              WRITE(tmpchar,'(i0)') param%val(i)
+              string=string//delimDef//TRIM(ADJUSTL(tmpchar))
+            ENDDO
```
aarograh: It's worth noting that `IOUtils` has a `str(...)` function for integers and reals that eliminates the need for a temporary character string and, in many cases, formatting statements.  Might be good to start using that more for the SSK/SDK/SNK/SLK branches here.

djabaay: Ah, good to know Aaron.  I can go through an update those.  Do the SSK/SDK routines have an option for a format string?

aarograh: Yeah, there's a version that just returns all the digits, and a version that returns a specified number of decimal places.  I didn't look closely at how you're formatting the reals, so the `str` function may or may not work there.  But the integer version of `str` uses the `(i0)` formatting by default, so that one would definitely work.

src/ParameterLists.f90
```diff
@@ -2031,20 +2039,158 @@ SUBROUTINE getString_ParamType_scalar(thisParam,name,string,sskfmt,sdkfmt)
...
           TYPE IS(ParamType_SLK)
-            WRITE(tmpchar,'(i0)') param%val
+            string=str(param%val)
           TYPE IS(ParamType_SBK)
             WRITE(tmpchar,'(L1)') param%val
+            string=TRIM(ADJUSTL(tmpchar))
           TYPE IS(ParamType_STR)
-            tmpchar=param%val
+            string=param%val
+          TYPE IS(ParamType_SSK_a1)
```
lamilamilami: No indentation for the TYPE IS ... according to the style guide?

src/ParameterLists.f90
```diff
@@ -2031,20 +2039,158 @@ SUBROUTINE getString_ParamType_scalar(thisParam,name,string,sskfmt,sdkfmt)
...
+                ENDDO
+              ENDDO
+            ENDDO
+!          TYPE IS(ParamType_SBK_a3)
+!            ALLOCATE(string(SIZE(param%val))
+!            DO i=1,SIZE(param%val)
+!              WRITE(tmpchar,'(L1)') param%val(i)
+!              string(i)=TRIM(ADJUSTL(tmpchar))
+!            ENDDO
+!          TYPE IS(ParamType_STR_a3)
```
lamilamilami: Why is this commented out?

src/ParameterLists.f90
```diff
@@ -2054,13 +2200,14 @@ SUBROUTINE getString_ParamType_scalar(thisParam,name,string,sskfmt,sdkfmt)
 !> @param name
 !> @param hasname
 !>
-    SUBROUTINE getString_ParamType_a1(thisParam,name,string,sskfmt,sdkfmt)
+    SUBROUTINE getString_ParamType_a1(thisParam,name,string,sskfmt,sdkfmt,delim)
```
lamilamilami: Why were the converters of arrays implemented in the routine for scalar? Are these supposed to go to the following routines?

djabaay: These routines are the original set of procedures for the generic `%getString` interface.  Originally, I envisioned doing a consistent dimensional conversion from an arbitrary intrinsic type to the corresponding string type (e.g. scalar int to scalar string, 1-D float array to 1-D string array, and so on...)  For the conversion of parameter list to output table, I wanted to be able to put a 1-D array into a single table entry, so I needed to modify `%getString_ParamType` to handle the conversion of multidimensional types to a single scalar string (e.g. scalar int to scalar string, 1-D float array to scalar string, and so on...).

The only change here is adding an optional delimiter, just so the procedures for the interfaces were consistent and simpler to understand.  The delimiter isn't used in this or other higher dimensional routines.  So, keep for consistency, or remove because not used?

src/ParameterLists.f90
```diff
@@ -2031,20 +2039,158 @@ SUBROUTINE getString_ParamType_scalar(thisParam,name,string,sskfmt,sdkfmt)
...
+              ENDDO
+            ENDDO
+          TYPE IS(ParamType_SLK_a2)
+            string=''
+            DO j=1,SIZE(param%val,DIM=2)
+              DO i=1,SIZE(param%val,DIM=1)
+                string=string//str(param%val(i,j))//delimDef
+              ENDDO
+            ENDDO
+!          TYPE IS(ParamType_SBK_a2)
```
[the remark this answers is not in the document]
djabaay: The logical arrays are commented out because there is no parameter list type for a 2-D array of booleans.  In the event that parameter type gets added, this code can be uncommented.  Still want to keep it?

src/ParameterLists.f90
djabaay: The 3-D string arrays are commented out because there is no parameter list type for that dimensionality. In the event that parameter type gets added, this code can be uncommented. Still want to keep it?

src/ParameterLists.f90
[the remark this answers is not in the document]
lamilamilami: Probably take them out for now since they are not valid pList types yet. Looks like pretty easy to add them back if there's a need to add these types.

src/ParameterLists.f90
lamilamilami: same here.

src/ParameterLists.f90
lamilamilami: Oh..that makes sense. Maybe remove 'delim' for non-scalar routines? That way compiler will complain if you call it with 'delim' but accidentally pass a multi-dimentional string, instead of silently converting it using the non-scalar routines.

djabaay: Sounds good to me.  I'll make the change.

src/ParameterLists.f90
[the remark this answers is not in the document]
djabaay: Roger that.

djabaay:
rebased with master and addressed comments

djabaay:
@aarograh and @lamilamilami, can both of you guys review the changes that got pushed and if they are satisfactory, close your conversations?

src/ParameterLists.f90
```diff
@@ -2011,40 +2018,164 @@ SUBROUTINE getString_ParamType_scalar(thisParam,name,string,sskfmt,sdkfmt)
...
+          string=''
+          DO k=1,SIZE(param%val,DIM=3)
+            DO j=1,SIZE(param%val,DIM=2)
+              DO i=1,SIZE(param%val,DIM=1)
+                string=string//str(param%val(i,j,k))//delimDef
+              ENDDO
+            ENDDO
+          ENDDO
+        CLASS DEFAULT
+          !Error, not a supported parameter type...
```
aarograh: Actually throw an error

src/ParameterLists.f90
```diff
@@ -2072,54 +2203,54 @@ SUBROUTINE getString_ParamType_a1(thisParam,name,string,sskfmt,sdkfmt)
...
+            WRITE(tmpchar,'(L1)') param%val(i)
+            string(i)=TRIM(ADJUSTL(tmpchar))
+          ENDDO
+        TYPE IS(ParamType_STR_a1)
+          ALLOCATE(string(SIZE(param%val,DIM=1)))
+          DO i=1,SIZE(param%val)
+            string(i)=param%val(i)
+          ENDDO
+        CLASS DEFAULT
+          !Error, not a 1-D array...
```
aarograh: Actually throw an error

src/ParameterLists.f90
```diff
@@ -2147,64 +2278,58 @@ SUBROUTINE getString_ParamType_a2(thisParam,name,string,sskfmt,sdkfmt)
...
+          ALLOCATE(string(SIZE(param%val,DIM=1),SIZE(param%val,DIM=2)))
+          DO j=1,SIZE(param%val,DIM=2)
+            DO i=1,SIZE(param%val,DIM=1)
+              string(i,j)=param%val(i,j)
             ENDDO
-          CLASS DEFAULT
-            !Error, not a scalar...
+          ENDDO
+        CLASS DEFAULT
+          !Error, not a 2-D array...
```
aarograh: Actually throw an error

src/ParameterLists.f90
```diff
@@ -2011,40 +2018,164 @@ SUBROUTINE getString_ParamType_scalar(thisParam,name,string,sskfmt,sdkfmt)
...
-!> @brief
-!> @param thisParam
-!> @param name
-!> @param hasname
+!> @brief This subroutine takes a parameter type and a path, and converts
+!>        the 1-D intrinsic parameter type it finds into a 1-D array of strings.
+!>        This will not work if the parameter type is a parameter list.
+!> @param thisParam The parameter type to be searched
+!> @param name The path name to the parameter to be converted to a string
+!> @param string The output scalar string type
```
aarograh: fix the comment to say 1-D array string type

src/ParameterLists.f90
```diff
@@ -2072,54 +2203,54 @@ SUBROUTINE getString_ParamType_a1(thisParam,name,string,sskfmt,sdkfmt)
...
-!> @brief
-!> @param thisParam
-!> @param name
-!> @param hasname
+!> @brief This subroutine takes a parameter type and a path, and converts
+!>        the 2-D intrinsic parameter type it finds into a 2-D array of strings.
+!>        This will not work if the parameter type is a parameter list.
+!> @param thisParam The parameter type to be searched
+!> @param name The path name to the parameter to be converted to a string
+!> @param string The output scalar string type
```
aarograh: fix the comment to say 2-D array string type

src/ParameterLists.f90
```diff
@@ -2147,64 +2278,58 @@ SUBROUTINE getString_ParamType_a2(thisParam,name,string,sskfmt,sdkfmt)
...
-!> @brief
-!> @param thisParam
-!> @param name
-!> @param hasname
+!> @brief This subroutine takes a parameter type and a path, and converts
+!>        the 3-D intrinsic parameter type it finds into a 3-D array of strings.
+!>        This will not work if the parameter type is a parameter list.
+!> @param thisParam The parameter type to be searched
+!> @param name The path name to the parameter to be converted to a string
+!> @param string The output scalar string type
```
aarograh: fix the comment to say 3-D array string type

src/ParameterLists.f90
```diff
@@ -2232,63 +2357,46 @@ SUBROUTINE getString_ParamType_a3(thisParam,name,string,sskfmt,sdkfmt)
...
-!            DO j=1,SIZE(param%val,DIM=2))
-!              DO i=1,SIZE(param%val,DIM=1))
-!                string(i,j)=param%val(i,j)
-!              ENDDO
-!            ENDDO
-          CLASS DEFAULT
-            !Error, not a scalar...
+          ENDDO
+        CLASS DEFAULT
+          !Error, not a 3-D array...
```
aarograh: Actually throw an error

unit_tests/testParameterLists/testParameterLists.f90
```diff
@@ -4634,6 +4635,113 @@ SUBROUTINE testGetString()
...
+    ASSERT(tmpbool,'testSNKa3 string')
+    CALL testParam%getString('testPL->testSNKa3_2',tmpstr)
+    tmpbool=(tmpstr == '311; 411; 321; 421; 312; 412; 322; 422')
+    ASSERT(tmpbool,'testSNKa3_2 string')
+    CALL testParam%getString('testPL->testSLKa3',tmpstr)
+    tmpbool=(tmpstr == '111; 211; 121; 221; 112; 212; 122; 222')
+    ASSERT(tmpbool,'testSLKa3 string')
+    CALL testParam%getString('testPL->testSLKa3_2',tmpstr)
+    tmpbool=(tmpstr == '311; 411; 321; 421; 312; 412; 322; 422')
+    ASSERT(tmpbool,'testSLKa3_2 string')
```
aarograh: Should really be using the new unit test macros... so the above 2 lines become `ASSERT_EQ(tmpstr, '311; 411; 321; 421; 312; 412; 322; 422','testSLKa3_2 string')`.  Switch to those throughout the test instead of using `tmpbool`.  The main times it's ok not to use the new macros is when an `ANY(...)` or `ALL(...)` statement is required for testing.

djabaay: Or when there is a comparison that extends beyond the 80 char limit I assume as well, right?

src/ParameterLists.f90
aarograh: I meant the `!> @param string The output scalar string type` line

src/ParameterLists.f90
aarograh: I meant the `!> @param string The output scalar string type` line

src/ParameterLists.f90
aarograh: I meant the `!> @param string The output scalar string type` line

unit_tests/testParameterLists/testParameterLists.f90
```diff
@@ -4636,41 +4636,31 @@ SUBROUTINE testGetString()
...
-    tmpbool=(tmpstr == '1.500000E+00; 1.600000E+00')
-    ASSERT(tmpbool,'testSSKa1 string')
+    ASSERT_EQ(CHAR(tmpstr),'1.500000E+00; 1.600000E+00','testSSKa1 string')
     CALL testParam%getString('testPL->testSSKa1_2',tmpstr)
-    tmpbool=(tmpstr == '1.700000E+00; 1.800000E+00')
-    ASSERT(tmpbool,'testSSKa1_2 string')
+    ASSERT_EQ(CHAR(tmpstr),'1.700000E+00; 1.800000E+00','testSSKa1_2 string')
     CALL testParam%getString('testPL->testSDKa1',tmpstr)
     tmpbool=(tmpstr == '2.500000000000000E+00; 2.600000000000000E+00')
     ASSERT(tmpbool,'testSDKa1 string')
```
aarograh: Change this one too

unit_tests/testParameterLists/testParameterLists.f90
```diff
@@ -4636,41 +4636,31 @@ SUBROUTINE testGetString()
...
     CALL testParam%getString('testPL->testSDKa1_2',tmpstr)
-    tmpbool=(tmpstr == '2.700000000000000E+00; 2.800000000000000E+00')
+    tmpbool=tmpstr == '2.700000000000000E+00; 2.800000000000000E+00'
     ASSERT(tmpbool,'testSDKa1_2 string')
```
aarograh: Change this one too?

unit_tests/testParameterLists/testParameterLists.f90
aarograh: The 80 character limit is defunct now.  We typically cap at 100 now (and allow beyond 100 if there's a good reason for it, though that's not too common).  But yeah, long lines with the macros are also exceptions.

unit_tests/testParameterLists/testParameterLists.f90
aarograh: Change this one too?

djabaay:
Ran Continuous tests with MPACT + CTF:
```
100% tests passed, 0 tests failed out of 472

Subproject Time Summary:
CTeuchos         =   1.21 sec*proc (4 tests)
ForTeuchos       =   2.16 sec*proc (4 tests)
Futility         =  42.26 sec*proc (60 tests)
MPACT_API        =  42.83 sec*proc (3 tests)
MPACT_Drivers    =   2.42 sec*proc (4 tests)
MPACT_exe        = 15745.81 sec*proc (218 tests)
MPACT_libs       = 1095.68 sec*proc (179 tests)

Label Time Summary:
BASIC         =  42.26 sec*proc (60 tests)
CONTINUOUS    =  42.26 sec*proc (60 tests)
HEAVY         =  42.26 sec*proc (60 tests)
NIGHTLY       =  42.26 sec*proc (60 tests)

Total Test time (real) = 1134.28 sec
```

djabaay:
A thought @aarograh, I have the delimiter set to be defaulted to the semi-colon.  I then use the semi-colon later to use with the `getField()` subroutine.  Wouldn't it be better if the delimiter for non-scalar arrays was to just surround the entry in quotes and separate them with a single space?  That way, the `getField()` subroutine will have direct correspondence and interoperability.  Does what I'm saying make sense?

To say it another way, I'm using the semi-colon delimiter as a proxy for the field separator, which is white space.  Wouldn't it be better to make them consistent?

aarograh:
@djabaay So are you saying that instead of the output string for an array (taken from one of our discussions above) being `'311; 411; 321; 421; 312; 412; 322; 422'` that it would instead be `"311" "411" "321" "421" " 312" "412" "322" "422"`?

djabaay:
Correct.  Then I can call `getField()` and just input the field number and get the entry I need.  Currently, there is an underlying assumption that the semi-colon needs a space after it, since that is what `getField()` parses.  I think that just enforcing the syntax `getField()` expects directly makes things less fragile and murky.

aarograh:
Dan and I talked about the delimeter change on slack and agreed that was fine.  I'm good with changes to go from semicolon to quotes for the delimiters.  This should be all set to merge if @bkochuna or @stimpsonsg would like to push the buttom.

djabaay:
ping @bkochuna or @stimpsonsg

bkochuna:
I'm good with merging. I assume Aaron resolved all discussions?

djabaay:
Changes have all been approved.

PR #198, opened 2019-10-04
More info when receiving unexpected variable type in ParameterList

mattkab2:
This patch raises a fatal error when receiving a bad character string to a single float, double, or integer value in the parameter list. This error contains the name of the property that is being read, to help a user debug the cause.

The error is fatal because if the program is not terminated it will exit on a Fortran runtime error when executing PL%add().


aarograh:
This request has been open for a few months now... what's the status?  Is it something we want?

aarograh:
This is stale and conflicted.   Going to close it.  If we decide this is useful someone can grab the branch and update it.

PR #201, opened 2019-11-01
Moving writeTable routines from Edits to IO_Strings.

djabaay:
Description:
Adding the routine to take a 2-D string array and write a preformatted
table to an output file.


djabaay:
We should always be writing to a file, so this belongs on FileType_Fortran, not IO_Strings.

PR #202, opened 2019-11-04
Write table on fortran file type.

djabaay:
Adding functionality to write a 2-D string array as a table on the fortran file type.

### DIFF
--- a/src/FileType_Fortran.f90
+++ b/src/FileType_Fortran.f90
@@ -985,6 +985,25 @@ MODULE FileType_Fortran
       ENDDO
     ENDSUBROUTINE getTableBounds
 !
+!------------------------------------------------------------------------------
+!> @brief This subroutine returns the formatted row string to separate rows and
+!>        to bound the top and bottom of the table
+!> @param maxcolsize The integer array of the width of each column, from
+!>        getTableBounds.
+!> @param rowstr The formatted row string
+!>
+    FUNCTION getRowStr(maxcolsize) RESULT(rowstr)
+      INTEGER(SIK),INTENT(IN) :: maxcolsize(:)
+      TYPE(StringType) :: rowstr
+      INTEGER(SIK) :: i,dashlen
+
+      rowstr='+'
+      DO i=1,SIZE(maxcolsize)
+        dashlen=maxcolsize(i)+2
+        rowstr=rowstr//REPEAT('-',dashlen)//'+'
+      ENDDO
+    ENDFUNCTION getRowStr
+!
 !-------------------------------------------------------------------------------
 !> @brief Returns a unit number that is presently not in use.
 !> @param istt optional input for where to start searching for an unused unit
--- a/src/FileType_Fortran.f90
+++ b/src/FileType_Fortran.f90
@@ -170,6 +170,9 @@ MODULE FileType_Fortran
       !> @copybrief FileType_Fortran::setStatus_fortran_file
       !> @copydetails FileType_Fortran::setStatus_fortran_file
       PROCEDURE,PASS :: setStatus => setStatus_fortran_file
+      !> @copybrief FileType_Fortran::writeTable_fortran_file
+      !> @copydetails FileType_Fortran::writeTable_fortran_file
+      PROCEDURE,PASS :: writeTable => writeTable_fortran_file
   ENDTYPE FortranFileType
 !
 !===============================================================================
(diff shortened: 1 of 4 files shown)


djabaay:
@sooyoung0321 When you've got some free time, please have a look at the code for a review.

src/FileType_Fortran.f90
```diff
@@ -857,6 +860,151 @@ SUBROUTINE setStatus_fortran_file(file,status)
       ENDIF
     ENDSUBROUTINE setStatus_fortran_file
 !
+!------------------------------------------------------------------------------
+!> @brief This subroutine writes a 2-D array of strings to an output file as a
+!>        formatted table.
+!> @param funit The output unit of the file where the table is written.
```
sooyoung0321: funit is not used in this subroutine.

stimpsonsg:
What ticket is this for?

djabaay:
https://vminfo.casl.gov/trac/casl_phi_kanban/ticket/6089

aarograh:
Just out of curiosity, why is this a type-bound procedure on the file type?  None of the formatting work requires the file object at all.  Seems like the formatting of the 2D strings into a table should be a separate function from the write function.

stimpsonsg:
ok, I'm seeing now mention of the ticket number in any of the commits or anything...it needs to be there

djabaay:
@stimpsonsg Agreed, missed that.  I'll go back and add.
@aarograh Is it worth it to convert the 2-D string array and process it into a 1-D string array that can just be looped over and written to a file?  Is that essentially what you're asking?

aarograh:
> @aarograh Is it worth it to convert the 2-D string array and process it into a 1-D string array that can just be looped over and written to a file? Is that essentially what you're asking?

Yeah, that's essentially what I had in mind.  The formatting of the 2D array of strings seems to me to be a separate action from writing it to a file.  And it's conceivable that one would want to use that capability to write to stdout or using `WRITE(unit_number,*)` or something, even if just for debugging purposes.

If there's a compelling reason to leave it this way, that's totally fine.  But it seems like much of the `writeTable_fortran_file` method could just be cut out into a separate function that's called from `writeTable_fortran_file`.

djabaay:
> > @aarograh Is it worth it to convert the 2-D string array and process it into a 1-D string array that can just be looped over and written to a file? Is that essentially what you're asking?
> 
> Yeah, that's essentially what I had in mind. The formatting of the 2D array of strings seems to me to be a separate action from writing it to a file. And it's conceivable that one would want to use that capability to write to stdout or using `WRITE(unit_number,*)` or something, even if just for debugging purposes.
> 
> If there's a compelling reason to leave it this way, that's totally fine. But it seems like much of the `writeTable_fortran_file` method could just be cut out into a separate function that's called from `writeTable_fortran_file`.

So, how about something like stringTableToLines() that lives in `IO_Strings.f90` taking a 2-D formatted string array that changes it into a 1-D array of strings that can be looped over and written to a file.  The question is, where should the write linesToFile() or whatever live?  Still here?  Should that also be in `IO_Strings.f90`?

aarograh:
I definitely agree with the `stringTableToLines()` in `IO_Strings.f90`.  I think having a method on the file type still makes sense to actually write the 1D array of strings to a file.  I'm kind of thinking how the HDF5 file works with a `GENERIC :: write`, but only implementing the 1D StringType version right now for the fortran file type.

djabaay:
Roger that, I can break stuff up and move it around.

djabaay:
Rebased master onto the branch, but forgot to update the commit messages.  The writeTable routine is now split up as @aarograh requested, and the commit messages have PHI ticket numbers as @stimpsonsg requested.

djabaay:
Ran the Nightlies with MPACT:
```
99% tests passed, 6 tests failed out of 750

Subproject Time Summary:
CTeuchos         =   1.32 sec*proc (4 tests)
ForTeuchos       =   2.86 sec*proc (4 tests)
Futility         =  44.19 sec*proc (60 tests)
MPACT_API        = 240.85 sec*proc (4 tests)
MPACT_Drivers    =   2.64 sec*proc (4 tests)
MPACT_exe        = 126447.21 sec*proc (498 tests)
MPACT_libs       = 1088.74 sec*proc (176 tests)

Label Time Summary:
BASIC         =  44.19 sec*proc (60 tests)
CONTINUOUS    =  44.19 sec*proc (60 tests)
HEAVY         =  44.19 sec*proc (60 tests)
NIGHTLY       =  44.19 sec*proc (60 tests)

Total Test time (real) = 8510.05 sec

The following tests FAILED:
	321 - MPACT_exe_testValid_verify_subplane_rod_coupled (Failed)
	391 - MPACT_exe_testValid_transient_4-mini_3D (Failed)
	392 - MPACT_exe_testValid_transient_4-mini_1GAccel (Failed)
	393 - MPACT_exe_testValid_transient_4-mini_3D_SCRAM (Failed)
	396 - MPACT_exe_testValid_steadystate_4mini_crw (Failed)
	397 - MPACT_exe_testValid_transient_4mini_crw (Failed)
Errors while running 
```
subplane_rod_coupled fails because I didn't compile with CTF (I have a commit to IFDEF this in CMake with ENABLED_CTF).  The others fail because of the negative fluxes with DBC.
```
DBC Failure: .NOT. ANY(tmpTS%phis(:,ig) < 0.0_SRK) in /home/djabaay/MPACT/MPACT_libs/CoreSolvers/src/FixedSrcSolver.f90 on line  953:  process   14 of   16
```

stimpsonsg:
I think we need to reconsider this DBC check.  The scalar flux can go negative without being completely catastrophic.

djabaay:
That's fine, I can make a MR on our end and remove them if that is the consensus.

bscollin:
I do think we need a check on the final solution to ensure its positive and throw a warning, but agree during the iteration, things going negative isn't the end of the world.  It might run into issues if negative fluxes get passed to ORIGEN (but that would need to be a negative total flux... group integrated).

djabaay:
That makes sense.  Would ORIGEN need to ensure that all source region fluxes for all groups on a single cross section region sum to a positive value at the end of a solve?

src/FileType_Fortran.f90
```diff
@@ -857,6 +861,28 @@ SUBROUTINE setStatus_fortran_file(file,status)
...
+!> @param lines The 1-D array of strings to write.
+!>
+    SUBROUTINE write_str_1a_fortran_file(file,lines)
+      CHARACTER(LEN=*),PARAMETER :: myName='writeTable_fortran_file'
+      CLASS(FortranFileType),INTENT(INOUT) :: file
+      TYPE(StringType),INTENT(IN) :: lines(:)
+      INTEGER(SIK) :: i
+
+      IF(file%initstat) THEN
+        IF(file%isOpen()) THEN
```
aarograh: Should probably make both of these IF checks DBC?  We shouldn't write to an unopened or nonexistent file

src/IO_Strings.f90
```diff
@@ -1629,4 +1630,150 @@ FUNCTION str_SDK_nDecimal(r,nDecimal) RESULT(string)
...
+      !Three for the top, first row, and last row separator strings.
+      ALLOCATE(lines(SUM(maxrowsize)+3))
+      rowstr=getRowStr(maxcolsize)
+      nrow=1
+      lines(nrow)=rowstr
+      rightpad=0
+
+      !
+      !Loop over the rows
+      DO j=1,SIZE(tablevals,DIM=2)
```
aarograh: So the indexing is (column,row)?  Seems backwards...

src/IO_Strings.f90
```diff
@@ -1629,4 +1630,150 @@ FUNCTION str_SDK_nDecimal(r,nDecimal) RESULT(string)
...
+              !If this table value is an array, get the kth value
+              !Remember that getField uses the space as a delimiter.
+              !We're using the comma+space so we have interoperability.
+              IF(INDEX(tablevals(i,j),';') > 0) THEN
+                CALL getField(k,tablevals(i,j),field)
+              !If it's not an array
+              ELSE
+                field=tablevals(i,j)
+              ENDIF
+              !hasString=hasString .AND. field /= '-'
```
aarograh: delete this line

unit_tests/testFileType_Fortran/testFileType_Fortran.f90
```diff
@@ -134,6 +138,75 @@ SUBROUTINE testFortranFileType()
...
+        POSITION='REWIND')
+      CALL testFile%fopen()
+      ALLOCATE(lines(5))
+      lines(1)='+-------+-----+'
+      lines(2)='| Test1 | 1.0 |'
+      lines(3)='+-------+-----+'
+      lines(4)='| Test2 | 10  |'
+      lines(5)='+-------+-----+'
+      CALL testFile%fwrite(lines)
+      !DEALLOCATE(lines)
```
aarograh: delete this line

unit_tests/testFileType_Fortran/testFileType_Fortran.f90
```diff
@@ -134,6 +138,75 @@ SUBROUTINE testFortranFileType()
...
+      lines(2)='| "Some Quotes"    | semi-             | more            |'
+      lines(3)='|                  | colon             | than            |'
+      lines(4)='|                  |                   | one semi colon  |'
+      lines(5)='+------------------+-------------------+-----------------+'
+      lines(6)='| True             | loooooooooooooong | -               |'
+      lines(7)='| Mostly False-ish | 1.000E+23         | 100,000,000,000 |'
+      lines(8)='| or True          |                   |                 |'
+      lines(9)='+------------------+-------------------+-----------------+'
+      CALL testFile%fwrite(lines)
+      !DEALLOCATE(lines)
```
aarograh: delete this line

unit_tests/testFileType_Fortran/testFileType_Fortran.f90
```diff
@@ -134,6 +138,75 @@ SUBROUTINE testFortranFileType()
...
+      ASSERT_EQ(TRIM(buffer),'   '//CHAR(lines(6)),"lines line 11")
+      READ(UNIT=testfile%getUnitNo(),FMT='(a)',ADVANCE='NO',IOSTAT=ioerr) buffer
+      ASSERT_EQ(TRIM(buffer),'   '//CHAR(lines(7)),"lines line 12")
+      READ(UNIT=testfile%getUnitNo(),FMT='(a)',ADVANCE='NO',IOSTAT=ioerr) buffer
+      ASSERT_EQ(TRIM(buffer),'   '//CHAR(lines(8)),"lines line 13")
+      READ(UNIT=testfile%getUnitNo(),FMT='(a)',ADVANCE='NO',IOSTAT=ioerr) buffer
+      ASSERT_EQ(TRIM(buffer),'   '//CHAR(lines(9)),"lines line 14")
+      DEALLOCATE(lines)
+      CALL testFile%fclose()
+      !CALL testFile%fdelete()
```
aarograh: delete this line

src/FileType_Fortran.f90
djabaay: Makes sense.  Do we want it to call `%fopen` if it isn't open?  How do we do it in the HDF5 file?

djabaay: Looks like we throw an error if the file isn't opened before the call.  I can make both DBC checks.

src/IO_Strings.f90
djabaay: I was thinking x position and y position in the table.

src/FileType_Fortran.f90
aarograh: I think the HDF5 requires you to open it before calling read/write, but I see no harm in opening the first here if it's not yet.  Just make sure to close it after writing if it was closed upon entering the routine so that the open/closed status of the file is unchanged upon return.

I would kind of like to treat the hdf5 file the same way and delete a LOT of open/close statements through MPACT, but that's a separate discussion

src/IO_Strings.f90
aarograh: Ok.  I guess that's fine, just opposite of how I was thinking of it.  It looks like the memory is still accessed in optimal order, so it's ok as is.

I do think it would be good to note the indexing in the header comments for this routine though: "The first index of tablevals is the x position (from left to right) in the table and the second index is the y position (from top to bottom) in the table".  This should help alleviate any confusion by other users.

djabaay: Agreed, that's a good note to have.

djabaay:
@aarograh Pushed updates based on your feedback.  I added the ticket number to the last few commits, which should show up in the merge commit when squashed, right?

djabaay:
@aarograh, I pushed changes, when you can, let me know if those comments are resolved.

aarograh:
Looks ok to me.  @stimpsonsg or @bkochuna can merge whenever

bkochuna:
Merging since @aarograh has given the greenlight.

PR #208, opened 2019-11-15
Revert "Partition graph metrics update"

aarograh:
Reverts CASL/Futility#196

### DIFF
--- a/src/PartitionGraph.f90
+++ b/src/PartitionGraph.f90
@@ -1726,34 +1726,31 @@ MODULE PartitionGraph
 !> @param ecut the total weight of edges cut
 !> @param comm the total weight of communication between groups
 !>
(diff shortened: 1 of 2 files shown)


djabaay:
Guys, help me understand what the problem is on your end.

I tested this branch of Futility with MPACT master, everything configured, compiled, and all the tests passed.  This argument is optional, and I added another attribute on the partition graph type `%wtfactor` that defaults to 1.0 if it isn't specified in the input parameter list.  I only ran the basic tests because the `%metrics` call happens once for all input cases, and nothing would be gained by running the heavies.  I am rerunning everything again, with the heavies, and everything looks to still be passing.  I'm on 625 of 761, there is one unrelated failure for the 2d excore detector case.

What's the issue?

PR #213, opened 2019-12-30
Refactor MPACT Debug Build Flags

jpjones6:
Change set of gcc flags for debug build to warn for more problems and to increase performance using -Og (gcc's debug specific optimization level). Speeds up debug build by a factor of ~2 while not increasing compile time. This work is part of #6196.


stimpsonsg:
To verify...you've run what level of regression tests?

jpjones6:
Basic

stimpsonsg:
Okay...since it's debug, let's at least verify it with CONTINUOUS, but should be good.

jpjones6:
All CONTINUOUS tests passed.

stimpsonsg:
alright, let's see what's up

PR #214, opened 2019-12-31
User defined exceptions and exception tags

wgurecky:
Addresses #215

### DIFF
--- a/src/ExceptionHandler.f90
+++ b/src/ExceptionHandler.f90
@@ -862,7 +862,7 @@ MODULE ExceptionHandler
       e%surrogate%lastMesg=mesg
       nRe = SIZE(e%surrogate%exceptionRegistry)
       DO i=1,nRe
-        CALL e%surrogate%exceptionRegistry(i)%expobj%getTag(tmpTag)
+        tmpTag = e%surrogate%exceptionRegistry(i)%expobj%getTag()
         IF (tmpTag == tag) THEN
           CALL e%surrogate%exceptionRegistry(i)%expobj%onRaise( &
           e%surrogate%logFileActive, e%surrogate%logFileUnit, &
@@ -873,7 +873,7 @@ MODULE ExceptionHandler
       e%lastMesg=mesg
       nRe = SIZE(e%exceptionRegistry)
       DO i=1,nRe
-        CALL e%exceptionRegistry(i)%expobj%getTag(tmpTag)
+        tmpTag = e%exceptionRegistry(i)%expobj%getTag()
         IF (tmpTag == tag) THEN
           CALL e%exceptionRegistry(i)%expobj%onRaise( &
           e%logFileActive, e%logFileUnit, e%lastMesg)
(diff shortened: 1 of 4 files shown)


src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+!                                                                              !
+! Futility is a jointly-maintained, open-source project between the University !
+! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
+! can be found in LICENSE.txt in the head directory of this repository.        !
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief Utility module for defining exception types used by the
+!>  ExceptionHandler
+!>
+!> @author William Gurecky
+!>   @date 12/16/2019
```
aarograh: Delete the `@author` and `@date` lines.  Those probably exist throughout Futility but are no longer part of the code standard (because git)

src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+MODULE ExceptionTypes
```
aarograh: All lines between `MODULE` and `ENDMODULE` need to be unindented by 2 spaces.  The style you did matches the old way of doing it, not the current code standard.  Much of Futility needs to be updated still to match the code standard (something I'd like to do soon).

src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+  SUBROUTINE et_genprefix_absintfc(this,prefix)
+    IMPORT :: ExceptionTypeBase
+    IMPORT :: EXCEPTION_MAX_MESG_LEN
+    CLASS(ExceptionTypeBase),INTENT(INOUT) :: this
+    CHARACTER(LEN=EXCEPTION_MAX_MESG_LEN),INTENT(INOUT) :: prefix
+  ENDSUBROUTINE et_genprefix_absintfc
+ENDINTERFACE
+!
+!===============================================================================
+  CONTAINS
```
aarograh: Everything after this `CONTAINS` statement needs to be unindented an additional 2 spaces on top of the 2 I mentioned above, so that `SUBROUTINE`, `FUNCTION`, and their matching `END` statements start in the first column.

src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+  PUBLIC :: ExceptionTypeWarning
+  PUBLIC :: ExceptionTypeFatal
+  PUBLIC :: ExceptionTypeDebug
+  PUBLIC :: ExceptionTypeInformation
+  PUBLIC :: EXCEPTION_MAX_MESG_LEN
+
+  !> The maximum size of an exception message
+  INTEGER(SIK),PARAMETER :: EXCEPTION_MAX_MESG_LEN=512
+
+  TYPE, ABSTRACT :: ExceptionTypeBase
```
aarograh: Another thing we've moved away from stylisticly is having the word `Type` in extended types.  It's pretty well never helpful and just makes things wordier.  Do some search and replace to change `ExceptionType` to `Exception`, basically everywhere.

src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+      !> @copydetails ExceptionTypes::getTag_ExceptionTypeBase
+      PROCEDURE,PASS :: getTag => getTag_ExceptionTypeBase
+      !>
+      !> abstract interface functions (absintfc)
+      !>
+      !> @copybrief ExceptionTypes::et_genprefix_absintfc
+      !> @copydetails ExceptionTypes::et_genprefix_absintfc
+      PROCEDURE(et_genprefix_absintfc),DEFERRED,PASS :: genPrefix
+
+      !> @
```
aarograh: Get rid of some of the extra stuff at lines 85-87 and 91-92

src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+!===============================================================================
+  CONTAINS
+!
+!-------------------------------------------------------------------------------
+!> @brief Initialize exception
+!>
+    SUBROUTINE init_ExceptionTypeBase(this,tag)
+      CLASS(ExceptionTypeBase),INTENT(INOUT) :: this
+      INTEGER(SIK),INTENT(IN) :: tag
+      ! REQUIRE(.NOT. this%isInit)
```
aarograh: Why comment this out?

src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+    CLASS(ExceptionTypeBase),INTENT(INOUT) :: this
+    CHARACTER(LEN=EXCEPTION_MAX_MESG_LEN),INTENT(INOUT) :: prefix
+  ENDSUBROUTINE et_genprefix_absintfc
+ENDINTERFACE
+!
+!===============================================================================
+  CONTAINS
+!
+!-------------------------------------------------------------------------------
+!> @brief Initialize exception
```
aarograh: All your headers are missing `@param` lines

src/ExceptionTypes.f90
```diff
@@ -0,0 +1,370 @@
...
+#endif
+      LOGICAL(SBK),INTENT(IN) :: stopmode
+#ifdef HAVE_MPI
+      IF(stopmode) CALL MPI_Abort(MPI_COMM_WORLD,666,ierr)
+#else
+      IF(stopmode) THEN
+        STOP 666
+      ENDIF
+#endif
+    ENDSUBROUTINE exceptionStop
```
aarograh: If this needs to exist, it probably only makes sense in the ExceptionHandlerType module since it isn't specific to a particular type of exception

unit_tests/testExceptionHandler/testExceptionHandler.f90
```diff
@@ -355,4 +370,40 @@ SUBROUTINE testSetCounter()
...
+      ! check that warning counter was incremented
+      DEALLOCATE(tags)
+      DEALLOCATE(counts)
+      CALL testE4%getTagList(tags, counts)
+      ASSERT(counts(1) == 0,'tag error final count')
+      ASSERT(counts(2) == 1,'tag warning final count')
+      ASSERT(testE4%getCountByTag(EXCEPTION_ERROR) == 0,'Error')
+      ASSERT(testE4%getCountByTag(EXCEPTION_WARNING) == 1,'Warning')
+      WRITE(*,*) tags
+      WRITE(*,*) counts
```
aarograh: These `WRITE` statements will eventually need to go away

src/ExceptionHandler.f90
```diff
@@ -366,18 +442,22 @@ PURE SUBROUTINE initCounter(e)
...
-      e%nFatal=0
-      e%nDebug=0
       e%lastMesg=''
       e%logFileUnit=666
       e%stopOnError=.TRUE.
       e%logFileActive=.FALSE.
       e%quiet=(/.FALSE.,.FALSE.,.TRUE.,.FALSE./)
       e%verbose=(/.TRUE.,.TRUE.,.FALSE.,.TRUE./)
+      IF(ALLOCATED(e%exceptionRegistry)) THEN
+        DO i=1,SIZE(e%exceptionRegistry)-1
```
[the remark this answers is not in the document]
djabaay: I want to hazard a guess and say it's `SIZE()-1` because we never want to change the `EXCEPTION_FATAL_ERROR` quiet mode and verbosity, since the `FATAL_ERROR` parameter value is 5 and the number of exceptions in the registry is also 5.  Is that the case?  If so, it should really check if the `%exceptionRegistry` is the `FATAL_ERROR` and skip, right?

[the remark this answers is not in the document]
aarograh: Dan, good point.  That's probably why.

This does raise a design-related question though: if we're generalizing the exception classes, should we have one that is always handled differently with regard to verbosity?  If so, we should probably allow any of the exceptions to be controlled similarly.  For example, maybe when the exception gets initialized, it has an `immutable` attribute that is false by default, but can be set to true for things like fatal errors.

This probably doesn't need to be decided on now, but wanted to raise the question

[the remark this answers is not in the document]
djabaay: I think an `immutable` attribute makes the most sense, and was thinking along the same lines.

aarograh:
Talked to Will on Slack.  He would like to revisit this, but it's low priority right now.  We'll close this pull request and he can open a new one down the road if he has time to revisit this.

PR #219, opened 2020-01-10
Removes unused parameter definitions

HendersonSC:
This just cleans up some unused code, namely parameters which are defined but unused.


src/MatrixTypes.f90
```diff
@@ -803,7 +803,6 @@ PURE SUBROUTINE trsv_sparse(uplo,trans,diag,a,ia,ja,x,incx_in)
 !> @param beta the scalar used to scale @c C
 !>
     SUBROUTINE matmult_MatrixType(A,B,C,alpha,beta,transA,transB)
-      CHARACTER(LEN=*),PARAMETER :: myName='matmult_MatrixType'
```
aarograh: Really could just move this one down inside the preproc symbol instead

aarograh:
I'm good with these if @stimpsonsg can merge

stimpsonsg:
hmmm, we might need unused var, private module var, parameter, etc. checks in this CI too?

aarograh:
yeah, I think so.  Shane and I just talked about some tweaks to how we do that check in the MPACT CI.  Once that's nailed down, we can do the same for this one too.  Probably good as a separate merge request though.

PR #220, opened 2020-01-14
Fix indentations

aarograh:
Addresses the following types of indentations, bringing the project into conformity with the new style guide:
1. 2 spaces of indentation for all code inside MODULE and PROGRAM blocks has been removed
2. 2 spaces of indentation for all code inside a module CONTAINS statement have been removed (so that function and subroutines headers start in column 1). Per the code standard, this does not apply to CONTAINS statements inside type definitions.
3. Line continuations have been adjusted so that all continued lines are indented exactly 4 spaces from the first line of the statement (excluding special cases of indentations meant to improve readability)
4. SELECTTYPE and SELECTCASE statements have been adjusted so that the CASE statements align with their parent SELECT statements. Only the code inside the CASE should be indented instead of the CASE itself.


aarograh:
Note that I ran all the Futility tests in a serial build, parallel debug, and parallel release.  All tests passed in all 3 builds, so I don't think I fixed anything incorrectly in a way that caused defects.

PR #222, opened 2020-01-31
Add Kokkos initialize and finalize fortran interfaces

tghaddar:
1. This branch adds interfaces to call Kokkos::initialize and Kokkos::finalize from a Fortran front end. 
2. It also adds in a macro FUTILITY_HAVE_Kokkos. This in case Futility is built with Kokkos but not Trilinos. 

On Ben Collin's recommendation, I'm assigning Aaron Graham to review.


src/trilinos_interfaces/test_kokkos_init.f90
```diff
@@ -0,0 +1,19 @@
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
```
aarograh: Shouldn't this and `test_trilinios_interfaces` really live in the `unit_tests` directory instead of the `src` directory?

tghaddar: I would think so, but I just used the existing infrastructure. I can move both `test_trilinos_interfaces` and `test_kokkos_init` into the `unit_tests` directory if you think that's best.

aarograh: It makes sense to me, but I'm not sure if there was a particular reason they were placed here instead.  @stimpsonsg , @bscollin any comment on that?

stimpsonsg: yeah, I'd make a directory in `Futility/unit_tests` to house these.  I don't really love that `trilinos_interfaces` is a separate directory within source, but not a huge deal, so long as it's consistent.

tghaddar: Ok this change is pushed up.

unit_tests/trilinos_interfaces/testKokkosInit/testKokkosInit.f90
```diff
@@ -10,10 +10,13 @@ PROGRAM test_kokkos_init
...
   USE trilinos_interfaces
   IMPLICIT NONE
-
+#ifdef FUTILITY_HAVE_Kokkos
   !Initializing Kokkos.
   CALL InitializeKokkos(4,-1)
+  WRITE(*,*) "Kokkos initialization successful."
   !Finalizing Kokkos.
   CALL FinalizeKokkos()
+  WRITE(*,*) "Kokkos finalization successful."
```
aarograh: Do we want these 2 write statements on master?

aarograh: Didn't realize I was looking at a test, so I actually don't really care about these.  Let's just go ahead and push this in.

stimpsonsg:
Let's hold this until the build issues from the previous MR are resolved.

PR #224, opened 2020-02-06
New preconditioners

mattkab2:
This code adds new preconditioners (most notably Yunlin's RSOR) in serial and parallel
It also slightly changes how they are used and called in the LinearSolverTypesIterative class.


mattkab2:
Pull request is now sane again.
Merging/rebasing was tricky because it was built upon the multiple commits that lead to the NativeParallelDatatypes Branch, which were squashed. So merging or rebasing would try to apply all changes from that branch twice.

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,699 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+    SELECT TYPE(LpU => thisPC%LpU); TYPE IS(DistributedBlockBandedMatrixType)
+      CALL LpU%setBlockMask(.TRUE.)
+    ENDSELECT
+
+    WRITE(*,*) "w",w%b
+    WRITE(*,*) "v",v%b
+    CALL BLAS_matvec(THISMATRIX=thisPC%LpU,X=w,Y=v,BETA=1.0_SRK,&
+                      ALPHA=-thisPC%omega)
+    WRITE(*,*) "w",w%b
+    WRITE(*,*) "v",v%b
```
nfherrin: These checking write statements should probably be removed.

mattkab2: Cool, these will be removed. My mistake.
The RSORPrecon and RSORPreconParallel Unit tests should probably remain separate, because one is designed to be called with MPI and the other is not. So I think that will stay the same as it always has.

mattkab2:
Before being merged, I will need to make some checks to ensure the NATIVE -> LS_NATIVE parameter name change does not break MPACT

mattkab2:
I've run the nightly tests on MPACT compiled with this branch of Futility, and they all pass (except for the single test that requires COBRATF). Log attached.
[ctest.log](https://github.com/CASL/Futility/files/4182476/ctest.log)

src/LinearSolverTypes.f90
```diff
@@ -338,25 +350,41 @@ SUBROUTINE init_LinearSolverType_Base(solver,Params,A)
...
+  IF(validParams%has('LinearSolverType->PC->PreCondType')) THEN
+    CALL validParams%get('LinearSolverType->PC->PreCondType',pListPtr)
+    pcPList = pListPtr
+    CALL validParams%get('LinearSolverType->PC->PreCondType->pcType',PreCondType)
+
+    SELECTTYPE(solver); TYPE IS(LinearSolverType_Iterative)
+      solver%pcIters = -1_SIK
+      IF (validParams%has('LinearSolverType->PC->PreCondType->pcIters')) &
+        CALL validParams%get('LinearSolverType->PC->PreCondType->pcIters',solver%pcIters)
+      solver%pcSetup = 0_SIK
```
djabaay: Don't need the _SIK value.

src/LinearSolverTypes.f90
```diff
@@ -338,25 +350,41 @@ SUBROUTINE init_LinearSolverType_Base(solver,Params,A)
...
+        CALL validParams%get('LinearSolverType->PC->PreCondType->pcSetup',solver%pcSetup)
+    ENDSELECT
   ELSE
-    PreCondType='NOPC'
+    CALL pcPList%clear()
+    CALL pcPList%add('PreCondType->pcType','DEFAULT')
+    SELECTTYPE(solver); TYPE IS(LinearSolverType_Iterative)
+      PreCondType = 'DEFAULT'
+      solver%pcIters = -1_SIK
+      solver%pcSetup = 0_SIK
```
djabaay: delete _SIK

src/LinearSolverTypes.f90
```diff
@@ -656,17 +662,65 @@ SUBROUTINE init_LinearSolverType_Base(solver,Params,A)
...
+            CALL solver%PreCondType%init(solver%A,solver%pcParams)
+            CALL solver%PreCondType%setup()
+          ENDIF
+          CALL solver%pcParams%clear()
+        ELSE
+          IF (solver%TPLType /= PETSC) CALL solver%PreCondType%setup()
+        ENDIF
+        ! If pcsetup was 0, we've successfully set up the PC
+        ! Make it huge so it is never set up again.
+        IF (solver%pcsetup == 0) solver%pcsetup = HUGE(0_SIK)
```
djabaay: delete _SIK

src/LinearSolverTypes.f90
```diff
@@ -1073,6 +1127,9 @@ SUBROUTINE solve_LinearSolverType_Iterative(solver)
       ENDSELECT
 
     CASE(GMRES)
+      ! Ensure setupPrecond has been called
+      IF (.NOT. ALLOCATED(solver%preCondType) &
+          .AND. solver%pcTypeName /= "NOPC") CALL solver%setupPC()
```
djabaay: Would it be useful to add a logical to the base or specific linear solver type for `%hasSetPC` or something like that?  Then you can just check that value instead?  Or maybe on the preconditioner type?

src/PreconditionerTypes.f90
```diff
@@ -107,6 +287,20 @@ SUBROUTINE precond_applyLU_absintfc(thisPC,v)
...
   ENDSUBROUTINE precond_applyLU_absintfc
+
+  SUBROUTINE precond_applySOR_absintfc(thisPC,v)
+    IMPORT :: SOR_PreCondType,VectorType
+    CLASS(SOR_PreCondType),INTENT(INOUT) :: thisPC
+    CLASS(VectorType),INTENT(INOUT) :: v
+  ENDSUBROUTINE precond_applySOR_absintfc
+
+#ifdef HAVE_MPI
+  SUBROUTINE precond_applyDistributedSOR_absintfc(thisPC,v)
```
djabaay: Need abstract interface definition

src/PreconditionerTypes.f90
```diff
@@ -119,13 +313,26 @@ SUBROUTINE precond_LU_absintfc(thisPC)
     IMPORT :: LU_PreCondType
     CLASS(LU_PreCondType),INTENT(INOUT) :: thisPC
   ENDSUBROUTINE precond_LU_absintfc
+
+  SUBROUTINE precond_SOR_absintfc(thisPC)
```
djabaay: Need abstract interface definition

src/PreconditionerTypes.f90
```diff
@@ -201,8 +410,8 @@ SUBROUTINE init_LU_PreCondtype(thisPC,A)
   !   ALLOCATE(PETScMatrixType :: thisPC%U)
   !
   !   !initialize L and U
-  !   SELECTTYPE(U => thisPC%U); TYPE IS(PETScMatrixType)
-  !     SELECTTYPE(L => thisPC%L); TYPE IS(PETScMatrixType)
+  !   SELECT TYPE(U => thisPC%U); TYPE IS(PETScMatrixType)
+  !     SELECT TYPE(L => thisPC%L); TYPE IS(PETScMatrixType)
   !       ! Initialize L and U (add preallocation eventually)
   !       CALL PL%add('MatrixType->matType',SPARSE)
   !       CALL PL%add('MatrixType->n',mat%n)
```
djabaay: Is this block still needed?  Or is it going to be uncommented in the next PR?

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+  CLASS(MatrixType),ALLOCATABLE,TARGET,INTENT(IN),OPTIONAL :: A
+  TYPE(ParamType),INTENT(IN),OPTIONAL :: params
+  TYPE(ParamType)::PListMat_LU
+  INTEGER(SIK)::k
+
+  !required statuses
+  REQUIRE(.NOT. thisPC%isinit)
+  REQUIRE(PRESENT(A))
+  REQUIRE(ALLOCATED(A))
+  REQUIRE(A%isInit)
```
djabaay: Why require this if `A` is declared `OPTIONAL`?

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+
+  !makes sure that the number of blocks is valid
+  REQUIRE(thisPC%numBlocks .GT. 0)
+
+  !makes sure that the number of blocks is valid
+  REQUIRE(MOD(thisPC%A%n,thisPC%numBlocks) .EQ. 0)
+
+  !makes sure that the value of omega is valid
+  REQUIRE(thisPC%omega .LE. 2)
+  REQUIRE(thisPC%omega .GE. 0)
```
djabaay: All of these require statements, I think they need to use the >=, <=, ==, <, and >.

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+  ENDDO
+
+  SELECT TYPE(mat => thisPC%A)
+  TYPE IS(DenseSquareMatrixType)
+    ALLOCATE(DenseSquareMatrixType :: thisPC%LpU)
+    ! Assign A to LpU
+    SELECT TYPE(LpU => thisPC%LpU); TYPE IS(DenseSquareMatrixType)
+      LpU=mat
+    ENDSELECT
+    REQUIRE(thisPC%LpU%isInit)
```
djabaay: If at the top `REQUIRE(A%isInit)`, `thisPC%A => A`, so `%A` is init, and we assign mat to `%thisPC%A`, and then we assign mat to LpU, requiring `LpU%isInit` is redundant.  We can get rid of all of them.

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+    REQUIRE(thisPC%LU(k)%isInit)
+  ENDDO
+
+  !setup the Upper and Lower portion of the diagonal
+  DO k=1,thisPC%numBlocks
+    DO i=1,thisPC%blockSize
+      DO j=1,thisPC%blockSize
+        CALL thisPC%A%get((k-1)*thisPC%blockSize+i,(k-1)*thisPC%blockSize+j,tempreal)
+        CALL thisPC%LU(k)%set(i,j,tempreal)
+        IF(tempreal .NE. 0.0_SRK)THEN
```
djabaay: change `.NE.` to `/=`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+    thisPC%blockOffset = A%iOffsets(rank+1)/A%blockSize
+    thisPC%comm = A%comm
+  CLASS DEFAULT
+    CALL ePreCondType%raiseError('Incorrect input to '//modName//'::'//myName// &
+        ' - RSOR Preconditioners are not supported for input matrix type!')
+  ENDSELECT
+
+  !makes sure that the value of omega is valid
+  REQUIRE(thisPC%omega .LE. 2)
+  REQUIRE(thisPC%omega .GE. 0)
```
djabaay: change to <= and >=

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+  ENDDO
+
+  SELECT TYPE(mat => thisPC%A)
+  TYPE IS(DistributedBandedMatrixType)
+    ALLOCATE(DistributedBandedMatrixType :: thisPC%LpU)
+    ! Assign A to LpU
+    SELECT TYPE(LpU => thisPC%LpU); TYPE IS(DistributedBandedMatrixType)
+      LpU = mat
+    ENDSELECT
+    REQUIRE(thisPC%LpU%isInit)
```
djabaay: Delete redundant require.

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+  ENDSELECT
+ENDSUBROUTINE init_DistributedSOR_PreCondtype
+!
+!-------------------------------------------------------------------------------
+!> @brief Clears the Distributed SOR Preconditioner Type with a parameter list
+!> @param thisPC The preconditioner to act on
+!>
+SUBROUTINE clear_DistributedSOR_PreCondtype(thisPC)
+  CLASS(DistributedSOR_PrecondType),INTENT(INOUT) :: thisPC
+  INTEGER(SIK)::i
```
djabaay: spaces around `::`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+!> @param A The matrix to precondition
+!> @param params The parameter list
+!>
+SUBROUTINE init_DistributedSOR_PreCondtype(thisPC,A,params)
+  CHARACTER(LEN=*),PARAMETER :: myName='init_RSOR_PreCondType'
+  CLASS(DistributedSOR_PrecondType),INTENT(INOUT) :: thisPC
+  CLASS(MatrixType),ALLOCATABLE,TARGET,INTENT(IN),OPTIONAL :: A
+  TYPE(ParamType),INTENT(IN),OPTIONAL :: params
+  TYPE(ParamType)::PListMat_LU
+  INTEGER(SIK)::k,mpierr,rank
```
djabaay: spaces around `::`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+!> @param thisPC The preconditioner to act on
+!> @param v The vector to apply the preconditioner to to act on
+!>
+SUBROUTINE apply_RSOR_PreCondType(thisPC,v)
+  CLASS(RSOR_PrecondType),INTENT(INOUT) :: thisPC
+  CLASS(Vectortype),ALLOCATABLE,INTENT(INOUT) :: v
+  CHARACTER(LEN=*),PARAMETER :: myName='apply_RSOR_PreCondType'
+  TYPE(RealVectorType)::w
+  TYPE(ParamType)::PListVec_RSOR
+  INTEGER(SIK)::k,lowIdx,highIdx
```
djabaay: spaces around `::`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+ENDSUBROUTINE clear_SOR_PreCondtype
+!
+!-------------------------------------------------------------------------------
+!> @brief Sets up the RSOR Preconditioner Type with a parameter list
+!> @param thisPC The preconditioner to act on
+!>
+SUBROUTINE setup_RSOR_PreCondtype(thisPC)
+  CLASS(RSOR_PrecondType),INTENT(INOUT) :: thisPC
+  INTEGER(SIK)::k,i,j
+  REAL(SRK)::tempreal
```
djabaay: spaces around `::`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+  ENDSELECT
+ENDSUBROUTINE init_SOR_PreCondtype
+!
+!-------------------------------------------------------------------------------
+!> @brief Clears up the SOR Preconditioner Type with a parameter list
+!> @param thisPC The preconditioner to act on
+!>
+SUBROUTINE clear_SOR_PreCondtype(thisPC)
+  CLASS(SOR_PrecondType),INTENT(INOUT) :: thisPC
+  INTEGER(SIK)::i
```
djabaay: spaces around `::`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+!> @param A The matrix to precondition
+!> @param params The parameter list
+!>
+SUBROUTINE init_SOR_PreCondtype(thisPC,A,params)
+  CHARACTER(LEN=*),PARAMETER :: myName='init_RSOR_PreCondType'
+  CLASS(SOR_PrecondType),INTENT(INOUT) :: thisPC
+  CLASS(MatrixType),ALLOCATABLE,TARGET,INTENT(IN),OPTIONAL :: A
+  TYPE(ParamType),INTENT(IN),OPTIONAL :: params
+  TYPE(ParamType)::PListMat_LU
+  INTEGER(SIK)::k
```
djabaay: spaces around `::`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+
+!
+!-------------------------------------------------------------------------------
+!> @brief Sets up the Distributed RSOR Preconditioner Type with a parameter list
+!> @param thisPC The preconditioner to act on
+!>
+SUBROUTINE setup_DistributedRSOR_PreCondtype(thisPC)
+  CLASS(DistributedRSOR_PrecondType),INTENT(INOUT) :: thisPC
+  INTEGER(SIK)::k,i,j
+  REAL(SRK)::tempreal
```
djabaay: spaces around `::`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+    ! blocks do not need to be set to zero because we can set the block mask
+  CLASS DEFAULT
+    !setup the Upper and Lower portion of the diagonal
+    DO k=1,thisPC%nLocalBlocks
+      DO i=1,thisPC%blockSize
+        DO j=1,thisPC%blockSize
+          CALL A%get((k+thisPC%blockOffset-1)*thisPC%blockSize+i, &
+              (k+thisPC%blockOffset-1)*thisPC%blockSize+j,tempreal)
+          CALL thisPC%LU(k)%set(i,j,tempreal)
+          IF(tempreal .NE. 0.0_SRK)THEN
```
djabaay: `.NE.` to `/=`

src/PreconditionerTypes.f90
```diff
@@ -343,6 +550,695 @@ SUBROUTINE setup_ILU_PreCondtype(thisPC)
...
+!> @param thisPC The preconditioner to act on
+!> @param v The vector to apply the preconditioner to to act on
+!>
+SUBROUTINE apply_DistributedRSOR_PreCondType(thisPC,v)
+  CLASS(DistributedRSOR_PrecondType),INTENT(INOUT) :: thisPC
+  CLASS(Vectortype),ALLOCATABLE,INTENT(INOUT) :: v
+  CHARACTER(LEN=*),PARAMETER :: myName='apply_DistributedRSOR_PreCondType'
+  TYPE(NativeDistributedVectorType)::w
+  TYPE(ParamType)::PListVec_RSOR
+  INTEGER(SIK)::k,lowIdx,highIdx
```
djabaay: spaces around `::`

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -2785,7 +2787,8 @@ SUBROUTINE testIterativeSolve_GMRES()
   thisX=1.0_SRK
 
   SELECTTYPE(thisLS); TYPE IS(LinearSolverType_Iterative)
-      CALL thisLS%setX0(thisX)
+    CALL thisLS%setX0(thisX)
+    CALL thisLS%pcparams%edit(6)
```
djabaay: still need the edit?

unit_tests/testRSORprecon/testRSORPrecon.f90
```diff
@@ -0,0 +1,383 @@
...
+               0_SRK,  0_SRK,  0_SRK, 16_SRK,  0_SRK,  0_SRK,-64_SRK, 16_SRK,  0_SRK,&
+               0_SRK,  0_SRK,  0_SRK,  0_SRK, 16_SRK,  0_SRK, 16_SRK,-64_SRK, 16_SRK,&
+               0_SRK,  0_SRK,  0_SRK,  0_SRK,  0_SRK, 16_SRK,  0_SRK, 16_SRK,-64_SRK/)
+
+  !setup the dense matrixk and count the number of nonzeroentries
+  numnonzero=0
+  DO i=1,9
+    DO j=1,9
+      CALL testDenseMatrix%set(i,j,tmpreal1((i-1)*9+j))
+      IF(tmpreal1((i-1)*9+j) .NE. 0_SRK) numnonzero=numnonzero+1
```
djabaay: `.NE.` to `/=`

unit_tests/testRSORprecon/testRSORPrecon.f90
```diff
@@ -0,0 +1,383 @@
...
+  ENDDO
+
+  !setup the sparse version of the matrix
+  CALL PListMat%add('MatrixType->nnz',numnonzero)
+  ALLOCATE(SparseMatrixType :: testSparseMatrix)
+  CALL testSparseMatrix%init(PListMat)
+  SELECT TYPE(testSparseMatrix); TYPE IS(SparseMatrixType)
+    DO i=1,9
+      DO j=1,9
+        IF(tmpreal1((i-1)*9+j) .NE. 0_SRK)THEN
```
djabaay: `.NE.` to `/=`

unit_tests/testRSORprecon/testRSORPrecon.f90
```diff
@@ -0,0 +1,383 @@
...
+    ENDDO
+  ENDSELECT
+
+  !setup the banded version of the matrix
+  ALLOCATE(BandedMatrixType :: testBandedMatrix)
+  CALL testBandedMatrix%init(PListMat)
+  SELECT TYPE(testBandedMatrix); TYPE IS(BandedMatrixType)
+    DO i=1,9
+      DO j=1,9
+        IF(tmpreal1((i-1)*9+j) .NE. 0_SRK) THEN
```
djabaay: `.NE.` to `/=`

unit_tests/testRSORprecon/testRSORPreconParallel.f90
```diff
@@ -0,0 +1,302 @@
...
+  CALL PListRSOR%add('PreCondType->comm',PE_COMM_WORLD)
+  CALL PListVec%add('VectorType->n',9_SIK)
+  CALL PListVec%add('VectorType->chunkSize',3_SIK)
+  CALL PListVec%add('VectorType->MPI_Comm_ID',PE_COMM_WORLD)
+  CALL PListVec%add('VectorType->nlocal',-1)
+  CALL testVector%init(PListVec)
+  CALL workVector%init(PListVec)
+  CALL refVector%init(PListVec)
+  CALL PListMat%add('MatrixType->n',9_SIK)
+  CALL PListMat%add('MatrixType->m',9_SIK)
```
djabaay: remove _SIKs

unit_tests/testRSORprecon/testRSORPreconParallel.f90
```diff
@@ -0,0 +1,302 @@
...
+  ALLOCATE(DistributedBandedMatrixType :: testBandedMatrix)
+  ALLOCATE(DistributedBlockBandedMatrixType :: testBlockBandedMatrix)
+  CALL testBandedMatrix%init(PListMat)
+  CALL testBlockBandedMatrix%init(PListMat)
+  SELECT TYPE(testBandedMatrix); TYPE IS(DistributedBandedMatrixType)
+    SELECT TYPE(testBlockBandedMatrix); TYPE IS(DistributedBlockBandedMatrixType)
+      DO i=1,9
+        IF (rank==0) THEN
+          DO j=1,6
+            IF(tmpreal1((i-1)*9+j) .NE. 0_SRK) THEN
```
djabaay: `.NE.` to `/=`

unit_tests/testRSORprecon/testRSORPreconParallel.f90
```diff
@@ -0,0 +1,302 @@
...
+              CALL testBandedMatrix%set(i,j,tmpreal1((i-1)*9+j))
+              CALL testBlockBandedMatrix%set(i,j,tmpreal1((i-1)*9+j))
+            ENDIF
+          ENDDO
+         ELSE
+          DO j=7,9
+            IF(tmpreal1((i-1)*9+j) .NE. 0_SRK)THEN
```
djabaay: `.NE.` to `/=`

src/PreconditionerTypes.f90
mattkab2: The abstract interface specifies A as optional, so this subroutine must match. However, the SOR preconditioners require A to be present, so we have to require that A is passed in.

unit_tests/testLinearSolver/testLinearSolver.f90
mattkab2: Nope, removing

src/LinearSolverTypes.f90
mattkab2: Yeah, I debating doing that. I'll add it and check that instead.

src/PreconditionerTypes.f90
mattkab2: No, this stuff was here before and I didn't actually change it. It just got caught by a find/replace for SELECT TYPE. It's part of the original LU_PreCondtype.
I'll drop the change to the commented lines

HendersonSC: Or just deleted the dead code?

src/PreconditionerTypes.f90
mattkab2: This is correct for the serial types. In all cases, the serial types require a duplication of the matrix (without the dense diagonal blocks) to work. The initialization sets LpU (the duplicated matrix) to A because of this. The setup routine then goes through LpU and makes the necessary changes, which cannot be made to A without loss of information. So LpU itself has to stay.

This would be expensive and not worth our time except for the fact that the blockbanded matrix is designed in a specific way to avoid this copying. In that case, LpU is set as a pointer => A because A will have a routine to "mask" the dense blocks without copying.

I will get rid of the extraneous require statements though.

src/PreconditionerTypes.f90
mattkab2: I can but I didn't write it and don't know for sure the reason it was commented out.

According to git blame it was commented out on 2017-07-18 by Mitchell Young as part of a MatrixTypes refactor.

HendersonSC: My input would be to delete. I would say after nearly three years, the functionality has either been replaced or is not needed.

src/PreconditionerTypes.f90
```diff
@@ -58,43 +74,206 @@ MODULE PreconditionerTypes
...
+    !> @copybrief MatrixTypes::setup_DistributedRSOR_PreCondType
+    !> @copydetails MatrixTypes::setup_DistributedRSOR_PreCondType
+    PROCEDURE,PASS :: setup => setup_DistributedRSOR_PreCondType
+    !> @copybrief MatrixTypes::apply_DistributedRSOR_PreCondType
+    !> @copydetails MatrixTypes::apply_DistributedRSOR_PreCondType
+    PROCEDURE,PASS :: apply => apply_DistributedRSOR_PreCondType
+ENDTYPE DistributedRSOR_PreCondType
+#endif
+
+TYPE,EXTENDS(PreConditionerType) :: Jacobi_PreCondType
```
djabaay: Needs a type brief.

src/PreconditionerTypes.f90
```diff
@@ -58,43 +74,206 @@ MODULE PreconditionerTypes
...
+    !> @copydetails PreconditionerTypes::setup_Jacobi_PreCondType
+    PROCEDURE,PASS :: setup => setup_Jacobi_PreCondType
+    !> @copybrief PreconditionerTypes::apply_Jacobi_PreCondType
+    !> @copydetails PreconditionerTypes::apply_Jacobi_PreCondType
+    PROCEDURE,PASS :: apply => apply_Jacobi_PreCondType
+
+ENDTYPE Jacobi_PreCondType
+
+#ifdef HAVE_MPI
+TYPE,EXTENDS(DistributedPreCond) :: DistributedJacobi_PreCondType
```
djabaay: Needs a type brief.

src/PreconditionerTypes.f90
djabaay: Can we make it required in the interface?  Are there preconditioners that don't require a matrix?

mattkab2:
Regarding the optional matrix question, the answer is that unfortunately one of the preconditioner types implemented in Futility is a Shell preconditioner for PETSc, which does not have a Futility matrix but a PETSc one. Since these are not type-compatible, we cannot make the matrix a requirement.

src/PreconditionerTypes.f90
aarograh: I did a brief glance and it seems that all of them assume the presence of A.  And quite frankly, I think it's very reasonable to assume that when setting up a preconditioner, a matrix is also available.  Dan, I think your suggestion is right.  Let's just make it required.  If some preconditioner happens to ignore it (I don't even know how that would be possible, really), then that's fine.

djabaay: Quoting from Matt:
Regarding the optional matrix question, the answer is that unfortunately one of the preconditioner types implemented in Futility is a Shell preconditioner for PETSc, which does not have a Futility matrix but a PETSc one. Since these are not type-compatible, we cannot make the matrix a requirement.

aarograh: @mattkab2 I see 'PETSC_PCSHELL_PC` in `PreconditionerTypes` which is a `CLASS(PreConditionerType)`, which owns a `CLASS(MatrixType),POINTER :: A`.  There is also a `PETScMatrixType` which has the futility `MatrixType` as its base class.  I'm not sure I'm understanding what the issue is?

mattkab2: I've dug a little deeper, because your comment is correct. The trouble actually comes from the intended use case of the SHELL_PC. From what I gather in the unit tests, the intent is to allow the user to define the intended init/clear/setup/apply functions. This may be confusing or strange but it is the functionality expected and tested by the unit tests.

Within the unit test, the preconditioner is initialized without a matrix, which is set only after the fact (see lines 440-560 in the testPreconditioner unit test). Since this is the one and only test of PCShell I have to assume that this is the intended use case and the preconditioner must be able to be initialized without an A.

If we want to change this, I'm open to it but I'm not sure what repercussions it might have to other parts of the code. They are probably minimal, but I'm not sure. I'm also not certain it should be part of this work.

aarograh: I did some searching.  I expect virtually 0 side effects.  However, you're right that it doesn't need to go in with this work.  I may do a follow-up merge to clean up the weird interface.

PR #229, opened 2020-02-18
Add build flag to instruct gcc to use pipes instead of intermediate files

jpjones6:
Using pipes (keeping needed files loaded in memory instead of writing to disk) tends to speed up compilation as less disk writing/reading is done. It does, however, tend to increase memory usage during compilation. The speedup is machine dependent.


aarograh:
To questions to clarify:
1. Did you succeed in getting this flag to filter down into Trilinos as well?  I assume using the "general flags" instead of the "cxx flags" might accomplish that, but want to double check.
2. If the machine does not have enough memory, does the behavior automatically just revert to as if the `-pipe` flag were not used?

cmake/tribits/core/package_arch/TribitsDefineStandardCompileVars.cmake
```diff
@@ -56,15 +56,15 @@ MACRO(TRIBITS_DEFINE_STANDARD_COMPILE_FLAGS_VARS  ENABLE_SHADOWING_WARNINGS)
   # Setup and general flags
   #
 
-  SET(GENERAL_BUILD_FLAGS) # Applies to all builds, period
+  SET(GENERAL_BUILD_FLAGS "-pipe") # Applies to all builds, period
```
aarograh: So I'm concerned about changing this file.  I believe this is actually part of TriBITS itself.  I'm not sure what downstream effects this could potentially have... @bscollin @stimpsonsg care to comment?  You guys know a bit more about tribits than I do

aarograh: Is it possible to set these flags (or modify them) via commandline instead?

jpjones6: > To questions to clarify:
> 
> 1. Did you succeed in getting this flag to filter down into Trilinos as well?  I assume using the "general flags" instead of the "cxx flags" might accomplish that, but want to double check.
> 2. If the machine does not have enough memory, does the behavior automatically just revert to as if the `-pipe` flag were not used?

1) Yes it does filter down to trilinos doing it this way. Adding it to the general flags obviates the need to change more files to add the -pipe flag such as Futility_Configurations

2) That is a serious question. Unfortunately the gcc manual does not specify. It either page files or reverts to intermediates, I don't know which actually happens.

stimpsonsg:
Can we set this from the command line instead?

cmake/tribits/core/package_arch/TribitsDefineStandardCompileVars.cmake
aarograh: We've had several offline conversations about this, but just for the sake of thoroughness, here's where we've arrived:

1. The flag can be set via commandline.  We're going to just add it to some of the CI build scripts for now rather than modifying any tribits/cmake code.
2. If the machine runs out of memory with this option, it will simply start using swap space (tested locally on my laptop).  However, the amount of memory use for a build isn't that crazy.  Something else had to be pushing it close to the edge for a single build to run out of memory (I saw < 8 GB as the highwater mark for my build).

aarograh:
We determined how to do this via commandline instead, so this change isn't needed.  Furthermore, after testing on the MPACT CI, it seems that any differences from using `-pipe` are in the noise, so it's likely not worth pursuing any further.

Closing this PR since it's unneeded.

PR #230, opened 2020-02-26
Implements string partition function

HendersonSC:
Implements a StringType partition functions similar to that encountered in python.

### DIFF
--- a/src/Strings.f90
+++ b/src/Strings.f90
@@ -79,7 +79,7 @@ TYPE :: StringType
     GENERIC :: split => split_string, split_string_space
     !> copybrief Strings::partition
     !> copydetails Strings::partition
-    PROCEDURE,PASS :: partition => partition
+    PROCEDURE,PASS :: partition
     !> copybrief Strings::rPartition
     !> copydetails Strings::rPartition
     PROCEDURE,PASS :: rPartition
@@ -249,8 +249,9 @@ ENDSUBROUTINE clear_str
 !-------------------------------------------------------------------------------
 !> @brief breaks a string into partitions at the first instance of a delimiter
 !> @param this the string being partitioned
-!> @param sep the delimter used for partitioning
-!>
+!> @param sep the delimiter used for partitioning
+!> @returns tokens size 3 array containing the characters before the delimiter,
+!> the delimiter, and the characters after the delimiter in that order
 PURE FUNCTION partition(this,sep) RESULT(tokens)
   CLASS(StringType),INTENT(IN) :: this
   CHARACTER(LEN=*),INTENT(IN) :: sep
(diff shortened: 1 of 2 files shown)


src/Strings.f90
```diff
@@ -241,6 +247,48 @@ SUBROUTINE clear_str(this)
 ENDSUBROUTINE clear_str
 !
 !-------------------------------------------------------------------------------
+!> @brief breaks a string into partitions at the first instance of a delimiter
+!> @param this the string being partitioned
+!> @param sep the delimter used for partitioning
+!>
```
aarograh: Both versions would benefit from a bit more description on the return values, especially for those (like me) who aren't familiar with `partition`.  Just state clearly what each element of `tokens` contains, and briefly describe the results for corner cases such as `sep=''` or if `sep` is not found in `this`.

src/Strings.f90
```diff
@@ -241,6 +247,48 @@ SUBROUTINE clear_str(this)
...
+PURE FUNCTION partition(this,sep) RESULT(tokens)
+  CLASS(StringType),INTENT(IN) :: this
+  CHARACTER(LEN=*),INTENT(IN) :: sep
+  TYPE(StringType) :: tokens(3)
```
aarograh: How does this behave if you assign the return value to an unallocated allocatable array of string types?  It might be good to make the return value allocatable instead, but I'm not sure if it's actually necessary.

unit_tests/testStrings/testStrings.f90
```diff
@@ -589,7 +589,75 @@ SUBROUTINE testStrFunct()
...
+    ASSERT_EQ(SIZE(str_list),3,'string with multiple delimter found')
+    ASSERT_EQ(CHAR(str_list(1)),'abc','string with multiple delimter found')
+    ASSERT_EQ(CHAR(str_list(2)),'---','string with multiple delimter found')
+    ASSERT_EQ(CHAR(str_list(3)),'def---ghi','string with multiple delimter found')
+    str_list = str%rPartition('---')
+    ASSERT_EQ(SIZE(str_list),3,'string with multiple delimter found')
+    ASSERT_EQ(CHAR(str_list(1)),'abc---def','string with multiple delimter found')
+    ASSERT_EQ(CHAR(str_list(2)),'---','string with multiple delimter found')
+    ASSERT_EQ(CHAR(str_list(3)),'ghi','string with multiple delimter found')
+
```
aarograh: Could you also add tests for 2 more cases: when the delimiter starts the string, and when the delimiter ends the string?  I think it should be fine, but just want to make sure there are no indexing gotchas

src/Strings.f90
HendersonSC: It seems to work in the unit test, which know is not the strongest endorsement, but I think it should be fine...
```
Note that even if A is an allocatable array, A(:) is not an allocatable array; it is an
array subsection.
integer, dimension(:), allocatable :: a
a = [ 1, 2, 3 ] ! Size of a becomes 3
a = [ 1, 2, 3, 4 ] ! Size of a becomes 4
a(:) = [ 1, 2, 3 ] ! Illegal because a(:) is size 4 and not allocatable
```

If a were a string type then when the function returns it should be a = [str1,str2,str3]. So the key is to not put `(:)` unless a is size 3.

aarograh: Ok, that's fair.  I'm still not used to relying on that because of old compilers and the fact that it has tripped me up when I've done it on accident, but I suppose that's how the language standard is written, and the compiler versions we use now tend to be fine with that implicit allocation.  So I'm good leaving it as is.

PR #231, opened 2020-02-26
Native GMRES

mattkab2:
Improves the native GMRES capability of Futility.


src/LinearSolverTypes.f90
```diff
@@ -1125,6 +1118,9 @@ SUBROUTINE solve_LinearSolverType_Iterative(solver)
       ENDSELECT
 
     CASE(GMRES)
+      ! Ensure setupPrecond has been called
+      IF (.NOT. ALLOCATED(solver%preCondType) &
+          .AND. solver%pcTypeName /= "NOPC") CALL solver%setupPC()
```
djabaay: Didn't you add a logical on the preconditioner type for whether it is set up or not?

src/LinearSolverTypes.f90
```diff
@@ -1431,17 +1427,53 @@ SUBROUTINE getResidual_LinearSolverType_Iterative(solver,resid)
...
+#ifdef HAVE_MPI
+      TYPE IS(NativeDistributedVectorType)
+        SELECTTYPE(resid)
+        TYPE IS(NativeDistributedVectorType)
+          resid%b=-b%b
+          CALL BLAS_matvec(THISMATRIX=solver%A,X=solver%X,Y=resid,ALPHA=1.0_SRK,BETA=1.0_SRK)
+        CLASS DEFAULT
+          CALL eLinearSolverType%raiseError('Incorrect call to '// &
+              modName//'::getResidual'//' - Type mismatch between getResidual argument and solver datatypes')
+        ENDSELECT
```
djabaay: Would it be better to put the serial portion of the code into an `#else` block?

src/LinearSolverTypes.f90
```diff
@@ -1597,228 +1629,237 @@ SUBROUTINE solveBiCGSTAB(solver)
...
+    ALLOCATE(RealVectorType :: u)
+    ALLOCATE(RealVectorType :: Vy)
+#ifdef HAVE_MPI
+  TYPE IS(NativeDistributedVectorType)
+    CALL vecPlist%add('VectorType -> n',thisLS%A%n)
+    CALL vecPlist%add('VectorType -> MPI_Comm_ID',thisLS%MPIparallelEnv%comm)
+    CALL vecPlist%add('VectorType -> chunkSize',x%chunkSize)
+    CALL vecPlist%add('VectorType -> nlocal',SIZE(x%b))
+    ALLOCATE(NativeDistributedVectorType :: u)
+    ALLOCATE(NativeDistributedVectorType :: Vy)
```
djabaay: Another possible `#else` for the serial portion?

src/LinearSolverTypes.f90
```diff
@@ -1597,228 +1629,237 @@ SUBROUTINE solveBiCGSTAB(solver)
...
-  CALL PreCondType%apply(u)
-  CALL LNorm(u%b,2,beta)
-  solver%iters=0
+  ! Compute norm of b
+  norm_b = BLAS_nrm2(thisLS%b)
+
+  ! Check if solving null system
+  IF (norm_b <= EPSILON(0.0_SRK)) THEN
+    CALL thisLS%X%set(0.0_SRK)
+    thisLS%residual = 0.0
```
djabaay: _SRK

src/LinearSolverTypes.f90
```diff
@@ -1597,228 +1629,237 @@ SUBROUTINE solveBiCGSTAB(solver)
...
+  ALLOCATE(givens_cos(thisLS%nRestart))
+  ALLOCATE(givens_sin(thisLS%nRestart))
+  ALLOCATE(f(thisLS%nRestart))
+  ALLOCATE(h(thisLS%nRestart))
+  ALLOCATE(orthogReq(thisLS%nRestart))
+
+  ! Initialize relevant quantities
+  currResid = norm_r0
+  divTmp = -1.0_SRK/norm_r0
+  R(:,:) = 0.0
```
djabaay: _SRK

src/LinearSolverTypes.f90
```diff
@@ -1597,228 +1629,237 @@ SUBROUTINE solveBiCGSTAB(solver)
...
+      ENDSELECT
+      EXIT
+    ENDIF
+  ENDDO
+
+  !CALL thisLS%getResidual(u)
+  !norm_r0 = BLAS_dot(u%b,u%b)
+  !CALL thisLS%MPIparallelEnv%allReduce_scalar(norm_r0)
+  !norm_r0 = SQRT(norm_r0)
+  !thisLS%residual = norm_r0
```
djabaay: If these are needed, delete.

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -2794,7 +2799,7 @@ SUBROUTINE testIterativeSolve_GMRES()
 
   !set iterations and convergence information and build/set M
   SELECTTYPE(thisLS); TYPE IS(LinearSolverType_Iterative)
-    CALL thisLS%setConv(2_SIK,1.0E-9_SRK,1.0E-9_SRK,1000_SIK,30_SIK)
+    CALL thisLS%setConv(2_SIK,1.0E-6_SRK,1.0E-6_SRK,100_SIK,3_SIK)
```
djabaay: Don't need _SIK's, even though they were already there.

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -2874,7 +2885,7 @@ SUBROUTINE testIterativeSolve_GMRES()
...
```
djabaay: No _SIKs

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -2891,6 +2902,8 @@ SUBROUTINE testIterativeSolve_GMRES()
   thisB(9)=0.6875_SRK
   thisB=thisB*10000._SRK
   match=.TRUE.
+  SELECTTYPE(thisLS); TYPE IS(LinearSolverType_Iterative)
+  ENDSELECT
```
djabaay: Does this end select need to be below the following loop?

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -3021,6 +3020,96 @@ SUBROUTINE testIterativeSolve_GMRES()
...
+
+  !set b
+  SELECTTYPE(b => thisLS%b); TYPE IS(RealVectorType)
+    CALL b%set(1.0_SRK)
+  ENDSELECT
+
+  !set iterations and convergence information and build/set M
+  SELECTTYPE(thisLS); TYPE IS(LinearSolverType_Iterative)
+    CALL thisLS%setConv(2_SIK,1.0E-6_SRK,1.0E-6_SRK,100_SIK,3_SIK)
```
djabaay: delete _SIKs

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -3021,6 +3020,96 @@ SUBROUTINE testIterativeSolve_GMRES()
...
+  CALL pList%add('LinearSolverType->solverMethod',GMRES)
+  CALL pList%add('LinearSolverType->MPI_Comm_ID',PE_COMM_SELF)
+  CALL pList%add('LinearSolverType->numberOMP',1_SNK)
+  CALL pList%add('LinearSolverType->timerName','testTimer')
+  CALL pList%add('LinearSolverType->matType',BANDED)
+  CALL pList%add('LinearSolverType->A->MatrixType->n',9_SNK)
+  CALL pList%add('LinearSolverType->A->MatrixType->m',9_SNK)
+  CALL pList%add('LinearSolverType->A->MatrixType->nnz',33_SNK)
+  CALL pList%add('LinearSolverType->x->VectorType->n',9_SNK)
+  CALL pList%add('LinearSolverType->b->VectorType->n',9_SNK)
```
djabaay: Is the _SNK necessary?

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -3098,7 +3191,7 @@ SUBROUTINE testIterativeSolve_GMRES()
 
   !set iterations and convergence information and build/set M
   SELECTTYPE(thisLS); TYPE IS(LinearSolverType_Iterative)
-    CALL thisLS%setConv(2_SIK,1.0E-9_SRK,1.0E-9_SRK,1000_SIK,30_SIK)
+    CALL thisLS%setConv(2_SIK,1.0E-6_SRK,1.0E-6_SRK,1000_SIK,3_SIK)
```
djabaay: delete _SIKs

unit_tests/testLinearSolver/testLinearSolver.f90
```diff
@@ -3178,7 +3275,7 @@ SUBROUTINE testIterativeSolve_GMRES()
...
```
djabaay: delete _SIKs

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,206 @@
...
+  CALL pList%add('LinearSolverType->solverMethod',GMRES)
+  CALL pList%add('LinearSolverType->MPI_Comm_ID',PE_COMM_WORLD)
+  CALL pList%add('LinearSolverType->numberOMP',1_SNK)
+  CALL pList%add('LinearSolverType->timerName','testTimer')
+  CALL pList%add('LinearSolverType->matType',DISTRIBUTED_BANDED)
+  CALL pList%add('LinearSolverType->A->MatrixType->n',9_SNK)
+  CALL pList%add('LinearSolverType->A->MatrixType->m',9_SNK)
+  CALL pList%add('LinearSolverType->A->MatrixType->nnz',33_SNK)
+  CALL pList%add('LinearSolverType->x->VectorType->n',9_SNK)
+  CALL pList%add('LinearSolverType->b->VectorType->n',9_SNK)
```
djabaay: Are the _SNKs needed?

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,206 @@
...
+  ENDSELECT
+
+  SELECTTYPE(b => thisLS%b); TYPE IS(NativeDistributedVectorType)
+    b%b = 1.0_SRK
+  ENDSELECT
+
+  !set iterations and convergence information and build/set M
+  SELECTTYPE(thisLS); TYPE IS(LinearSolverType_Iterative)
+    thisLS%hasX0 = .TRUE.
+    CALL thisLS%setConv(2_SIK,1.0E-6_SRK,1.0E-6_SRK,20_SIK,4_SIK)
```
djabaay: delete _SIKs

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,206 @@
...
+  ENDSELECT
+
+  ! Use default initial guess (all ones)
+  SELECTTYPE(x => thisLS%x); TYPE IS(NativeDistributedVectorType)
+    x%b = 1.0_SRK
+  ENDSELECT
+
+  SELECTTYPE(b => thisLS%b); TYPE IS(NativeDistributedVectorType)
+    b%b = 1.0_SRK
+  ENDSELECT
```
djabaay: Assertions for if the LS is initialized and A is setup properly?

cmake/Futility_Configurations.cmake
```diff
@@ -191,7 +191,7 @@ IF(CMAKE_Fortran_COMPILER_ID STREQUAL "Intel" OR
 
     SET(Fortran_FLAGS_RELEASE
         ${CSYM}O3
-        ${CSYM}Os
+        ${CSYM}march=native
```
aarograh: Not familiar with these flags... what did the old one do, and what does the new one do?

src/EigenvalueSolverTypes.f90
```diff
@@ -78,7 +78,10 @@ MODULE EigenvalueSolverTypes
 #endif
 
 !> set enumeration scheme for TPLs
-INTEGER(SIK),PARAMETER,PUBLIC :: SLEPC=0,ANASAZI=1,NATIVE=4
+INTEGER(SIK),PARAMETER,PUBLIC :: SLEPC=0,ANASAZI=1,EV_NATIVE=4
+!> Temporary parameter to be retained until corresponding MPACT merge:
+INTEGER(SIK),PARAMETER,PUBLIC :: NATIVE=4
```
aarograh: Fine to leave this for now, but making a comment here so I can generate an issue to come back and address this

src/EigenvalueSolverTypes.f90
```diff
@@ -78,7 +78,10 @@ MODULE EigenvalueSolverTypes
 #endif
 
 !> set enumeration scheme for TPLs
-INTEGER(SIK),PARAMETER,PUBLIC :: SLEPC=0,ANASAZI=1,NATIVE=4
+INTEGER(SIK),PARAMETER,PUBLIC :: SLEPC=0,ANASAZI=1,EV_NATIVE=4
```
aarograh: So we should really give all of these the same prefix.  That's closer to our standard practice.  So `EV_SLEPC`, `EV_ANASAZI`.  May need a full set of temporary parameters to go along with those.

src/LinearSolverTypes.f90
```diff
@@ -89,7 +89,10 @@ MODULE LinearSolverTypes
 PUBLIC :: LinearSolverType_Clear_ValidParams
 
 !> set enumeration scheme for TPLs
-INTEGER(SIK),PARAMETER,PUBLIC :: PETSC=0,TRILINOS=1,PARDISO_MKL=2,MKL=3,NATIVE=4
+INTEGER(SIK),PARAMETER,PUBLIC :: PETSC=0,TRILINOS=1,PARDISO_MKL=2,MKL=3,LS_NATIVE=4
```
aarograh: Same here, use the same prefix for all of these

src/LinearSolverTypes.f90
```diff
@@ -1431,17 +1427,53 @@ SUBROUTINE getResidual_LinearSolverType_Iterative(solver,resid)
...
+        SELECTTYPE(resid)
+        TYPE IS(NativeDistributedVectorType)
+          resid%b=-b%b
+          CALL BLAS_matvec(THISMATRIX=solver%A,X=solver%X,Y=resid,ALPHA=1.0_SRK,BETA=1.0_SRK)
+        CLASS DEFAULT
+          CALL eLinearSolverType%raiseError('Incorrect call to '// &
+              modName//'::getResidual'//' - Type mismatch between getResidual argument and solver datatypes')
+        ENDSELECT
+#endif
       ENDSELECT
```
aarograh: Should have a `CLASS DEFAULT` statement that throws an error

src/LinearSolverTypes.f90
```diff
@@ -1431,17 +1427,53 @@ SUBROUTINE getResidual_LinearSolverType_Iterative(solver,resid)
...
```
aarograh: `CLASS DEFAULT` again

src/LinearSolverTypes.f90
```diff
@@ -1597,228 +1629,237 @@ SUBROUTINE solveBiCGSTAB(solver)
...
+
 !
 !-------------------------------------------------------------------------------
-!> @brief Solves the Iterative Linear System using the GMRES method with no
-!>        preconditioning
+!> @brief Control program for restarted GMRES solver. Handles restarts,
+!> parallelism, and a few other things. Solver body is located in
+!> solveGMRES_partial
 !> @param solver The linear solver to act on
+!> @param PreCondType The preconditioner object to use on the system
```
aarograh: These param names do not match the actual dummy arguments

src/LinearSolverTypes.f90
```diff
@@ -1597,228 +1629,237 @@ SUBROUTINE solveBiCGSTAB(solver)
...
-  solver%info=0
-  CALL u%clear()
-ENDSUBROUTINE solveGMRES_nopc
 !
 !-------------------------------------------------------------------------------
-!> @brief Solves the Iterative Linear System using the GMRES method with
-!>        left-preconditioning
+!> @brief GMRES core solver routines.
 !> @param solver The linear solver to act on
 !> @param PreCondType The preconditioner object to use on the system
```
aarograh: mismatched param names, as well as 2 missing params

unit_tests/testLinearSolver/CMakeLists.txt
```diff
@@ -8,3 +8,4 @@
 #++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
 INCLUDE(Futility_CreateUnitTest)
 Futility_CreateUnitTest(testLinearSolver)
+Futility_CreateParUnitTest(testLinearSolverParallel 2)
```
aarograh: This needs to be inside a `IF(${PACKAGE_NAME}_ENABLE_MPI)` or similar statement

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,206 @@
...
+  !    -1     4    -1     0    -1     0     0     0     0
+  !     0    -1     4     0     0    -1     0     0     0
+  !    -1     0     0     4    -1     0    -1     0     0
+  !     0    -1     0    -1     4    -1     0    -1     0
+  !     0     0    -1     0    -1     4     0     0    -1
+  !     0     0     0    -1     0     0     4    -1     0
+  !     0     0     0     0    -1     0    -1     4    -1
+  !     0     0     0     0     0    -1     0    -1     4
+  SELECTTYPE(A => thisLS%A); TYPE IS(DistributedBandedMatrixType)
+      CALL A%set(1,1, 4.0_SRK)
```
aarograh: indentation in this block only needs to be 2 spaces, not 4 (check for tabs?)

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,206 @@
...
+    thisLS%hasX0 = .TRUE.
+    CALL thisLS%setConv(2_SIK,1.0E-6_SRK,1.0E-6_SRK,20_SIK,4_SIK)
+  ENDSELECT
+
+
+  !solve it
+  CALL thisLS%solve()
+
+  !Store expected solution (from MATLAB) in B
+  IF (mpiTestEnv%rank == 0) THEN
```
aarograh: No space after `IF`

cmake/Futility_Configurations.cmake
stimpsonsg: additionally, these are changes to Intel compiler flags...was Intel actually tested?

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,206 @@
...
+    thisB(4)=0.875_SRK
+    thisB(5)=1.125_SRK
+    thisB=10000.0_SRK*thisB
+    match=.TRUE.
+    DO i=1,SIZE(thisB)
+      SELECTTYPE(X => thisLS%X); TYPE IS(NativeDistributedVectorType)
+        IF(ALLOCATED(dummyvec)) DEALLOCATE(dummyvec)
+        ALLOCATE(dummyvec(SIZE(X%b)))
+        dummyvec = X%b
+        IF(NINT(thisB(i)) /= NINT(10000.0_SRK*dummyvec(i))) THEN
```
HendersonSC: This may be obvious to someone, not me, why multiply both test and reference values by 10,000 and `nint`? If it is just about tolerance why not `softeq`?

cmake/Futility_Configurations.cmake
mattkab2: The intel flag should have been dropped; I'll correct this
-march=native enables cpu-specific instructions; such as avx512. These are not enabled by default, even when using O3. When enabling this option, I saw runtimes for 5a-2d drop from 10 minutes to 7.5 for standard PETSc runs.

src/LinearSolverTypes.f90
mattkab2: I sure did. Changes must have gotten clobbered... excellent.

unit_tests/testLinearSolver/testLinearSolver.f90
mattkab2: These were copies from existing tests. Removing them changed nothing so I'll commit an updated unit test file with them all removed (seeing as they're pointless)

mattkab2: These were copies from existing tests. Removing them changed nothing so I'll commit an updated unit test file with them all removed (seeing as they're pointless)

src/LinearSolverTypes.f90
mattkab2: This would prevent one from running a solver in serial when using a parallel build; which seems like an unnecessary restriction to me. If you believe it's the better course of action, we can pursue it though.

cmake/Futility_Configurations.cmake
aarograh: I've been reading a bit about this option.  I THINK it is safe to use, and that speedup is certainly attractive.  However, I think it would be good to do some more extensive testing before we merge this in, such as:
1. Run heavies with and without the flag to ensure there are no performance degradations across a larger set of problems
2. Ensure this isn't going to cause any issues for clusters.  Reading about the flag a bit just now, I found the following:
image
My concern in that case is what if someone builds on a head node of a cluster and then runs on the compute nodes?  My expectation is that the CPU between the two will generally be the same or very similar and it's probably fine, but I think it'd be good to test some builds on a couple different machines before putting this on master.

aarograh: An additional thought: we've talked a bit about supporting more compilers, so IF we add this as a default for gfortran, let's leave it enabled for Intel as well.

jpjones6: I had done a good bit of testing using machine specific optimizations with MPACT. For a single assembly ge14 case with TH turned on I saw ~2% speedup (it was consistent, i.e., not likely to be just noise). However, many many tests failed when doing a full ctest. The reason was due to rounding differences making some .APPROXEQA. to fail. I saw that setting -mfpmath=sse also helps slightly.

aarograh: @mattkab2 Need to do a full MPACT build using this Futility branch to see if there are any failures introduced by this compiler flag change.

mattkab2: It seems my comments were lost or not posting for some reason. Here's a summary:
These command line changes have a strong chance of breaking things when the processors on login nodes are newer than the compute nodes, even by a single model. @djabaay and I actually had a conversation about this some time ago and one suggestion that came up was to perhaps add another build type "optimized" for developers or cluster admins that creates a non-portable executable.
@aarograph Yes, I already have one actually, and I've identified and fixed the floating point errors that occur when enabling march=native. I've been iterating between that branch and these changes to try and minimize the disruption to code. Since some changes on the MPACT side are required to integrate GMRES completely anyways, one option I'm leaning towards is to merge this code in without the command line changes, update on the MPACT side, then add this back in afterwards.

unit_tests/testLinearSolver/testLinearSolverParallel.f90
mattkab2: It should absolutely be softeq. Some of these edits are quite old, when I was first learning the code base. In this case, I just copied existing tests but I will go back and update them all to use the new (since the original files were written anyways) methods. This NINT method is a bit archaic.

cmake/Futility_Configurations.cmake
aarograh: @mattkab2 @djabaay There are a lot of potentially issues with changing release compiler flags.  If we end up doing this, it needs to be done completely separately from any other changes so that it can be thoroughly vetted and easily reverted if necessary.

Back out this change and any downstream MPACT changes you've made that were due to this compiler flag change.  If you'd like to keep them, move them to a separate branch, but don't merge them in with any of this GMRES work.

djabaay: Sounds good @aarograh

src/LinearSolverTypes.f90
mattkab2: This code's been refactored; getresidual has been moved to 2 routines, one for the residual norm and one for the residual vector (and iteration count).
This makes it a more common and reliable interface overall in the LinearSolverTypes framework which alleviates some headaches when integrating with MPACT.

unit_tests/testLinearSolver/testLinearSolver.f90
mattkab2: SIK's and SNK's have been purged from the linear solver test files

src/LinearSolverTypes.f90
mattkab2: This routine's been updated. The getresidual method's been changed around to use the more general BLAS_matvec routines.
It also gets rid of a lot of headaches when integrating with CMFD; basically there's an interface now which controls whether you get iteration count and residual norm or just the residual vector. It's also implemented for the direct solvers so you don't have to select type iterative all the time.

src/LinearSolverTypes.f90
mattkab2: Got it. I'll add this in the next commit. If you have ideas for other naming schemes I'm open to them. I've confirmed that leaving the NATIVE temporary causes no problems, so leaving these names as temporaries shouldn't cause any problems either.

src/LinearSolverTypes.f90
```diff
@@ -755,6 +775,54 @@ SUBROUTINE updatedA(solver)
   solver%isDecomposed=.FALSE.
 
 ENDSUBROUTINE updatedA
+
+SUBROUTINE getResidualVec_LinearSolverType_Direct(solver,resid,nIters)
```
nfherrin: No subroutine description or brief?

nfherrin: No subroutine description or brief?

src/LinearSolverTypes.f90
```diff
@@ -755,6 +775,54 @@ SUBROUTINE updatedA(solver)
...
+  ELSE
+    CALL eLinearSolverType%raiseError('Incorrect call to '// &
+       modName//'::'//myName//' - Residual or solver not initialized.')
+  ENDIF
+
+  ! Direct method; nIters has no meaning
+  IF (PRESENT(nIters)) nIters = 0
+ENDSUBROUTINE getResidualVec_LinearSolverType_Direct
+
+SUBROUTINE getResidualNorm_LinearSolverType_Direct(solver,resid,nIters)
```
nfherrin: Same comment as above. Need a subroutine brief and variable descriptions.

src/LinearSolverTypes.f90
```diff
@@ -755,6 +775,54 @@ SUBROUTINE updatedA(solver)
...
+  IF (resid%isInit .AND. solver%isInit) THEN
+    CALL BLAS_copy(solver%b,resid)
+    CALL BLAS_matvec(thisMatrix=solver%A,alpha=-1.0_SRK,x=solver%x,beta=1.0_SRK,y=resid)
+  ELSE
+    CALL eLinearSolverType%raiseError('Incorrect call to '// &
+       modName//'::'//myName//' - Residual or solver not initialized.')
+  ENDIF
+
+  ! Direct method; nIters has no meaning
+  IF (PRESENT(nIters)) nIters = 0
```
nfherrin: Should direct methods return 0 iterations or a known negative value to indicate the method was direct?

src/LinearSolverTypes.f90
```diff
@@ -1420,43 +1494,55 @@ SUBROUTINE setConv_LinearSolverType_Iterative(solver,normType_in,  &
...
   ENDIF
-ENDSUBROUTINE getResidual_LinearSolverType_Iterative
-!
-!-------------------------------------------------------------------------------
-!>
-SUBROUTINE getIterResidual_LinearSolverType_Iterative(solver,niters,resid)
+
+ENDSUBROUTINE getResidualVec_LinearSolverType_Iterative
+
+SUBROUTINE getResidualNorm_LinearSolverType_Iterative(solver,resid,nIters)
```
nfherrin: No subroutine description

src/LinearSolverTypes.f90
```diff
@@ -1597,228 +1686,237 @@ SUBROUTINE solveBiCGSTAB(solver)
...
+    IF (V(i)%isInit) CALL V(i)%clear()
+  ENDDO
+  DEALLOCATE(V)
+  CALL Vy%clear()
+  DEALLOCATE(Vy)
+  DEALLOCATE(orthogReq)
+
+ENDSUBROUTINE solveGMRES_partial
+
+SUBROUTINE arnoldi(thisLS,V,k,h,orthogReq)
```
nfherrin: Needs description

src/VectorTypes.f90
```diff
@@ -293,11 +293,18 @@ SUBROUTINE VectorFactory(vector, params)
...
+!> NOTE: When we can compile with the 2003/08 standard, it would be best to
+!>       write an interface and distinguish between pointer/allocatable
+!>       attributes
+SUBROUTINE VectorResemble(dest, source, p)
   CHARACTER(LEN=*),PARAMETER :: myName="VectorResemble"
-  CLASS(VectorType),POINTER,INTENT(INOUT) :: dest
+  CLASS(VectorType),INTENT(INOUT),ALLOCATABLE :: dest
   CLASS(VectorType),POINTER,INTENT(IN) :: source
-  CLASS(ParamType),INTENT(INOUT) :: params
+  CLASS(ParamType),INTENT(INOUT),OPTIONAL :: p
```
nfherrin: Changed parameter name needs to be reflected in the subroutine brief.

aarograh:
@mattkab2 Any plans to address review comments on this?  The pull request is getting a bit stale.

mattkab2:
@aarograh There have been, but things keep piling up. Since I'm working on a writeup for this, it's a little higher priority than usual and I've been bringing it up to recent master.
Many of these comments have been addressed in past commits, so once it is no longer stale I will ping the commenters to get their feedback on the changes.

mattkab2:
@djabaay @nfherrin @aarograh @HendersonSC If you guys could go over your comments, I believe all of them should be addressed. Let me know if you have further questions.

src/LinearSolverTypes.f90
mattkab2: It's an open question; I figured 0 would reflect the direct nature, as I figured a negative value is typically used as a flag that an error occurred. @djabaay do you have any thoughts?

src/LinearSolverTypes.f90
mattkab2: Oops, thought I caught this one on the last commit. I'll add the brief.

src/LinearSolverTypes.f90
nfherrin: My thought is, the default value is zero, so if an error happens that doesn't change the default value (like the method itself is somehow accidentally skipped), then you don't necessarily have an indication of an error. But if direct methods always returned like a -151 (or whatever) when they finished, then you can be more certain an error didn't occur in the solve.

cmake/Futility_Configurations.cmake
```diff
@@ -190,7 +190,6 @@ IF(CMAKE_Fortran_COMPILER_ID STREQUAL "Intel" OR
     ENDIF()
 
     SET(Fortran_FLAGS_RELEASE
-        ${CSYM}O3
```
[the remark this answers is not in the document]
aarograh: For some reason this didn't post severeal weeks ago when it yped it... but anyways, don't change any compiler flags in this merge

src/EigenvalueSolverTypes.f90
aarograh: Still need to update these.  Just leave the old names there too since I miagine they're used in MPACT.  We can clear out the old names in favor of `EV_*` once MPACT is using thmem.

unit_tests/testLinearSolver/testLinearSolverParallel.f90
mattkab2: Since the parallel solver is the same object as the serial one (just with parallel datatypes) I figured that duplicating the initialization tests (and setup for parallel datatypes) would be overdoing it.
I can add them, but anything wrong with the initialization should be caught in the serial test.

src/EigenvalueSolverTypes.f90
mattkab2: The only one used is NATIVE, actually, which is why it was the only one I left alone. I tested MPACT with these changes and it worked fine. I'll follow your recommendations, but it's not necessary for the eigenvalue TPLs.

cmake/Futility_Configurations.cmake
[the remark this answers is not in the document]
mattkab2: This shouldn't have stayed in the MR. Thanks for catching it.

src/LinearSolverTypes.f90
aarograh: My preference would be `LINSYS_*` for each of them.  `LS_*` would be ok too, but I tend to prefer slightly longer, more descriptive names.  Just so long as they're all the same

mattkab2:
With the latest commits the native CMFD now runs faster than PETSc for my test problem (5a-2d) by 15%, with a larger speedup expected in 3D problems. There's also a memory reduction (100-200 MB).

mattkab2:
@HendersonSC @djabaay Could you guys take a look at your comments and see if they've been addressed?

stimpsonsg:
@mattkab2 did you do a scaling study on 5a-2D?  ...if not, it would be worthwhile.

How many cores did you see a 15% speedup on (16?)?  

Be sure to keep the ticket updated with any results you have:
https://vminfo.casl.gov/trac/casl_phi_kanban/ticket/5855

mattkab2:
I did a scaling study but only for 1 node; I had some difficulties with the new SLURM system for multi-node jobs.
The 15% speedup remained constant from 8-24 processors.

Thanks for reminding me, I'll make sure to update it.

aarograh:
@mattkab2 In your scaling study, were you using the `-march=native` flag to compile?  If so, did you perform the study without that flag as well?

stimpsonsg:
Also, was that with the same number of outer iterations?  You mentioned that the preconditioning isn't there yet, so the number of iterations could vary a bit in theory.  RBSOR, for example, is faster running but slower to converge, but still observed a speedup even with additional iterations...just worth noting to really understand if the gains are from the parallel efficiency or something else about the solver.

mattkab2:
@aarograh The march=native flag was used on the build. The same build was used to generate the native solver results and petsc results. Without the flag, the native solver is not competitive because it has been written to take advantage of fused-multiply-add and vector instructions.

@stimpsonsg The number of outers is the same. Preconditioning does not change the number of outer iterations because all that is changing in this code is the method of solving the CMFD matrix, which is not altered in any way. This is part of the motivation for this work as it can be used with any variant of CMFD, even multilevel schemes as a drop-in replacement.

stimpsonsg:
Okay, that's good that they have consistent outers.  If you lack preconditioning and hit the max inner iteration limits, it could run into a situation where you need more outers.

Do you have actual numbers without --march=native?

mattkab2:
@stimpsonsg I only did it for 8 processors and discarded the results, but I can regenerate them fairly easily. I'll include it when I update the ticket.

aarograh:
@mattkab2 so my concern with using that flag is this: the comparisons to petsc are not fair if petsc was not also compiled using that flag.  Since we typically link against petsc as a TPL, I'm guessing that's the case.  Did you build your own version of petsc using `-march=native`?  Do you know if they use similar compiler flags to improve optimizations?  I'm guessing they don't since clusters frequently can have small variations in hardware.

mattkab2:
@aarograh You are right, I suppose this is not a fair comparison then. I recall building a separate version of PETSc for the much older tests on Flux (old cluster), but I forgot that step for the new tests on Lighthouse (new cluster).

Thanks for pointing that out. I'll have to go build it and see what results we see.

mattkab2:
@aarograh I do want to be clear though that there is reason to believe that the native solver will still be faster. It makes extensive use of BLAS/LAPACK, which, like PETSc, would also not have received the march=native flags. I'm building these libraries now, and have results by early tomorrow.

mattkab2:
@aarograh @HendersonSC Getting the build process to link to the march=native compilations of PETSC/LAPACK rather than the cluster defaults was trickier than expected, but the overall results are very similar.
The speedup is not quite as large, but still ranges from 8-12%. Memory savings are not affected, and are still about 150 MB per core in the 8 processor case.

The reason for the decline in speedup, however, is not due to PETSc getting faster but the native solver becoming slower. I do not know the reason for this but I am looking into it.

mattkab2:
Below is a table of the results for the 5a-2D test case on the Lighthouse cluster at Michigan. Speedup is expected to be better for 3D cases, and for architectures Skylake or newer (Lighthouse is Haswell).

speedupFactor

The "unoptimized" here means no machine-specific optimizations (march=native). The top line is the standard MPACT build we have on Lighthouse, which shows an improvement even without these enabled. 
The two lines that follow are MPACT builds compiled with PETSc and linked to one of two linear algebra packages (OpenBLAS or LAPACK). In these cases, all packages were compiled with machine-specific optimizations. OpenBLAS is a good candidate for pushing results even further.

src/LinearSolverTypes.f90
aarograh: Personally I think 0 is fine here

unit_tests/testLinearSolver/testLinearSolverParallel.f90
aarograh: did this get updated?

src/LinearSolverTypes.f90
aarograh: I think this still needs to be addressed

aarograh:
@mattkab2 Thanks for posting the results.  make sure they find their way to the casl ticket too

@HendersonSC @nfherrin If you guys could resolve anything that has been addressed, please do so.  If it still needs to be addressed, comment to let Matt know.  I'd like to get this pull request finished up before it gets stale again.

unit_tests/testLinearSolver/testLinearSolverParallel.f90
mattkab2: Yes, the unit test now uses ALL's and APPROXEQ's so these checks are much cleaner.

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,206 @@
...
+          match=.FALSE.
+          EXIT
+        ENDIF
+      ENDSELECT
+    ENDDO
+  ELSE
```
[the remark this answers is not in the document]
HendersonSC:
```suggestion
  IF(rank ==0) THEN
    thisB = (/0.6875_SRK,0.875_SRK,0.6875_SRK,0.875_SRK,1.125_SRK/)
  ELSE
    thisB = (/0.875_SRK,0.6875_SRK,0.875_SRK,0.6875_SRK/)
  ENDIF
  CALL thisLS%X%get(dummyvec)
  ASSERT(ALL(SOFTEQ(thisB(:),dummyvec(:),1.0E-05_SRK)), 'CALL Iterative%solve() -GMRES FAILED!')
```
This is of course assuming that thisLS%get doesn't require a pre-allocated array

unit_tests/testLinearSolver/testLinearSolverParallel.f90
```diff
@@ -0,0 +1,177 @@
...
+    thisB(:) = (/0.6875_SRK,0.875_SRK,0.6875_SRK,0.875_SRK,1.125_SRK/)
+    CALL thisLS%X%get(dummyvec)
+    ASSERT(ALL(SOFTEQ(thisB(:),dummyvec(:),1.0E-05_SRK)), 'CALL Iterative%solve() -GMRES FAILED!')
+  ELSE
+    ALLOCATE(dummyvec(4))
+    ALLOCATE(thisB(4))
+    thisB(:)=(/0.875_SRK,0.6875_SRK,0.875_SRK,0.6875_SRK/)
+    CALL thisLS%X%get(dummyvec)
+    ASSERT(ALL(SOFTEQ(thisB(:),dummyvec(:),1.0E-05_SRK)), 'CALL Iterative%solve() -GMRES FAILED!')
+  ENDIF
```
HendersonSC, line 169:
```suggestion
  IF(rank == 0) THEN
    thisB(:) = (/0.6875_SRK,0.875_SRK,0.6875_SRK,0.875_SRK,1.125_SRK/)
  ELSE
    thisB(:)=(/0.875_SRK,0.6875_SRK,0.875_SRK,0.6875_SRK/)
  ENDIF
  CALL thisLS%X%get(dummyvec)
  ASSERT(ALL(SOFTEQ(thisB(:),dummyvec(:),1.0E-05_SRK)), 'CALL Iterative%solve() -GMRES FAILED!')
```

HendersonSC: This is assuming that thisLS%get doesn't require a pre-allocated array

mattkab2: Actually the %get function is a member of the VectorTypes class, which is specified to work only on pre-allocated arrays. The interface %get handles accessing a single element, range of elements, or a subset of elements, so behavior in the case of an unallocated input is not well-defined.

src/LinearSolverTypes.f90
mattkab2: Names have been updated

unit_tests/testLinearSolver/testLinearSolverParallel.f90
HendersonSC: So first things first, this is a super trivial request in the grand scheme of things. The bigger thing,
```
SUBROUTINE getAll_PETScVectorType(thisVector,getval,ierr)
  CLASS(PETScVectorType),INTENT(INOUT) :: thisVector     
  REAL(SRK),INTENT(OUT) :: getval(:)                     
```
I am pretty sure that using the allocation status/size of an `intent(out)` is undefined. At a minimum it is probably bad form, I invite others to comment/correct as necessary.

aarograh: You can only check allocation status on an `ALLOCATABLE` dummy argument.  That dummy argument should probably be allocatable, but that's very much outside the scope of this PR.

I like the spirit of your suggestion but I'm not sure how much of it we can use.  You good just resolving this as is?

src/LinearSolverTypes.f90
```diff
@@ -1578,22 +1577,20 @@ SUBROUTINE getResidualNorm_LinearSolverType_Iterative(solver,resid,nIters)
   REQUIRE(ASSOCIATED(solver%x))
   REQUIRE(solver%x%isInit)
 
-  IF(solver%TPLType == PETSC) THEN
+  !IF(solver%TPLType == PETSC) THEN
 #ifdef FUTILITY_HAVE_PETSC
-    IF (PRESENT(nIters)) CALL KSPGetIterationNumber(solver%ksp,niters,ierr)
-    CALL KSPGetResidualNorm(solver%ksp,resid,ierr)
+    !IF (PRESENT(nIters)) CALL KSPGetIterationNumber(solver%ksp,niters,ierr)
+    !CALL KSPGetResidualNorm(solver%ksp,resid,ierr)
```
aarograh, line 1583: What's the story with this change?

unit_tests/testLinearSolver/testLinearSolverParallel.f90
HendersonSC: In light of the constraints with existing code, I updated the suggestion to something of a middle ground. It is essentially take it or leave it, just be aware of chances to streamline that sort of logic and make things easier for follow on development.

src/LinearSolverTypes.f90
mattkab2: It was a mistake. I'm preparing the merge request for the MPACT side of things and this leaked in from there. I'm fixing it and a few other minor mistakes at the moment.

src/LinearSolverTypes.f90
```diff
@@ -236,10 +264,12 @@ SUBROUTINE linearsolver_sub_absintfc(solver)
     PROCEDURE,PASS :: solve => solve_LinearSolverType_Iterative
     !> @copybrief LinearSolverTypes::getResidual_LinearSolverType_Iterative
     !> @copydetails LinearSolverTypes::getResidual_LinearSolverType_Iterative
-    PROCEDURE,PASS :: getResidual_LinearSolverType_Iterative
-    PROCEDURE,PASS :: getIterResidual_LinearSolverType_Iterative
-    GENERIC :: getResidual => getResidual_LinearSolverType_Iterative, &
```
stimpsonsg, line 241: why did this generic interface go away?

mattkab2: It didn't go away, exactly. It was generalized to the LinearSolverType_Base, which specifies an interface "getResidual" to be implemented by the child classes (LinearSolverType_Direct and iterative).
A residual is still computable for direct solvers, and this generalization cleans up the CMFD code in MPACT significantly, especially in the getConvNorm subroutines.

stimpsonsg: yeah, ok....moving to the base type is fair

aarograh: so which interfaces(s) had the issue with switching the order of the arguments?

mattkab2: It is this one, the getResidual.

I'm working on a patch for this (it was delayed by the bwr thermal expansion branch). The goal was to make these subroutines have all optional arguments so that the calls in MPACT could be kept, and things would not break in the transition.
Unfortunately that won't work out, since making both of the subroutines in the interface have optional arguments makes this ambiguous.
So the solution is to replace the interface with a single function with 3 optional arguments (iteration count, residual norm, and residual vector). These can be ordered to conform to the function calls to MPACT (I believe) and the Futility patch conventions can be switched over to these. This should not take much longer, hopefully there will be some changes pushed up today/tomorrow.

aarograh: Another option would be to have a couple versions of the routine behind a GENERIC interface.  One version of the generic would match the old version that MPACT calls.  Only one version would have to actually do something: the rest could simply wrap the main one, passing in temporary variables just to match the interface.  Sometimes the GENERIC interface is cleaner than having a bunch of optional arguments.

mattkab2: That might work too, but it's been fixed now.
However, another incompatibility has popped up with the changes to VectorResemble. The old code passed in a pointer, but it's been modified now to accept an allocatable.
Unfortunately, in Fortran 95 a distinction can't be made in the function signature between an allocatable and a pointer (in '03 or '08 you can) so this is a bit trickier to deal with but I think it can be done with a (relatively) simple patch in MPACT thanks to the native distributed vector type which is already in.

aarograh: I'm really nervous about how many interface changes are going on in this pull request.  Is this allocatable/pointer change really necessary?  It doesn't seem like it to me from reading the code.

jpjones6: I can't speak to how necessary it is for this MR, however, the addition of resemble being able to handle allocatable vectors would be helpful.

mattkab2: There should only be 2; getResidual and this one.
I adjusted this one for 2 reasons: It is called only once in all of the MPACT code (CMFD_Base) and it allows me to use the VectorResemble routine to allocate non-pointer vector types. Which is important, as making everything in the linear solver a pointer or target is just not practical.

aarograh: Maybe there needs to just be `VectorResemble` and `VectorResemble_Alloc`.  Having both pointer and allocatable versions could be useful, especially since you can't apply the TARGET attribute to an allocatable array.

aarograh: In all fairness, it probably should have started as an allocatable anyways... some of the early mpact/futility code is pretty pointer-happy.  I'm just not a fan of simultaneous merges in multiple repos

mattkab2: The different name should work. I'm not sure why I didn't think of that; it's a lot less work than the fix I'm trying to make.

mattkab2: An update on this; the newest build causes no incompatibilities, and all unit tests in MPACT pass. I'm doing some more in-depth tests (Heavies, unused variables, serial versions) but it's looking good.

unit_tests/testInterpolators/testInterpolators.f90
```diff
@@ -229,21 +229,21 @@ SUBROUTINE test3DInterp()
...
+  ASSERT_SOFTEQ(Interpolant,46.0_SRK,2e-14_SRK,'Incorrect Interpolant')
   Interpolant=Interp(label1,label2,label3,table_3D,[1.6_SRK,3.5_SRK,1.0_SRK])
-  ASSERT_SOFTEQ(Interpolant,106.0_SRK,1e-14_SRK,'Incorrect Interpolant')
+  ASSERT_SOFTEQ(Interpolant,106.0_SRK,2e-14_SRK,'Incorrect Interpolant')
   Interpolant=Interp(label1,label2,label3,table_3D,[1.6_SRK,2.4_SRK,0.5_SRK])
-  ASSERT_SOFTEQ(Interpolant,85.6_SRK,1e-14_SRK,'Incorrect Interpolant')
+  ASSERT_SOFTEQ(Interpolant,85.6_SRK,2e-14_SRK,'Incorrect Interpolant')
   Interpolant=Interp(label1,label2,label3,table_3D,[1.6_SRK,2.4_SRK,3.5_SRK])
-  ASSERT_SOFTEQ(Interpolant,85.6_SRK,1e-14_SRK,'Incorrect Interpolant')
+  ASSERT_SOFTEQ(Interpolant,85.6_SRK,2e-14_SRK,'Incorrect Interpolant')
```
aarograh, line 246: It's not a big deal, but do you know why these needed to change?

mattkab2: I didn't realize this got committed, I had changed it and made a mental note to investigate.

It happened when I made a modification to the getResidual functions. However, checking again the increase in tolerance seems to be unnecessary, so this file is being reverted.

aarograh:
@mattkab2 I think everything here seems reasonable.  Tests are passing in MPACT with your most recent Futility head?

mattkab2:
@aarograph Yes, all heavies on MPACT which I can run are passing with these changes. Same with unused variables/serial debug.

PR #234, opened 2020-03-03
Move PolygonType centroid and area calculations into their own routines

aarograh:
This is useful for when the PolygonType is modified by an external
code and these values should be updated without completely clearing
and reinitializing the object.


aarograh:
The specific use-case that necessitates this change is rotation of GenPinMeshType objects in MPACT for BWR calculations.  The vertices and edges were being rotated, but there was no way to update the centroid, which is needed for the CTF coupling.

aarograh:
All I did was pull logic out of `set` into 2 new routines.  Those routines are both called from `set`, which is already unit tested.  Nothing about the functionality actually changed here in Futility.

PR #235, opened 2020-03-03
Interpolators

jpjones6:
This MR contains the addition of interpolation routines to Futility as a new module. The interpolators do linear interpolation and flat extension of values if a value out of range is requested (i.e., no extrapolation is done). 1D, 2D, and 3D are included. Monotonically increasing and monotonically decreasing axis labels are allowed.


src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+!                             All rights reserved.                             !
+!                                                                              !
+! Futility is a jointly-maintained, open-source project between the University !
+! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
+! can be found in LICENSE.txt in the head directory of this repository.        !
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module implementing interpolation routines for input
+!>        arrays returning either interpolated values or a flat "extrapolation"
+!>        when interpolation is out of range.
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
```
stimpsonsg: maybe add some detail here clarifying that it's limited to 3D or smaller

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+  !> @copybrief Interpolators::Interp_1D
+  !> @copydetails Interpolators::Interp_1D
+  MODULE PROCEDURE Interp_1D
+  !> @copybrief Interpolators::Interp_2D
+  !> @copydetails Interpolators::Interp_2D
+  MODULE PROCEDURE Interp_2D
+  !> @copybrief Interpolators::Interp_3D
+  !> @copydetails Interpolators::Interp_3D
+  MODULE PROCEDURE Interp_3D
+ENDINTERFACE Interp
```
stimpsonsg: thinking a generic interface could be good

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+!===============================================================================
+CONTAINS
+!
+!-------------------------------------------------------------------------------
+!> @brief This routine takes a 1D data "table" and linearly interpolates to find
+!>        a desired point.input arrays and interpolates as needed. It is important
+!>        to note that if an out of range extrapolation is required a flat extension
+!>        of the table data is performed.
+!>
+FUNCTION Interp_1D(labels,table,pp) RESULT(p)
```
stimpsonsg: need doxy for input variables in all three routines

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+  !> @copydetails Interpolators::Interp_3D
+  MODULE PROCEDURE Interp_3D
+ENDINTERFACE Interp
+!
+!===============================================================================
+CONTAINS
+!
+!-------------------------------------------------------------------------------
+!> @brief This routine takes a 1D data "table" and linearly interpolates to find
+!>        a desired point.input arrays and interpolates as needed. It is important
```
aarograh: got some typo action going on here and in the other 2 `Interp_*` routine headers

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+!>        to note that if an out of range extrapolation is required a flat extension
+!>        of the table data is performed.
+!>
+FUNCTION Interp_1D(labels,table,pp) RESULT(p)
+  REAL(SRK),INTENT(IN) :: labels(:),table(:),pp
+  REAL(SRK) :: p,f(2)
```
aarograh: All dummy arguments and return values need to be on their own line, per the code standard.  That change needs to be made to all routines

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+        p=p+f1(i-i_p+2)*f2(j-j_p+2)*f3(k-k_p+2)*table(i,j,k)
+      ENDDO
+    ENDDO
+  ENDDO
+
+ENDFUNCTION Interp_3D
+!
+!-------------------------------------------------------------------------------
+!> @brief This routine takes a 1D array of table labels (the points along a single
+!>        dimension at which table entries are defined) the interpolation point
```
aarograh: need an "and" or something after the closing paranthesis

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+    !Acending order
+    DO i_p=1,N_i
+      IF(pp < labels(i_p)) EXIT
+    ENDDO
+    IF(i_p > 1 .AND. i_p < N_i+1) THEN
+      f(1)=(labels(i_p)-pp)/(labels(i_p)-labels(i_p-1))
+      f(2)=1.0_SRK-f(1)
+    ENDIF
+  ELSE
+    !Decending order
```
aarograh: "Descending"

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+  REAL(SRK),INTENT(IN) :: labels(:),pp
+  REAL(SRK),INTENT(OUT) :: f(2)
+  INTEGER(SIK),INTENT(OUT) :: i_p,N_i
+  INTEGER(SIK) :: i
+  N_i=SIZE(labels)
+
+  !Determine labels index(s) and weights
+  f(:)=1.0_SRK
+  IF(labels(1) < labels(2)) THEN
+    !Acending order
```
aarograh: "Ascending"

src/Interpolators.f90
aarograh: I second this

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+MODULE Interpolators
+
+USE IntrType
+IMPLICIT NONE
+
+PUBLIC :: Interp
+
+!> @brief Generic interface to interpolate whatever is handed in.
+!>
+INTERFACE Interp
```
aarograh: Maybe more of a preference thing, but I'd prefer to just spell out the whole word `Interpolate` for the public interface...

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+!                                                                              !
+! Futility is a jointly-maintained, open-source project between the University !
+! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
+! can be found in LICENSE.txt in the head directory of this repository.        !
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module implementing interpolation routines for input
+!>        arrays returning either interpolated values or a flat "extrapolation"
+!>        when interpolation is out of range.
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+MODULE Interpolators
```
aarograh: Technically the code standard says this should be `InterpolatorsModule`... admittedly, that rule was made assuming most module define classes, and appending `Module` was a way to avoid name collisions, but we should still get in the habit of following the standard

src/Interpolators.f90
aarograh: One small detail: the return argument of a function should use `@returns` instead of `@param`

src/Interpolators.f90
```diff
@@ -0,0 +1,162 @@
...
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module implementing interpolation routines for input
+!>        arrays returning either interpolated values or a flat "extrapolation"
+!>        when  requested interpolation is out of range. Right now only 1D, 2D,
+!>        and 3D routines are available.
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+MODULE Interpolators
+
+USE IntrType
+IMPLICIT NONE
```
djabaay: Needs a `PRIVATE` statement.

src/Interpolators.f90
jpjones6: Must have been tired when I wrote those...

src/Interpolators.f90
jpjones6: So, I don't really have a preference in this regard. I went with Interp just to shorten what needs to be typed when someone wants to use the interpolator. I am fine with putting the whole word. Whatever the consensus is on this I'm happy to change it to.

src/Interpolators.f90
```diff
@@ -0,0 +1,162 @@
...
+!> @param axis_p point at which interpolation needs to be done to return a value
+!> @param f weights associated with the points on either side of axis_p
+!> @param i_p index of point directly ahead of axis_p in labels
+!> @param N_i number of entries of labels
+!>
+SUBROUTINE Get_points_and_weights(labels,axis_p,f,i_p,N_i)
+  REAL(SRK),INTENT(IN) :: labels(:),axis_p
+  REAL(SRK),INTENT(OUT) :: f(2)
+  INTEGER(SIK),INTENT(OUT) :: i_p,N_i
+  INTEGER(SIK) :: i
```
aarograh: This is also an unused variable

src/Interpolators.f90
```diff
@@ -0,0 +1,162 @@
...
+!> @param table contains data to be interpolated between defined at points
+!>        given in labels1, labels2, and labels 3
+!> @param axis_p point at which interpolation needs to be done to return a value
+!> @param data_p interpolated data found through linear interpolation to the point axis_p
+!>
+FUNCTION Interp_3D(labels1,labels2,labels3,table,axis_p) RESULT(data_p)
+  REAL(SRK),INTENT(IN) :: labels1(:),labels2(:),labels3(:),table(:,:,:),axis_p(:)
+  REAL(SRK) :: data_p,f1(2),f2(2),f3(2)
+  INTEGER(SIK) :: i,j,k,N_i,N_j,N_k,i_p,j_p,k_p
+
```
bscollin: There's a few DBC checks I think you need to add here:
  the length of labels aligns with the table
  If you assume the length of an y axis is >1, test it
  Each set of labels is sorted either ascending or descending  (write a helper function to call that tests this)

src/Interpolators.f90
```diff
@@ -0,0 +1,162 @@
...
+!
+!-------------------------------------------------------------------------------
+!> @brief This routine takes a 2D data table and linearly interpolates to find
+!>        a desired point.
+!> @param labels1 axis labels at which data points in table are defined for 1st dimension
+!> @param labels2 axis labels at which data points in table are defined for 2nd dimension
+!> @param labels3 axis labels at which data points in table are defined for 3rd dimension
+!> @param table contains data to be interpolated between defined at points
+!>        given in labels1, labels2, and labels 3
+!> @param axis_p point at which interpolation needs to be done to return a value
```
bscollin: I am really confused by this comment.  What does this do?

bscollin: ok... I get what it is now, but its still confusing  This is the wording in numpy.interp "The x-coordinates at which to evaluate the interpolated values."... I'd drop "x-" and its at least clearer.  I'm probably most confused by the word axis since it doesn't have anything to do with the axis (I'm thinking which direction in the array)

src/Interpolators.f90
```diff
@@ -0,0 +1,162 @@
...
+!>
+SUBROUTINE Get_points_and_weights(labels,axis_p,f,i_p,N_i)
+  REAL(SRK),INTENT(IN) :: labels(:),axis_p
+  REAL(SRK),INTENT(OUT) :: f(2)
+  INTEGER(SIK),INTENT(OUT) :: i_p,N_i
+  INTEGER(SIK) :: i
+  N_i=SIZE(labels)
+
+  !Determine labels index(s) and weights
+  f(:)=1.0_SRK
```
bscollin: how can both of these weights be 1?  Should it be 1,0?

src/Interpolators.f90
```diff
@@ -0,0 +1,162 @@
...
+  IF(labels(1) < labels(2)) THEN
```
bscollin: Search.f90 defines a Find and Binary_search option.  Admittedly, it only searches for increasing labels, but consider modifying those to search increasing or decreasing might be a better long term solution.

src/Interpolators.f90
```diff
@@ -0,0 +1,162 @@
...
+!-------------------------------------------------------------------------------
+!> @brief This routine takes a 2D data table and linearly interpolates to find
+!>        a desired point.
+!> @param labels1 axis labels at which data points in table are defined for 1st dimension
+!> @param labels2 axis labels at which data points in table are defined for 2nd dimension
+!> @param labels3 axis labels at which data points in table are defined for 3rd dimension
+!> @param table contains data to be interpolated between defined at points
+!>        given in labels1, labels2, and labels 3
+!> @param axis_p point at which interpolation needs to be done to return a value
+!> @param data_p interpolated data found through linear interpolation to the point axis_p
```
bscollin: Alright... this is a little picky... but data is also kind of a misnomer since its the interpolant (interpolated value).  When I think data, I think the actual table.

src/Interpolators.f90
jpjones6: If both weights need to be set they are. If only one needs to be set, i.e. when there is an out of range "extrapolation" needed then the loop/index magic in the 1D, 2D, and 3D functions handles that. In other words there is no need to find out which weight needs to be 0.

src/Interpolators.f90
jpjones6: I'll look into that.

src/Interpolators.f90
aarograh: `i_p` and `N_i` in `Get_points_and_weights` also need to be on different lines.  The same it true of function return values (`Interpolant` in the threw interp routines)

src/Interpolators.f90
jpjones6: These details have been added.

src/Interpolators.f90
jpjones6: These changes have been made.

src/Interpolators.f90
jpjones6: Added

src/Interpolators.f90
jpjones6: DBC checks have been added through helper functions.

src/Interpolators.f90
```diff
@@ -0,0 +1,146 @@
...
+  DO j=MAX(j_p-1,1),MIN(j_p,N_j)
+    DO i=MAX(i_p-1,1),MIN(i_p,N_i)
+      p=p+f1(i-i_p+2)*f2(j-j_p+2)*table(i,j)
+    ENDDO
+  ENDDO
+
+ENDFUNCTION Interp_2D
+!
+!-------------------------------------------------------------------------------
+!> @brief This routine takes a 2D data table and linearly interpolates to find
```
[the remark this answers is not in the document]
aarograh: Looks like this one still needs to be addressed?

src/Interpolators.f90
aarograh: Did we decide those weren't quite what's needed?

src/Interpolators.f90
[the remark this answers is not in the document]
jpjones6: Yup, looks like this one fell through the cracks. Changed it now.

src/Interpolators.f90
jpjones6: It looks like the Search module has upperbound and lowerbound functions that do bisection searches that could be drop in replacements for the linear searches I put in. I will switch to them and test to see if they do what's needed.

jpjones6: upperbound does exactly what's needed for the ascending sorted case, however, there is no function for a descending ordered array.

aarograh: yeah, most likely.  How much effort to change it?

src/ArrayUtils.f90
```diff
@@ -202,7 +206,7 @@ PURE SUBROUTINE getDelta_1DReal(r,rout,xi)
 !> @param r The input array of reals
 !> @param delta The optional input for whether the array is incremental or not
 !> @param tol The tolerance for comparing two real values
-!> @param sout The number of unique entries in the array r.
+!> @returns sout The number of unique entries in the array r.
 !>
 FUNCTION findNUnique_1DReal(r,delta,tol) RESULT(sout)
```
stimpsonsg: you didn't do this, but why the name sout?  Seems like something more descriptive would be better

src/ArrayUtils.f90
```diff
@@ -991,4 +995,69 @@ PURE FUNCTION findUpBound_1DReal(r,pos,delta,incl,tol) RESULT(val)
   ENDIF
 ENDFUNCTION findUpBound_1DReal
 !
+!-------------------------------------------------------------------------------
+!> @brief Routine that checks that the entries of r either monotonically increase or decrease.
+!>        Note this is limited to 1 dimensional real arrays.
+!> @param r input array to check for monotonicity of values
+!> @returns good boolean indicating whether the check has passed or not
```
stimpsonsg: don't personally love using the name "good"...@ag6...thoughts?

src/ArrayUtils.f90
```diff
@@ -991,4 +995,69 @@ PURE FUNCTION findUpBound_1DReal(r,pos,delta,incl,tol) RESULT(val)
...
+!>
+FUNCTION Is_Mono_Increasing(r) RESULT(good)
+  REAL(SRK),INTENT(IN) :: r(:)
+  INTEGER(SIK) :: i
+  LOGICAL(SBK) :: good
+  good=.TRUE.
```
stimpsonsg: this isn't needed

better to change it to false and delete the ELSE block

src/ArrayUtils.f90
aarograh: Personally I don't really care.  For a routine that's checkign things, "good" seems to be clear enough to me.  But I'm open to a change as well.

src/ArrayUtils.f90
```diff
@@ -991,4 +995,69 @@ PURE FUNCTION findUpBound_1DReal(r,pos,delta,incl,tol) RESULT(val)
...
+  LOGICAL(SBK) :: good
+
+  IF(SIZE(r) > 1) THEN
+    IF(r(1) < r(2)) THEN
+      good=Is_Mono_Increasing(r)
+    ELSE
+      good=Is_Mono_Decreasing(r)
+    ENDIF
+  ELSE
+    good=.FALSE.
```
stimpsonsg: should set this above, no else block

src/ArrayUtils.f90
```diff
@@ -991,4 +995,69 @@ PURE FUNCTION findUpBound_1DReal(r,pos,delta,incl,tol) RESULT(val)
...
+  INTEGER(SIK) :: i
+  LOGICAL(SBK) :: good
+  good=.TRUE.
+
+  IF(SIZE(r) > 1) THEN
+    DO i=2,SIZE(r)
+      IF(r(i) < r(i-1)) good=.FALSE.
+    ENDDO
+  ELSE
+    good=.FALSE.
```
stimpsonsg: should this return true if the size is 1?

unit_tests/testInterpolators/testInterpolators.f90
```diff
@@ -0,0 +1,269 @@
...
+  table_3D(2,3,3)=50.0_SRK
+  table_3D(3,3,3)=40.0_SRK
+  Interpolant=Interp(label1,label2,label3,table_3D,[1.6_SRK,2.4_SRK,1.0_SRK])
+  ASSERT(Interpolant .APPROXEQR. 85.6_SRK,'Incorrect Interpolant')
+  FINFO()"Interpolant=", Interpolant, "True value=",85.6_SRK
+
+  COMPONENT_TEST('3D Descending Out Of Range')
+  Interpolant=Interp(label1,label2,label3,table_3D,[0.5_SRK,2.4_SRK,1.0_SRK])
+  ASSERT(Interpolant .APPROXEQR. 76.0_SRK,'Incorrect Interpolant')
+  FINFO()"Interpolant=", Interpolant, "True value=",76.0_SRK
```
stimpsonsg: in general, we've been moving away from `FINFO` in favor of `ASSERT_EQ`

src/ArrayUtils.f90
jpjones6: Guilty until proven innocent. I like it.

src/ArrayUtils.f90
jpjones6: @aarograh, @HendersonSC, and I had a discussion on this. Our general consensus was if something is size 1 it is impossible to be increasing, decreasing, or generally monotonic and it would be likely that any routine that is relying on monotonic behavior probably requires more than a single scalar entry, hence being false.

src/Interpolators.f90
jpjones6: Changed lowerbound function to meet the needs for use in interpolator without changing it's original intended functionality.

src/Search.f90
```diff
@@ -168,114 +182,193 @@ PURE FUNCTION lower_bound_i(list,val) RESULT(j)
...
+      i=j+step
+      IF(SOFTGT(list(i),val,eps)) THEN
+        i=i+1
+        j=i
+        n=n-step-1
+      ELSE
+        n=step
+      ENDIF
+    ENDDO
+  ENDIF
```
aarograh: So I know we talked about this yesterday, but I just had an idea that would eliminate the need for duplicating these blocks of code without sacrificing performance... basically (I can't believe I'm saying this) use a procedure pointer:

```
PROCEDURE(SOFTLT),POINTER :: procPtr

! Setup stuff you already have in there

IF(list(1) < list(n)) THEN
  procPtr => SOFTLT
ELSE
  procPtr => SOFTGT
ENDIF

!Then use the original code, replacing SOFTLT with procPtr... also, give ProcPtr a better name
```

Give that a try and see how it works.  Let me know if you have any questions.

src/Search.f90
```diff
@@ -168,114 +182,193 @@ PURE FUNCTION lower_bound_i(list,val) RESULT(j)
...
 
-ENDFUNCTION lower_bound_r
+ENDFUNCTION lowerBound_r
```
aarograh: Shouldn't similar changes also exist in `lowerBound_i`?

src/Search.f90
```diff
@@ -11,95 +11,87 @@
...
-!> @param val the value to find
-!> @param tol the comparison tolerance to use for floating-point
-!> implementations. Optional
-!> @returns the index to the first element of the list that compares greater
-!> than the passed search value
+!> @brief Interface for getting "upper bounds"
 !>
-!> If the value is greater than or equal to the largest element of the list,
-!> returns SIZE + 1.
 INTERFACE upperBound
```
aarograh: Didn't we talk about renaming these functions to something better?  `FindFirstGreater` and `FindFirstGreaterOrEqual` or some such thing?  (I'm VERY open to better names as long as they're accurate/descriptive, but these ones are very unclear).

src/Search.f90
```diff
@@ -11,95 +11,87 @@
...
+!>       returns SIZE(list)+1, and if val is smaller than the smallest element of
+!>       list this returns 1.
+!>
+PURE FUNCTION upperBound_i(list,val) RESULT (i)
   INTEGER(SIK),INTENT(IN) :: list(:)
   INTEGER(SIK),INTENT(IN) :: val
   INTEGER(SIK) :: j
   !
-  INTEGER(SIK) :: n, step, i
+  INTEGER(SIK) :: n,step,i
```
aarograh: I think basically all of the functions in this module are missing some potential DBC opportunities.  You've added these `isMonotonic`, `isDecreasing`, and `isIncreasing` functions.  I think they should see some use throughout this module.

unit_tests/testSearch/testSearch.f90
```diff
@@ -14,11 +14,11 @@ PROGRAM testSearch
 
 IMPLICIT NONE
 
-INTEGER(SIK),DIMENSION(11) :: test_sorted = (/1,8,12,17,32,32,32,32,32,88,89/)
-INTEGER(SIK),DIMENSION(20) :: test_random = &
+INTEGER(SIK),DIMENSION(11) :: increasing_i = (/1,8,12,17,32,32,32,32,32,88,89/)
```
aarograh: I'd like to see some tests for decreasing as well

unit_tests/testSearch/testSearch.f90
```diff
@@ -48,150 +48,150 @@ PROGRAM testSearch
 !-------------------------------------------------------------------------------
 SUBROUTINE testLowerBound_i()
 
-  ASSERT(lowerBound(test_sorted, 0) == 1, "Wrong result for 0")
+  ASSERT(lowerBound(increasing_i, 0) == 1, "Wrong result for 0")
```
aarograh: Also, since you're in here, go ahead and switch the ASSERT/FINFO to the proper form of ASSERT_EQ

unit_tests/testInterpolators/testInterpolators.f90
jpjones6: Switched these instances of FINFO over.

src/Search.f90
jpjones6: Added it so behavior is consistent.

src/Search.f90
jpjones6: Made this change.

src/Search.f90
jpjones6: Done.

unit_tests/testSearch/testSearch.f90
jpjones6: Done

unit_tests/testSearch/testSearch.f90
```diff
@@ -2,12 +2,12 @@
 !                          Futility Development Group                          !
 !                             All rights reserved.                             !
 !                                                                              !
-! Futility is a jointly-maintained, open-source project between the University !
+! Futility is a jointly-maintained,open-source project between the University !
```
aarograh: Undo this... not sure how that crept in?

unit_tests/testSearch/testSearch.f90
```diff
@@ -2,12 +2,12 @@
...
 ! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
 ! can be found in LICENSE.txt in the head directory of this repository.        !
 !++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
 PROGRAM testSearch
-#include "UnitTest.h"
+#include"UnitTest.h"
```
aarograh: And this one too

src/Search.f90
aarograh: Basically add a couple `REQUIRE(isMonotonic(array))` in the getFirstGreater/getFirstGreaterEqual and I'm good with it

unit_tests/testSearch/testSearch.f90
aarograh: Jesse pointed out that there is intention to rework a couple of these routines later (specifically for the decreasing case), and that it's indirectly tested via testInterpolators, so we won't worry about this for now

aarograh:
Jesse ran the MPACT BASIC tests with this Futility branch and everything passed, and the interfaces he just added match the names of the old ones, so assuming the CI passes I'll merge this.

PR #238, opened 2020-03-09
Add code to rotate polygon objects in 90-degree increments

aarograh:
CASL Ticket # - 6276

### DIFF
--- a/src/Geom_Poly.f90
+++ b/src/Geom_Poly.f90
@@ -2082,7 +2082,7 @@ FUNCTION rotateClockwise(this,nrotations) RESULT(new)
   TYPE(PolygonType) :: new
   !
   INTEGER(SIK) :: i,irot,nrot
-  REAL(SRK) :: oldx,oldy,newx,newy
+  REAL(SRK) :: x,y,tmpx,tmpy
 
   REQUIRE(this%isInit)
   SELECTTYPE(this)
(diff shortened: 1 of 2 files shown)


src/Geom_Poly.f90
```diff
@@ -2062,4 +2065,86 @@ PURE SUBROUTINE createArcFromQuad(thisPoly,iquad,circle)
   ENDIF
 ENDSUBROUTINE createArcFromQuad
 !
+!-------------------------------------------------------------------------------
+!> @brief Rotates a polygon type in 90-degree rotations in the x-y plane
+!> @param this the polygon to rotate
+!> @param nrotations the number of clockwise 90-degree rotations to apply
+!> @returns new the rotation polygon type
```
stimpsonsg: rotation => rotated?

unit_tests/testGeom_Poly/testGeom_Poly.f90
```diff
@@ -2955,4 +2960,116 @@ SUBROUTINE testEquivalence()
...
+    ASSERT_APPROXEQ(new%quadEdge(2,i),-old%quadEdge(1,i),'%quadEdge, y')
+    ASSERT_APPROXEQ(new%quadEdge(3,i),old%quadEdge(3,i),'%quadEdge, z')
+  ENDDO !i
+
+  !Cleanup
+  CALL testGraph%clear()
+  CALL old%clear()
+  CALL new%clear()
+
+ENDSUBROUTINE testRotateClockwise
```
stimpsonsg: excellent unit test

src/Geom_Poly.f90
aarograh: yeah, good catch.  fixed

aarograh:
The last commit fixes a defect that wasn't caught initially.  All the rotation was done right, but for the quadratic edges, the final (x,y) values were assigned using the x value for both x and y.  Because my unit test used a simple triangle with no quadratic edges, this was missed.  I switched the unit test to use a polygon with quadratic edges to cover this bug.

There are some minor changes required on the MPACT side, so this needs to be held until that branch is ready as well.

PR #242, opened 2020-04-09
Temperature Conversion Functions

stimpsonsg:
Adds temperature conversion functions and unit testing.


HendersonSC:
I agree with @aarograh I think it is perfectly reasonable to expect the client to only use physically meaningful temperatures.

stimpsonsg:
Alright, I agree that the DBC check is of minimal benefit.  Upgraded all routines to elemental.

stimpsonsg:
when the heck is this CI ever going to fire off?

aarograh:
The build has passed: https://travis-ci.org/github/CASL/Futility

Something's up with the Travis to Github status update, but the code itself is fine.  Merging this in

PR #245, opened 2020-04-21
PETSc 3.12.4 Upgrade

stimpsonsg:
This MR encompasses the changes necessary to use PETSc 3.12.4.  I think eventually, we'll want to consider deprecating older versions of PETSc support, but at present should preserve any previous versions supported.

### DIFF
--- a/src/EigenvalueSolverTypes.f90
+++ b/src/EigenvalueSolverTypes.f90
@@ -38,13 +38,18 @@ USE Strings
 USE ForTeuchos_ParameterList
 #endif
 
+#ifdef FUTILITY_HAVE_PETSC
+#include <petscversion.h>
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
 USE PETSCSNES
+#endif
+#endif
 
 IMPLICIT NONE
 
 #ifdef FUTILITY_HAVE_PETSC
 #include <petscversion.h>
-#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6))
+#if ((PETSC_VERSION_MAJOR==3) && (PETSC_VERSION_MINOR==6))
 #include <petsc/finclude/petsc.h>
 #else
 #include <finclude/petsc.h>
--- a/src/EigenvalueSolverTypes.f90
+++ b/src/EigenvalueSolverTypes.f90
@@ -40,7 +40,7 @@ USE ForTeuchos_ParameterList
 
 #ifdef FUTILITY_HAVE_PETSC
 #include <petscversion.h>
-#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
+#if (((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6)) || (PETSC_VERSION_MAJOR>=4))
 USE PETSCSNES
 #endif
 #endif
@@ -49,7 +49,7 @@ IMPLICIT NONE
 
 #ifdef FUTILITY_HAVE_PETSC
 #include <petscversion.h>
-#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6))
+#if (((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6)) || (PETSC_VERSION_MAJOR>=4))
 #include <petsc/finclude/petsc.h>
 #else
 #include <finclude/petsc.h>
@@ -57,7 +57,7 @@ IMPLICIT NONE
 
 #ifdef FUTILITY_HAVE_SLEPC
 !SLEPC version == PETSC version
-#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6))
+#if (((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6)) || (PETSC_VERSION_MAJOR>=4))
 #include <slepc/finclude/slepcsys.h>
 #include <slepc/finclude/slepceps.h>
 #else
@@ -701,7 +701,7 @@ SUBROUTINE getResidual_EigenvalueSolverType_Base(solver,resid,its)
   SELECTTYPE(solver)
   TYPE IS(EigenvalueSolverType_SLEPc)
 #ifdef FUTILITY_HAVE_SLEPC
-#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6))
+#if (((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6)) || (PETSC_VERSION_MAJOR>=4))
     CALL EPSComputeError(solver%eps,0,EPS_ERROR_RELATIVE,resid,ierr)
 #else
     CALL EPSComputeRelativeError(solver%eps,0,resid,ierr)
--- a/src/EigenvalueSolverTypes.f90
+++ b/src/EigenvalueSolverTypes.f90
@@ -49,7 +49,7 @@ IMPLICIT NONE
 
 #ifdef FUTILITY_HAVE_PETSC
 #include <petscversion.h>
-#if ((PETSC_VERSION_MAJOR==3) && (PETSC_VERSION_MINOR==6))
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6))
 #include <petsc/finclude/petsc.h>
 #else
 #include <finclude/petsc.h>
--- a/src/EigenvalueSolverTypes.f90
+++ b/src/EigenvalueSolverTypes.f90
@@ -37,6 +37,9 @@ USE Strings
 #ifdef FUTILITY_HAVE_Trilinos
 USE ForTeuchos_ParameterList
 #endif
+
+USE PETSCSNES
+
 IMPLICIT NONE
 
 #ifdef FUTILITY_HAVE_PETSC
--- a/src/LinearSolverTypes.f90
+++ b/src/LinearSolverTypes.f90
@@ -64,13 +64,18 @@ USE MKL_PARDISO
 USE ForTeuchos_ParameterList
 #endif
 
+#ifdef FUTILITY_HAVE_PETSC
+#include <petscversion.h>
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
 USE PETSCKSP
+#endif
+#endif
 
 IMPLICIT NONE
 
 #ifdef FUTILITY_HAVE_PETSC
 #include <petscversion.h>
-#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6))
+#if ((PETSC_VERSION_MAJOR==3) && (PETSC_VERSION_MINOR==6))
 #include <petsc/finclude/petsc.h>
 #else
 #include <finclude/petsc.h>
@@ -529,8 +534,13 @@ SUBROUTINE init_LinearSolverType_Base(solver,Params,A)
           !PC calls
           CALL KSPGetPC(solver%ksp,pc,ierr)
           CALL PCSetType(pc,PCLU,iperr)
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
           CALL PCFactorSetMatSolverType(pc,MATSOLVERSUPERLU_DIST,iperr)
           CALL PCFactorSetUpMatSolverType(pc,iperr)
+#else
+          CALL PCFactorSetMatSolverPackage(pc,MATSOLVERSUPERLU_DIST,iperr)
+          CALL PCFactorSetUpMatSolverPackage(pc,iperr)
+#endif
 
 #else
           CALL eLinearSolverType%raiseError('Incorrect call to '// &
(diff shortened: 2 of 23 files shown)


src/EigenvalueSolverTypes.f90
```diff
@@ -37,6 +37,14 @@ MODULE EigenvalueSolverTypes
 #ifdef FUTILITY_HAVE_Trilinos
 USE ForTeuchos_ParameterList
 #endif
+
+#ifdef FUTILITY_HAVE_PETSC
+#include <petscversion.h>
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
```
stimpsonsg, line 43: I'll add that I am just checking that `PETSC_VERSION_MINOR > 6`, but I should ideally check which version of PETSc actually switched to build modules and go from there.

src/LinearSolverTypes_Multigrid.f90
```diff
@@ -551,11 +560,14 @@ SUBROUTINE setupPETScMG_LinearSolverType_Multigrid(solver,Params)
 
   !For now, only Galerkin coarse grid operators are supported.
   !  Galerkin means A_c = R*A*I
-  CALL PCMGSetGalerkin(solver%pc,PETSC_TRUE,iperr)
+!  CALL PCMGSetGalerkin(solver%pc,PETSC_TRUE,iperr)
+  !TODO: figure out Galerkin type
```
stimpsonsg, line 564: there are several TODOs like this for advanced features

bkochuna: Does this still work for older versions of PETSc, just not 3.12.4?

stimpsonsg: I need to go through and add some version diffs for things like this

It compiles with 3.6.4, but I'm not sure it's right for that version.

What I'll do is go back through and make sure everything is consistent with 3.6.4...then we can go back and look at things for 3.12.4.

bkochuna: 👍

src/LinearSolverTypes.f90
```diff
@@ -535,7 +535,7 @@ SUBROUTINE init_LinearSolverType_Base(solver,Params,A)
           CALL KSPGetPC(solver%ksp,pc,ierr)
           CALL PCSetType(pc,PCLU,iperr)
 #if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
-          CALL PCFactorSetMatSolverType(pc,MATSOLVERSUPERLU_DIST,iperr)
+          CALL PCFactorSetMatSolverType(pc,MATSOLVERSUPERLU,iperr)
```
stimpsonsg, line 538: this is something I'm carefully monitoring, but was required for the unit test to pass

src/LinearSolverTypes_Multigrid.f90
stimpsonsg: alright, I think I've reinstated everything for 3.6.4.  Only some of the multigrid stuff and (maybe) parallel SuperLU features are not functional for 3.12.4.  Otherwise, should be good to go.

src/EigenvalueSolverTypes.f90
aarograh: Yeah.  And maybe for good measure add an `|| PETSC_VERSION_MAJOR >= 4` just to attempt to future proof it a bit.

src/LinearSolverTypes.f90
```diff
@@ -63,6 +63,14 @@ MODULE LinearSolverTypes
 #ifdef FUTILITY_HAVE_Trilinos
 USE ForTeuchos_ParameterList
 #endif
+
+#ifdef FUTILITY_HAVE_PETSC
+#include <petscversion.h>
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
```
aarograh, line 69: So would it be possible to process this into its own preproc symbol during configuration?  I can see already that this combination (and my recommended addition above) are going to show up in a bunch of places.

It's definitely not a must, but it would look a little cleaner.

src/LinearSolverTypes.f90
aarograh: hmm... so it looks like the changed `Type` to `Package`.  But is `MATSOLVERSUPERLU` different from `MATSOLVERSUPERLU_DIST`?

src/LinearSolverTypes.f90
```diff
@@ -581,8 +594,13 @@ SUBROUTINE init_LinearSolverType_Base(solver,Params,A)
               CALL PCSetType(solver%pc,PCJACOBI,iperr)
             ELSEIF(TRIM(PreCondType)=='BJACOBI_ILU') THEN
               CALL PCSetType(solver%pc,PCBJACOBI,iperr)
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
+              CALL PetscOptionsSetValue(PETSC_NULL_OPTIONS,"-sub_ksp_type","preonly",iperr)
+              CALL PetscOptionsSetValue(PETSC_NULL_OPTIONS,"-sub_pc_type","ilu",iperr)
```
aarograh, line 599: What's the extra first argument doing?

src/MatrixTypes_PETSc.f90
```diff
@@ -311,7 +320,11 @@ SUBROUTINE transpose_PETScMatrixType(matrix)
   !This is to avoid a deadlock in IBarrier in MPICH
   CALL PetscCommBuildTwoSidedSetType(matrix%comm, &
       PETSC_BUILDTWOSIDED_ALLREDUCE,iperr)
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
+  CALL MatTranspose(matrix%a,MAT_INPLACE_MATRIX,matrix%a,iperr)
+#else
   CALL MatTranspose(matrix%a,MAT_REUSE_MATRIX,matrix%a,iperr)
+#endif
```
aarograh, line 327: Well that's an unnecessary name change...

src/ODESolverTypes.f90
```diff
@@ -30,20 +30,10 @@ MODULE ODESolverTypes
...
-#ifdef FUTILITY_HAVE_PETSC
-#include <petscversion.h>
-#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>=6))
-#include <petsc/finclude/petsc.h>
-#else
-#include <finclude/petsc.h>
-#endif
-!petscisdef.h defines the keyword IS, and it needs to be reset
-#undef IS
-#endif
```
aarograh, line 44: Why was this including petsc things when it didn't use them?

src/ParallelEnv.f90
```diff
@@ -24,25 +24,38 @@ MODULE ParallelEnv
...
 #undef IS
-PetscErrorCode  :: ierr
-PetscBool :: petsc_isinit
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
+INCLUDE 'mpif.h'
+#endif
 #else
 INCLUDE 'mpif.h'
 #endif
+#endif
```
aarograh, line 51: well you include mpif.h in 2 different places... Seems like they can be combined somehow, but maybe I'm wrong.  This si really ugly looking.

src/ParallelEnv.f90
```diff
@@ -24,25 +24,38 @@ MODULE ParallelEnv
...
 
+#ifdef FUTILITY_HAVE_PETSC
+PetscErrorCode  :: ierr
+PetscBool :: petsc_isinit
+#endif
```
aarograh, line 56: Any reason this had to move down to its own section?

unit_tests/testLinearSolver_Multigrid/testLinearSolver_Multigrid.f90
```diff
@@ -74,13 +81,15 @@ PROGRAM testLinearSolver_Multigrid
 REGISTER_SUBTEST('testInit',testInit)
 #ifdef FUTILITY_HAVE_PETSC
 REGISTER_SUBTEST('testPreAllocPETScInterpMat',testPreAllocPETScInterpMat)
+#if ((PETSC_VERSION_MAJOR<=3) && (PETSC_VERSION_MINOR<8))
```
aarograh, line 84: Are these the write combination?  This makes 3.7 a weird one that will go into all `#if` sections

src/EigenvalueSolverTypes.f90
stimpsonsg: reasonable

src/LinearSolverTypes.f90
stimpsonsg: it is...the latter supports parallel SuperLU

src/LinearSolverTypes.f90
stimpsonsg: there are at least a few minor version variants here and there

src/LinearSolverTypes.f90
stimpsonsg: image

src/LinearSolverTypes_Multigrid.f90
```diff
@@ -728,8 +756,13 @@ SUBROUTINE setSmoother_LinearSolverType_Multigrid(solver,smoother,iLevel,num_smo
...
       CALL KSPGetPC(ksp_temp,pc_temp,iperr)
       CALL PCSetType(pc_temp,PCLU,iperr)
-      IF(solver%MPIparallelEnv%nproc > 1) &
-          CALL PCFactorSetMatSolverPackage(pc_temp,MATSOLVERSUPERLU_DIST,iperr)
+      IF(solver%MPIparallelEnv%nproc > 1) THEN
+#if ((PETSC_VERSION_MAJOR>=3) && (PETSC_VERSION_MINOR>6))
+        CALL PCFactorSetMatSolverType(pc_temp,MATSOLVERSUPERLU_DIST,iperr)
+#else
+        CALL PCFactorSetMatSolverPackage(pc_temp,MATSOLVERSUPERLU_DIST,iperr)
+#endif
```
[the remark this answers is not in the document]
stimpsonsg: there are probably other issues with the multigrid stuff (I disabled for 3.12) that it's possible I didn't change it because the unit test didn't fail with it...the other changes here were necessary for compilation

src/MatrixTypes_PETSc.f90
stimpsonsg: yeah...not a huge fan

src/ODESolverTypes.f90
stimpsonsg: not a clue

src/ParallelEnv.f90
stimpsonsg: so...in older versions of PETSc, mpi.h was autoincluded...appears that that may not be the case anymore

src/ParallelEnv.f90
stimpsonsg: preference

src/LinearSolverTypes.f90
aarograh: We dont' want that option here?

src/LinearSolverTypes.f90
aarograh: Gotcha.  Weird.

src/LinearSolverTypes_Multigrid.f90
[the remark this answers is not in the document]
aarograh: Alright.  Did I miss where you disabled it?  Or was it in MPACT somewhere?

src/ParallelEnv.f90
aarograh: Alright.  But could we consolidate them?  There are still 2 `#else` branches that `INCLUDE 'mpif.h'`.  Seems like they could be a single one.

unit_tests/testLinearSolver_Multigrid/testLinearSolver_Multigrid.f90
stimpsonsg: understandable...basically the interface changed in 3.8 (and the routines this affects has an if >=8 check)..so it's "right" but obviously untested with 3.7 directly.

aarograh: yeah... just stuck out to me because a lot of your checks were > 3.6 and this is < 3.8.  So it was odd.  But I'm fine with it if that's what it needs to be.

src/LinearSolverTypes.f90
stimpsonsg: we do, but it has issues that are unresolved (i.e. the unit test fails with it)

src/LinearSolverTypes_Multigrid.f90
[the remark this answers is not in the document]
stimpsonsg: testLinearSolvers_Multigrid...there's a version check around the subtests

src/ParallelEnv.f90
stimpsonsg: nope, need two

src/LinearSolverTypes.f90
aarograh: Oh, gotcha.  I'm not sure I want to merge this without getting that resolved.

stimpsonsg: only the fhp3 solver uses it, so I'm fine proceeding without this resolved, so long as we're still supporting 3.6.4

src/LinearSolverTypes_Multigrid.f90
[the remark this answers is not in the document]
aarograh: whoops, I totally missed that there wasn't' a corresponding `#else` for that one.

src/LinearSolverTypes.f90
aarograh: Ok.  I could be talked into merging then.  Any idea how much effort to resolve it though?  It'd be nice to just get it done now.

stimpsonsg: So, I'm thrown some additional time into this exclusively, but no luck.  This might be something to discuss with PETSc devs.   Looking through their fortran examples, non use SUPERLU_DIST...only SUPERLU.

aarograh: Alright.  I'm good deferring for now then.  We can revisit when we upgrade petsc.

PR #246, opened 2020-04-22
WIP: Native anderson

jpjones6:
Implements native Anderson Acceleration ability using analytically derived minimization coefficients.

### DIFF
--- a/src/AndersonAcceleration.f90
+++ b/src/AndersonAcceleration.f90
@@ -6,53 +6,53 @@
 ! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
 ! can be found in LICENSE.txt in the head directory of this repository.        !
 !++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
-!> @brief Module provides an interface to Anderson Acceleration
-!>
-!> Currently supported TPLs include:
-!>  - Trilinos
-!>
-!> @par EXAMPLES
-!> @code
-!>
-!> @endcode
-!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module that encompasses the native implementation of
+!>        Anderson Acceleration in Futility.
 MODULE AndersonAccelerationTypes
 USE IntrType
 USE BLAS
-USE trilinos_interfaces
 USE Times
 USE ExceptionHandler
 USE ParameterLists
 USE ParallelEnv
 USE VectorTypes
+
 IMPLICIT NONE
 
 PRIVATE
-!
-! List of public members
+
+!List of public members
 PUBLIC :: eAndersonAccelerationType
 PUBLIC :: AndersonAccelerationType
 
-!> @brief the base eigenvalue solver type
+!> @brief the native Anderson Accelerator Type
 TYPE :: AndersonAccelerationType
   !> Initialization status
   LOGICAL(SBK) :: isInit=.FALSE.
-  !> NOX Solver type
-  INTEGER(SIK) :: id=-1
   !> Pointer to the distributed memory parallel environment
   TYPE(MPI_EnvType),POINTER :: MPIparallelEnv => NULL()
-  !> Pointer to the shared memory parallel environment TODO: eventually
-  TYPE(OMP_EnvType) :: OMPparallelEnv
-  !> iteration count
-  INTEGER(SIK) :: iter=0
-  !> size of nonlinear system
-  INTEGER(SIK) :: n=-1
-  !> depth of anderson solver
+  !> Size of solution vector/matrix
+  INTEGER(SIK) :: N=-1
+  !> Current iteration count
+  INTEGER(SIK) :: s=0
+  !> Iteration depth of anderson solver
   INTEGER(SIK) :: depth=-1
-  !> value of mixing parameter
+  !> Starting iteration for Anderson
+  INTEGER(SIK) :: start=1
+  !> Value of mixing parameter
   REAL(SRK) :: beta=0.0_SRK
-  !> Pointer to solution vector, x
-  CLASS(VectorType),POINTER :: X=>NULL()
+  !> Initial iterates
+  REAL(SRK),ALLOCATABLE :: x(:,:)
+  !> Gx vectors:
+  REAL(SRK),ALLOCATABLE :: Gx(:,:)
+  !> Difference vectors r=Gx-x
+  REAL(SRK),ALLOCATABLE :: r(:,:)
+  !> Anderson coefficient matrix
+  REAL(SRK),ALLOCATABLE :: A(:,:)
+  !> Anderson coefficients
+  REAL(SRK),ALLOCATABLE :: alpha(:)
+  !> Anderson right hand side
+  REAL(SRK),ALLOCATABLE :: b(:)
   !> Timer to measure solution time
   TYPE(TimerType) :: SolveTime
 !
@@ -65,18 +65,18 @@ TYPE :: AndersonAccelerationType
     !> @copydetails AndersonAccelerationType::clear_AndersonAccelerationType
     PROCEDURE,PASS :: clear => clear_AndersonAccelerationType
     !> @copybrief AndersonAccelerationType::step_AndersonAccelerationType
-    !> @copydetails AndersonAccelerationType::step_AndersonAccelerationType
+    !> @copydetails NativesAndersonAccelerationType::step_AndersonAccelerationType
     PROCEDURE,PASS :: step => step_AndersonAccelerationType
-    !> @copybrief AndersonAccelerationType::reset_AndersonAccelerationType
-    !> @copydetails AndersonAccelerationType::reset_AndersonAccelerationType
-    PROCEDURE,PASS :: reset => reset_AndersonAccelerationType
+    !> @copybrief AndersonAccelerationType::set_AndersonAccelerationType
+    !> @copydetails AndersonAccelerationType::set_AndersonAccelerationType
+    PROCEDURE,PASS :: set => set_AndersonAccelerationType
 ENDTYPE AndersonAccelerationType
 
 !> Exception Handler for use in Anderson Acceleration
 TYPE(ExceptionHandlerType),SAVE :: eAndersonAccelerationType
 
 !> Name of module
-CHARACTER(LEN=*),PARAMETER :: modName='ANDERSONACCELERATIONTYPE'
+CHARACTER(LEN=*),PARAMETER :: modName='AndersonAccelerationTypes'
 
 !
 !===============================================================================
@@ -85,12 +85,9 @@ CONTAINS
 !-------------------------------------------------------------------------------
 !> @brief Initializes the Anderson Acceleration Type with a parameter list
 !> @param pList the parameter list
-!>
 !> @param solver The anderson acceleration solver to act on
 !> @param MPIEnv The MPI environment description
-!> @param Params The aprameterlist
-!>
-!> This routine initializes the data spaces for the NOX Anderson Acceleration
+!> @param Params The anderson options parameter list
 !>
 SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
   CHARACTER(LEN=*),PARAMETER :: myName='init_AndersonAccelerationType'
@@ -98,19 +95,9 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
   TYPE(MPI_EnvType),INTENT(IN),TARGET :: MPIEnv
   TYPE(ParamType),INTENT(IN) :: Params
 
-  TYPE(ParamType) :: validParams
-  INTEGER(SIK) :: n,nlocal,depth,start
-  REAL(SRK) :: beta
-#ifdef FUTILITY_HAVE_Trilinos_NOX
-  TYPE(ParamType) :: tmpPL
-#endif
+  INTEGER(SIK) :: i,M
 
-  !Check to set up required and optional param lists.
-  !IF(.NOT.AndersonType_Paramsflag) CALL AndersionType_Declare_ValidParams()
-  !Validate against the reqParams and OptParams
-  validParams=Params
-  !CALL validParams%validate(AndersonType_reqParams)
-
+  !Check to ensure parallel environment is initialized
   IF(.NOT. MPIEnv%isInit()) THEN
     CALL eAndersonAccelerationType%raiseError('Incorrect input to '// &
         modName//'::'//myName//' - MPI Environment is not initialized!')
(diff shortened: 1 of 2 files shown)


src/AndersonAcceleration.f90
```diff
@@ -85,109 +85,56 @@ MODULE AndersonAccelerationTypes
...
+    ENDDO
+    solver%alpha(:)=0.0_SRK
+    solver%b(:)=0.0_SRK
+    solver%s=0
     solver%isInit=.TRUE.
   ELSE
-    CALL eAndersonAccelerationType%raiseError('Incorrect call to '// &
-        modName//'::'//myName//' - AndersonAccelerationType already initialized')
+    CALL eAndersonAccelerationType%raiseError('Incorrect call to '//modName// &
+        '::'//myName//' - TrilinosAndersonAccelerationType already initialized')
```
bscollin, line 137: whaaa?

jpjones6: Hah, woops. Leftover from when I was trying to split it to keep both.

src/AndersonAcceleration.f90
```diff
@@ -85,109 +85,56 @@ MODULE AndersonAccelerationTypes
...
-    SELECTTYPE(x=>solver%X); TYPE IS(TrilinosVectorType)
-      CALL Anderson_Init(solver%id,solver%depth,solver%beta,start,x%b)
-    ENDSELECT
-#else
-    CALL eAndersonAccelerationType%raiseError(modName//'::'//myName// &
-        ' - NOX (Trilinos) is not present in build')
-#endif
+    !Allocate member arrays
+    M=solver%depth+1
+    ALLOCATE(solver%x(solver%N,M),solver%Gx(solver%N,M),solver%r(solver%N,M), &
```
bscollin, line 122: do we really need to store x, Gx, and r?  Is that saving flops or is that just making life easier?

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
+
 ENDSUBROUTINE clear_AndersonAccelerationType
 !
 !-------------------------------------------------------------------------------
-!> @brief Single step of the Anderson acceleration
-!> @param solver The anderson solver to act on
+!> @brief Single step of Anderson acceleration acting on solution vectors
+!> @param solver Anderson solver to take step with
+!> @param New_Gx new solution iterate
 !>
```
bscollin, line 172: Declare and define function outputs

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
 !> This routine takes a single anderson acceleration step
 !>
-SUBROUTINE step_AndersonAccelerationType(solver)
+FUNCTION step_AndersonAccelerationType(solver,New_Gx) RESULT(x_tmp)
```
bscollin, line 175: These need to be vector classes so you can maintain your generic interface.

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
-#ifdef FUTILITY_HAVE_Trilinos_NOX
-  IF(solver%iter==0) THEN
-    CALL Anderson_Reset(solver%id)
-  ELSE
-    CALL Anderson_Update(solver%id)
+  REAL(SRK),INTENT(IN) :: New_Gx(:)
+
+  INTEGER(SIK) :: i,j,k,depth_s
+  REAL(SRK) :: x_tmp(solver%N)
```
bscollin, line 180: Define this up by New_Gx since its effectively part of the contract.

src/AndersonAcceleration.f90
```diff
@@ -85,109 +85,56 @@ MODULE AndersonAccelerationTypes
...
 !> @param pList the parameter list
-!>
 !> @param solver The anderson acceleration solver to act on
 !> @param MPIEnv The MPI environment description
-!> @param Params The aprameterlist
-!>
-!> This routine initializes the data spaces for the NOX Anderson Acceleration
+!> @param Params The anderson options parameter list
 !>
 SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
```
aarograh, line 92: Pass in a `FutilityComputingEnvironmentType` here instead of an mpi object.  Then delete `eAndersonAccelerationType` and just use the computing environment exceptional handler.

src/AndersonAcceleration.f90
```diff
@@ -6,53 +6,53 @@
...
-!>
-!> @par EXAMPLES
-!> @code
-!>
-!> @endcode
-!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module that encompasses the native implementation of
+!>        Anderson Acceleration in Futility.
 MODULE AndersonAccelerationTypes
 USE IntrType
```
aarograh, line 12: Need to see some DBC throughought this module.

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
+      solver%Gx(:,i+1)=solver%Gx(:,i)
+      solver%r(:,i+1)=solver%r(:,i)
+    ENDDO
+
+    !Set new vectors based on input
+    solver%Gx(:,1)=New_Gx(:)
+    solver%r(:,1)=solver%Gx(:,1)-solver%x(:,1)
+
+    !Get fit coefficients
+    IF(depth_s == 1) THEN
```
bscollin, line 198: Is this really needed?  Even if its depth 1, the linear solve is just going to do the division.  Is there really something different?  Seems like duplicate code for very little gain.

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
+ENDSUBROUTINE set_AndersonAccelerationType
+!
+!-------------------------------------------------------------------------------
+!> @brief Function which solves the system Ax=b by Gaussian elimination with
+!>        full(complete) pivoting. NOTE that A is destroyed in the process.
+!> @param A coefficient matrix in Ax=b
+!> @param b right hand side of problem Ax=b
+!> @returns x the solution vector of the problem Ax=b
+!
+FUNCTION GECP(A,b) RESULT(x)
```
aarograh, line 268: "Gaussian Elimination Complete Pivoting" I'm guessing?  If so, 2 comments:
1. I _hate_ these super shortened names...
2. Don't we have this implemented in Futility using MatrixTypes and VectorTypes?  Really should use those insteaad.

unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
```diff
@@ -60,539 +55,213 @@ PROGRAM testAndersonAcceleration
 !
 !-------------------------------------------------------------------------------
 SUBROUTINE testInit()
+
   CALL testAndAcc%init(mpiTestEnv,optList)
   ASSERT(testAndAcc%isInit,'%isInit')
```
aarograh, line 60: Need to see more of the `ASSERT_EQ` and similar updated macros throughout this test.

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
 !-------------------------------------------------------------------------------
-!> @brief Reset the Anderson acceleration solver
+!> @brief Set the initial iterate for the Anderson acceleration solver
 !> @param solver The anderson solver to act on
+!> @param x initial iterate
 !>
-!> This routine resets the counters for the anderson acceleration
-!>
-SUBROUTINE reset_AndersonAccelerationType(solver)
+SUBROUTINE set_AndersonAccelerationType(solver,x)
```
bscollin, line 253: use a vector type

unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
```diff
@@ -20,16 +20,15 @@ PROGRAM testAndersonAcceleration
...
 TYPE(ExceptionHandlerType),TARGET :: e
 TYPE(MPI_EnvType) :: mpiTestEnv
-TYPE(ParamType) :: pList, optList
+TYPE(ParamType) :: pList,optList
 TYPE(AndersonAccelerationType) :: testAndAcc
 
-
 !> set up default parameter list
 CALL optList%clear()
-CALL optList%add('AndersonAccelerationType->n',5_SIK)
```
bscollin, line 29: so it looks like you deleted an example where I had worked out the solution for multiple iterations but didn't really replace it with anything.

unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
```diff
@@ -60,539 +55,213 @@ PROGRAM testAndersonAcceleration
...
 
-  CALL testAndAcc%X%set(1,1.000_SRK)
-  CALL testAndAcc%X%set(2,2.000_SRK)
-  CALL testAndAcc%X%set(3,3.000_SRK)
-  CALL testAndAcc%X%set(4,4.000_SRK)
-  CALL testAndAcc%X%set(5,5.000_SRK)
-
-  CALL testAndAcc%step()
+  !Esure rate of convergence is correct:
+  ASSERT(AccErr(10)/AccErr(9) .APPROXEQ. 0.90109868784461988_SRK,'Wrong Anderson Conv. Rate')
```
bscollin, line 211: E**N**sure... and 0.901098 is the correct rate of convergence for this method?

unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
```diff
@@ -60,539 +55,213 @@ PROGRAM testAndersonAcceleration
...
-
-  CALL testAndAcc%step()
+  !Init Anderson
+  CALL testAndAcc%init(mpiTestEnv,optList)
 
-  WRITE(*,*) "step(9)"
-  DO i=1,5
-    CALL testAndAcc%X%get(i,tmp)
-    WRITE(*,*) i, tmp
+  !Exact chosen solution. Operator G will simply be taking a weighted average between
```
bscollin, line 91: why make this random?  I'm fine with you hard coding a random solution... but you could literally get a different result every time you run the unit test?  Is anything forcing a constant seed?

unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
```diff
@@ -60,539 +55,213 @@ PROGRAM testAndersonAcceleration
...
-  CALL testAndAcc%step()
-
-  WRITE(*,*) "step(14)"
-  DO i=1,5
-    CALL testAndAcc%X%get(i,tmp)
-    WRITE(*,*) i, tmp
+  !Generate a slow element wise "convergence rate"
+  DO i=1,SIZE(R)
+    CALL RANDOM_NUMBER(R(i))
+    R(i)=0.899999_SRK+0.2_SRK*(R(i)-0.5_SRK)
```
bscollin, line 152: .8999999??

src/AndersonAcceleration.f90
```diff
@@ -6,53 +6,53 @@
...
-!>
-!> Currently supported TPLs include:
-!>  - Trilinos
-!>
-!> @par EXAMPLES
-!> @code
-!>
-!> @endcode
-!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module that encompasses the native implementation of
```
bscollin, line 9: do we really need to say its a fortran module?

src/AndersonAcceleration.f90
```diff
@@ -6,53 +6,53 @@
...
+!>        Anderson Acceleration in Futility.
 MODULE AndersonAccelerationTypes
 USE IntrType
 USE BLAS
-USE trilinos_interfaces
```
bscollin, line 22: not a huge fan of dropping the old implementation but agree we need to have a consistent implementation

src/AndersonAcceleration.f90
```diff
@@ -6,53 +6,53 @@
...
   REAL(SRK) :: beta=0.0_SRK
-  !> Pointer to solution vector, x
-  CLASS(VectorType),POINTER :: X=>NULL()
+  !> Initial iterates
+  REAL(SRK),ALLOCATABLE :: x(:,:)
+  !> Gx vectors:
+  REAL(SRK),ALLOCATABLE :: Gx(:,:)
+  !> Difference vectors r=Gx-x
+  REAL(SRK),ALLOCATABLE :: r(:,:)
+  !> Anderson coefficient matrix
```
bscollin, line 50: matrix class

src/AndersonAcceleration.f90
```diff
@@ -6,53 +6,53 @@
...
+  REAL(SRK),ALLOCATABLE :: A(:,:)
+  !> Anderson coefficients
+  REAL(SRK),ALLOCATABLE :: alpha(:)
+  !> Anderson right hand side
+  REAL(SRK),ALLOCATABLE :: b(:)
```
bscollin, line 55: vector class... actually A and b could be held in the linear solver class

src/AndersonAcceleration.f90
```diff
@@ -65,18 +65,18 @@ MODULE AndersonAccelerationTypes
     !> @copydetails AndersonAccelerationType::clear_AndersonAccelerationType
     PROCEDURE,PASS :: clear => clear_AndersonAccelerationType
     !> @copybrief AndersonAccelerationType::step_AndersonAccelerationType
-    !> @copydetails AndersonAccelerationType::step_AndersonAccelerationType
+    !> @copydetails NativesAndersonAccelerationType::step_AndersonAccelerationType
```
bscollin, line 68: this is different than the others?  why?

src/AndersonAcceleration.f90
```diff
@@ -85,109 +85,56 @@ MODULE AndersonAccelerationTypes
...
-  validParams=Params
-  !CALL validParams%validate(AndersonType_reqParams)
+  INTEGER(SIK) :: i,M
 
+  !Check to ensure parallel environment is initialized
   IF(.NOT. MPIEnv%isInit()) THEN
     CALL eAndersonAccelerationType%raiseError('Incorrect input to '// &
         modName//'::'//myName//' - MPI Environment is not initialized!')
   ELSE
     solver%MPIparallelEnv => MPIEnv
```
bscollin, line 105: You never use this.  Do you need to?

src/AndersonAcceleration.f90
bscollin: the more I think about this, I'm a little uneasy about this being a function.  What happens if the return value isn't the same type as what gets passed in.  
```
REAL :: x(5)
x = foo(3) ! tries to return size (8) array... gets more complicated for classes.
```

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
+
+    !Set new vectors based on input
+    solver%Gx(:,1)=New_Gx(:)
+    solver%r(:,1)=solver%Gx(:,1)-solver%x(:,1)
+
+    !Get fit coefficients
+    IF(depth_s == 1) THEN
+      solver%A(1,1)=0.0_SRK
+      solver%b(1)=0.0_SRK
+      DO j=1,solver%N
```
bscollin, line 201: Are these standard norms?  If you do this with vector types, this could be generic for any vector type and use the built in norms.

bscollin: this would make x, Gx, and r an array pointers to vector types

src/AndersonAcceleration.f90
```diff
@@ -196,59 +143,217 @@ SUBROUTINE init_AndersonAccelerationType(solver,MPIEnv,Params)
...
 !
 !-------------------------------------------------------------------------------
-!> @brief Reset the Anderson acceleration solver
+!> @brief Set the initial iterate for the Anderson acceleration solver
 !> @param solver The anderson solver to act on
+!> @param x initial iterate
 !>
-!> This routine resets the counters for the anderson acceleration
-!>
-SUBROUTINE reset_AndersonAccelerationType(solver)
```
bscollin, line 246: still think this should be reset.  maybe you need both but they do the same thing.  I'm fine with that.  Just something that says, this is new, clear the history and start over.

src/AndersonAcceleration.f90
bscollin: while I'm not convinced full pivoting is there, its unclear if its necessary.  And if it is needed and we don't have it, we should implement it in linear solvers.

src/AndersonAcceleration.f90
jpjones6: You're absolutely right about this. Needs to be reset for mutli-state problems.

src/AndersonAcceleration.f90
jpjones6: Don't need it, like Aaron pointed out we can use the Futility parallel environment.

src/AndersonAcceleration.f90
jpjones6: Not really any norms going on here. Just sums of differences.

src/AndersonAcceleration.f90
jpjones6: Another hold over from having them split.

src/AndersonAcceleration.f90
jpjones6: In the case of r, saves a lot of flops...well it's a simple problem compared to the actual simulation costs but cuts the number of flops by a significant fraction. Also you have no choice with regard to Gx and x. You have to save the last depth+1 vectors.

src/AndersonAcceleration.f90
```diff
@@ -6,53 +6,53 @@
...
-  TYPE(OMP_EnvType) :: OMPparallelEnv
-  !> iteration count
-  INTEGER(SIK) :: iter=0
-  !> size of nonlinear system
-  INTEGER(SIK) :: n=-1
-  !> depth of anderson solver
+  !> Size of solution vector/matrix
+  INTEGER(SIK) :: N=-1
+  !> Current iteration count
+  INTEGER(SIK) :: s=0
```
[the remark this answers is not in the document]
jpjones6: I'm a transport methods guy. s is always our iteration counter. Stands for sweep #.

src/AndersonAcceleration.f90
jpjones6: Good idea!

src/AndersonAcceleration.f90
jpjones6: I agree, this really is only an initial implementation. Was wanting a preliminary review and make changes accordingly. Once things are more finalized I'll add them.

src/AndersonAcceleration.f90
jpjones6: 1. I hate super long names.
2. Matrix types / vector types don't have it specifically, but I doubt it's necessary to do full pivoting. Mostly I just wanted something I could drop in that's easy to use and get an answer instead of spending a bunch of time figuring out how to use all this vector and matrix types stuff.

unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
jpjones6: No biggie. Can switch them over. Are there also ASSERT_LE and the like?

src/AndersonAcceleration.f90
bscollin: strictly speaking, the vector class will hold its own parallel env.  When you use the BLAS routines, those should be handled under the hood.

aarograh:
We're going to close this.  Github doesn't make it super easy to compare different versions, and Jesse's new changes will look pretty radically different from the first pass.  He'll open a new pull request with the updates and we can do a fresh review then.  We can always look back at this closed PR if needed.

PR #248, opened 2020-04-27
Add error stop for Intel compiled code

rks171:
Description:
Getting a traceback with Intel requires calling the C abort procedure, so
this should be called from Futility_DBC when the compiler is Intel.

CASL Ticket # - 6264

### DIFF
--- a/src/Futility_DBC.f90
+++ b/src/Futility_DBC.f90
@@ -52,6 +52,15 @@ LOGICAL :: DBC_STOP_ON_FAIL=.TRUE.
 !This integer counts the total number of DBC_FAIL statements called
 INTEGER :: DBC_COUNTER=0
 !
+! Add interface to C abort function to exit intel-compiled code abnormally and generate
+! traceback information.
+#ifdef __INTEL_COMPILER
+   interface
+      subroutine abort() bind(C, name="abort")
+      end subroutine
+   end interface
+#endif
+!
 !===============================================================================
 CONTAINS
 !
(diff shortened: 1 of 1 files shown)


src/Futility_DBC.f90
```diff
@@ -52,6 +52,15 @@ MODULE Futility_DBC
...
 INTEGER :: DBC_COUNTER=0
 !
+! Add interface to C abort function to exit intel-compiled code abnormally and generate
+! traceback information.
+#ifdef __INTEL_COMPILER
+   interface
+      subroutine abort() bind(C, name="abort")
+      end subroutine
+   end interface
+#endif
```
aarograh, line 62: Maybe put this inside an fortran function called `backtrace`?  That way down at lines 97-101 we can just do `CALL backtrace()` without regard for which compiler we're using.

rks171: You want the backtrace procedure defined in the Futility_DBC module or a different one?

aarograh: Hmm, now that I think about it, my suggestion isn't a good one.  Renaming the `abort` function to `backtrace` is a bad idea if people try to use it for other debugging purposes...

aarograh: Yeah, so unfortunately intel seems to only provide a backtrace if you call abort.  So ignore my previous comment.

The one thing I would suggest is only calling abort if `STOP_ON_FAIL` is set to true.  This is basically always the case except in a couple unit testing situations, but I don't think we want to crash the code if the client explicitly said not to.

rks171: yeah good call

rks171: should be fixed

aarograh: Looks good

PR #249, opened 2020-05-14
Native Anderson Acceleration

jpjones6:
Implements a fully native arbitrary depth Anderson acceleration algorithm with efficient (direct) calculation of Anderson coefficients.

### DIFF
--- a/src/AndersonAcceleration.f90
+++ b/src/AndersonAcceleration.f90
@@ -37,11 +37,11 @@ TYPE :: AndersonAccelerationType
   !> Current iteration count
   INTEGER(SIK) :: s=0
   !> Iteration depth of Anderson solver
-  INTEGER(SIK) :: depth=-1
+  INTEGER(SIK) :: depth=1
   !> Starting iteration for Anderson
   INTEGER(SIK) :: start=1
   !> Value of mixing parameter
-  REAL(SRK) :: beta=0.0_SRK
+  REAL(SRK) :: beta=0.5_SRK
   !> Initial iterates
   CLASS(VectorType),ALLOCATABLE :: x(:)
   !> Gx vectors:
@@ -86,25 +86,25 @@ CONTAINS
 !> @param ce The computing environment to use for the calculation
 !>
 SUBROUTINE init_AndersonAccelerationType(solver,ce,Params)
-  CHARACTER(LEN=*),PARAMETER :: myName='init_AndersonAccelerationType'
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
   TYPE(FutilityComputingEnvironment),TARGET,INTENT(IN) :: ce
   TYPE(ParamType),INTENT(IN) :: Params
 
+  CHARACTER(LEN=*),PARAMETER :: myName='init_AndersonAccelerationType'
   TYPE(ParamType) :: LSparams
   INTEGER(SIK) :: i,j,m
 
   REQUIRE(Params%has('AndersonAccelerationType->n'))
-  REQUIRE(Params%has('AndersonAccelerationType->depth'))
-  REQUIRE(Params%has('AndersonAccelerationType->beta'))
-  REQUIRE(Params%has('AndersonAccelerationType->start'))
   REQUIRE(.NOT.solver%isInit)
 
   !Pull Data from Parameter List
   CALL Params%get('AndersonAccelerationType->n',solver%n)
-  CALL Params%get('AndersonAccelerationType->depth',solver%depth)
-  CALL Params%get('AndersonAccelerationType->beta',solver%beta)
-  CALL Params%get('AndersonAccelerationType->start',solver%start)
+  IF(Params%has('AndersonAccelerationType->depth')) &
+      CALL Params%get('AndersonAccelerationType->depth',solver%depth)
+  IF(Params%has('AndersonAccelerationType->beta')) &
+      CALL Params%get('AndersonAccelerationType->beta',solver%beta)
+  IF(Params%has('AndersonAccelerationType->start')) &
+      CALL Params%get('AndersonAccelerationType->start',solver%start)
 
   IF(solver%n < 1) CALL ce%exceptHandler%raiseError('Incorrect input to '//modName// &
       '::'//myName//' - Number of unkowns (n) must be greater than 0!')
@@ -178,13 +178,12 @@ SUBROUTINE clear_AndersonAccelerationType(solver)
       DEALLOCATE(solver%tmpvec)
     ENDIF
     DEALLOCATE(solver%alpha)
-    CALL solver%LS%clear()
+    IF(solver%LS%isinit) CALL solver%LS%clear()
     solver%s=0
     solver%n=-1
-    solver%depth=-1
-    solver%start=0
-    solver%depth=-1
-    solver%beta=0.0_SRK
+    solver%depth=1
+    solver%start=1
+    solver%beta=0.5_SRK
     solver%isInit=.FALSE.
   ENDIF
 
@@ -294,10 +293,10 @@ ENDSUBROUTINE step_AndersonAccelerationType
 !> @param x Initial iterate solve is starting from
 !>
 SUBROUTINE reset_AndersonAccelerationType(solver,x)
-  CHARACTER(LEN=*),PARAMETER :: myName='reset_AndersonAccelerationType'
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
   CLASS(VectorType),INTENT(INOUT) :: x
 
+  CHARACTER(LEN=*),PARAMETER :: myName='reset_AndersonAccelerationType'
   INTEGER(SIK) :: i,j
 
   REQUIRE(x%n == solver%n)
--- a/src/AndersonAcceleration.f90
+++ b/src/AndersonAcceleration.f90
@@ -311,13 +311,17 @@ SUBROUTINE reset_AndersonAccelerationType(solver,x)
 
       !!!TODO Once missing BLAS interfaces have been added for the following vector types
       !do away with this error catch to allow use of these with Anderson Acceleration.
+#ifdef HAVE_MPI
       SELECT TYPE(x);TYPE IS(NativeDistributedVectorType)
         CALL solver%ce%exceptHandler%raiseError('Incorrect call to '//modName// &
             '::'//myName//' - Input vector type not supported!')
+#endif
+#ifdef FUTILITY_HAVE_Trilinos
       TYPE IS(TrilinosVectorType)
         CALL solver%ce%exceptHandler%raiseError('Incorrect call to '//modName// &
             '::'//myName//' - Input vector type not supported!')
       ENDSELECT
+#endif
 
       CALL VectorResembleAlloc(solver%x,x,solver%depth+1)
       CALL VectorResembleAlloc(solver%Gx,x,solver%depth+1)
--- a/src/AndersonAcceleration.f90
+++ b/src/AndersonAcceleration.f90
@@ -266,7 +266,7 @@ SUBROUTINE step_AndersonAccelerationType(solver,x_new)
       CALL BLAS_axpy(solver%Gx(1),x_new,solver%beta)
       CALL BLAS_axpy(solver%x(1),x_new,1.0_SRK-solver%beta)
       CALL solver%reset(x_new)
-   ELSE
+    ELSE
       !Get accelerated solution
       CALL x_new%set(0.0_SRK)
       DO i=1,depth_s+1
@@ -298,7 +298,7 @@ ENDSUBROUTINE step_AndersonAccelerationType
 SUBROUTINE reset_AndersonAccelerationType(solver,x)
   CHARACTER(LEN=*),PARAMETER :: myName='reset_AndersonAccelerationType'
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
-  CLASS(VectorType),INTENT(IN) :: x
+  CLASS(VectorType),INTENT(INOUT) :: x
 
   TYPE(ParamType) :: vecparams
   INTEGER(SIK) :: i,j,m
--- a/src/AndersonAcceleration.f90
+++ b/src/AndersonAcceleration.f90
@@ -41,15 +41,17 @@ TYPE :: AndersonAccelerationType
   !> Value of mixing parameter
   REAL(SRK) :: beta=0.0_SRK
   !> Initial iterates
-  REAL(SRK),ALLOCATABLE :: x(:,:)
+  CLASS(VectorType),ALLOCATABLE :: x(:)
   !> Gx vectors:
-  REAL(SRK),ALLOCATABLE :: Gx(:,:)
+  CLASS(VectorType),ALLOCATABLE :: Gx(:)
   !> Difference vectors r=Gx-x
-  REAL(SRK),ALLOCATABLE :: r(:,:)
+  CLASS(VectorType),ALLOCATABLE :: r(:)
   !> Anderson coefficients
   REAL(SRK),ALLOCATABLE :: alpha(:)
   !> Linear solver
   TYPE(LinearSolverType_Direct) :: LS
+  !> Futility computing environment
+  TYPE(FutilityComputingEnvironment),POINTER :: ce
 !
 !List of Type Bound Procedures
   CONTAINS
@@ -70,6 +72,9 @@ ENDTYPE AndersonAccelerationType
 !> Name of module
 CHARACTER(LEN=*),PARAMETER :: modName='AndersonAccelerationTypes'
 
+!> Intermediate calculation vector
+CLASS(VectorType),ALLOCATABLE :: tmpvec
+
 !
 !===============================================================================
 CONTAINS
@@ -85,7 +90,7 @@ SUBROUTINE init_AndersonAccelerationType(solver,ce,Params)
   CHARACTER(LEN=*),PARAMETER :: myName='init_AndersonAccelerationType'
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
   TYPE(ParamType),INTENT(IN) :: Params
-  TYPE(FutilityComputingEnvironment),INTENT(IN) :: ce
+  TYPE(FutilityComputingEnvironment),TARGET :: ce
 
   TYPE(ParamType) :: LSparams
   INTEGER(SIK) :: i,j,m
@@ -141,14 +146,12 @@ SUBROUTINE init_AndersonAccelerationType(solver,ce,Params)
       ENDDO
     ENDIF
 
-    !Allocate member arrays
+    !Allocate/create member arrays and vectors that can be done now and associate pointers
     m=solver%depth+1
-    ALLOCATE(solver%x(solver%n,m),solver%Gx(solver%n,m),solver%r(solver%n,m),solver%alpha(m))
-    solver%x(:,:)=0.0_SRK
-    solver%Gx(:,:)=0.0_SRK
-    solver%r(:,:)=0.0_SRK
+    ALLOCATE(solver%alpha(m))
     solver%alpha(:)=0.0_SRK
     solver%s=0
+    solver%ce => ce
     solver%isInit=.TRUE.
   ELSE
     CALL ce%exceptHandler%raiseError('Incorrect call to '//modName// &
@@ -164,18 +167,29 @@ ENDSUBROUTINE init_AndersonAccelerationType
 SUBROUTINE clear_AndersonAccelerationType(solver)
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
 
+  INTEGER(SIK) :: i
+
   IF(solver%isInit) THEN
+    IF(ALLOCATED(solver%x)) THEN
+      DO i=1,solver%depth+1
+        CALL solver%x(i)%clear()
+        CALL solver%Gx(i)%clear()
+        CALL solver%r(i)%clear()
+      ENDDO
+      CALL tmpvec%clear()
+      DEALLOCATE(solver%x)
+      DEALLOCATE(solver%Gx)
+      DEALLOCATE(solver%r)
+      DEALLOCATE(tmpvec)
+    ENDIF
+    DEALLOCATE(solver%alpha)
+    CALL solver%LS%clear()
     solver%s=0
     solver%n=-1
     solver%depth=-1
     solver%start=0
     solver%depth=-1
     solver%beta=0.0_SRK
-    DEALLOCATE(solver%x)
-    DEALLOCATE(solver%Gx)
-    DEALLOCATE(solver%r)
-    DEALLOCATE(solver%alpha)
-    CALL solver%LS%clear()
     solver%isInit=.FALSE.
   ENDIF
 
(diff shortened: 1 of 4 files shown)


src/AndersonAcceleration.f90
```diff
@@ -6,55 +6,50 @@
...
+  !> Starting iteration for Anderson
+  INTEGER(SIK) :: start=1
+  !> Value of mixing parameter
   REAL(SRK) :: beta=0.0_SRK
-  !> Pointer to solution vector, x
-  CLASS(VectorType),POINTER :: X=>NULL()
-  !> Timer to measure solution time
-  TYPE(TimerType) :: SolveTime
+  !> Initial iterates
+  REAL(SRK),ALLOCATABLE :: x(:,:)
```
bscollin, line 44: I thought we were changing the solver to use vector classes.

jpjones6: We had talked about it, and I tried that. Switched to Vector classes and BLAS calls. Made the code slower (only 4% but still). Without an increase in performance I don't see much of a point in using vector classes here, so I switched back. Everything we want to accelerate in MPACT comes as some form/shape of REAL array (not vectors) which needs to be repacked into a 1D array. Any packing/repacking and the Anderson code itself is much cleaner when not using vector classes. Also, if anyone outside of the MPACT team wanted to use the Anderson ability not having to go through vector types/classes may help a beginner use this more easily.

src/AndersonAcceleration.f90
```diff
@@ -72,183 +67,239 @@ MODULE AndersonAccelerationTypes
     PROCEDURE,PASS :: reset => reset_AndersonAccelerationType
 ENDTYPE AndersonAccelerationType
 
-!> Exception Handler for use in Anderson Acceleration
-TYPE(ExceptionHandlerType),SAVE :: eAndersonAccelerationType
-
 !> Name of module
-CHARACTER(LEN=*),PARAMETER :: modName='ANDERSONACCELERATIONTYPE'
+CHARACTER(LEN=*),PARAMETER :: modName='AndersonAccelerationTypes'
```
stimpsonsg, line 71: why this change?  I'm not exactly opposed to it, but out of Futility, there are only three that are not full caps.

We need them to be consistent (so those other ones will need to change to all caps too) or we change all to camelcase.

jpjones6: I'm used to seeing camelcase on the MPACT side, so when I added the missing s on the end of the mod name I switched it over. I don't REALLY have a preference, but the logical choice to me would be to be consistent with MPACT.

src/AndersonAcceleration.f90
stimpsonsg: are any of the arrays more than 2 dimensional?

jpjones6:
New version using vector types and BLAS is up. Caveats: 

1) Only RealVectorType and PetcVectorTypes are usable right now. NativeDistributed and Trilinos type vectors are both missing certain interfaces to needed BLAS routines.

2) Creation of the storage vectors is deferred until the first call of the set/reset Anderson routine. Did this because that is the first time a vector is actually handed to Anderson for any purpose. Could alternatively move it to the init routine with the addition of a parameter list entry specifying what vector type will be used. Didn't like that option since that will require consistency between where Anderson is inited and set/reset in the client code....I don't trust that will always happen plus it seems less flexible.

src/AndersonAcceleration.f90
```diff
@@ -85,7 +90,7 @@ SUBROUTINE init_AndersonAccelerationType(solver,ce,Params)
   CHARACTER(LEN=*),PARAMETER :: myName='init_AndersonAccelerationType'
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
   TYPE(ParamType),INTENT(IN) :: Params
-  TYPE(FutilityComputingEnvironment),INTENT(IN) :: ce
+  TYPE(FutilityComputingEnvironment),TARGET :: ce
```
bscollin, line 93: Don't we still need intent on this?

src/AndersonAcceleration.f90
```diff
@@ -205,40 +219,36 @@ SUBROUTINE step_AndersonAccelerationType(solver,x_new)
...
-      ENDDO
+      CALL BLAS_copy(solver%r(2),tmpvec)
+      CALL BLAS_axpy(solver%r(1),tmpvec,-1.0_SRK)
+      tmpA=BLAS_dot(tmpvec,tmpvec)
+      tmpb=BLAS_dot(solver%r(2),tmpvec)
       solver%alpha(1)=tmpb/tmpA
     ELSEIF(depth_s > 1) THEN
       !Construct coefficient matrix, right hand side, and solve for fit coefficients
       DO i=1,depth_s
+        CALL BLAS_copy(solver%r(i),x_new)
```
bscollin, line 242: You made a comment before that the blas routines make the math less readable... could always add a comment to communicate exactly what math you are trying to implement

src/AndersonAcceleration.f90
```diff
@@ -251,23 +261,33 @@ SUBROUTINE step_AndersonAccelerationType(solver,x_new)
...
     ENDDO
 
-    !Get accelerated solution
-    x_new(:)=0.0_SRK
-    DO i=1,depth_s+1
-      x_new(:)=x_new(:)+solver%alpha(i)*(solver%x(:,i)+solver%beta*solver%r(:,i))
-    ENDDO
+    !Ensure Anderson coefficient solve did not fail due to bad input vector
+    IF(ISNAN(solver%alpha(depth_s+1))) THEN
+      !Bad Anderson coefficient solve, revert to under-relaxation and reset Anderson solver
```
bscollin, line 266: should we make sure we don't drop into some infinite loop where we keep failing?

src/AndersonAcceleration.f90
```diff
@@ -278,15 +298,63 @@ SUBROUTINE step_AndersonAccelerationType(solver,x_new)
...
+  CLASS(VectorType),INTENT(IN) :: x
+
+  TYPE(ParamType) :: vecparams
+  INTEGER(SIK) :: i,j,m
 
-  INTEGER(SIK) :: i,j
+  REQUIRE(x%n == solver%n)
 
-  REQUIRE(SIZE(x) == solver%n)
+  !If this is the first call to set/reset must actually create vectors of needed type
```
bscollin, line 310: Vector types has a base capability to create a new vector identical to an old vector: `VectorResemble`

bscollin: also... what blas interface do you need that doesn't exist?

src/AndersonAcceleration.f90
jpjones6: Let's put it in. We are missing Intent on the futility computing env in multiple locations in MPACT (was just emulating an init routine in MPACT).

src/AndersonAcceleration.f90
jpjones6: We shouldn't end up in an infinite loop..."shouldn't". I've made the behavior here when Anderson fails to just do simple under-relaxation (like default) using the mixing parameter, and then that call to reset will throw away all the past solutions (that would have caused the failure). If we ran into a case where it would fail over and over it would just act like default.

jpjones6: I need to add a test for this ability in the unit test.

src/AndersonAcceleration.f90
jpjones6: Okay, I will look at the resemble routine and see about simplifying the reset/set function. For Trilinos vectors it is dot product (as hard as that is to believe I am looking right at the interface, it's got NativeDistributed, Petsc, and Real...no Trilinos). For NativeDistributed it's the _scal routine. Shouldn't be hard to add.

jpjones6: On the subject of BLAS interfaces. The vector interfaces to them have their arguments out of order...for instance _axpy for vectors actually has the arguments in the order xya...a little OCD for sure, but annoying.

src/AndersonAcceleration.f90
jpjones6: Improved the rescue to look for when Alpha = 0.0 this signals that a duplicate or other bad iterate has been handed in and added testing.

src/AndersonAcceleration.f90
bscollin: None of this was fixed yet

src/AndersonAcceleration.f90
jpjones6: Since the math is simple enough here I don't think we need to go that far.

src/AndersonAcceleration.f90
jpjones6: Everything is changed over now.

src/AndersonAcceleration.f90
jpjones6: Latest push adds resemble routines for allocatable vectors and arrays of vectors as well as associated testing. Also changed Anderson over to using them with a short check for vector types that are not currently fully supported in BLAS.

src/AndersonAcceleration.f90
```diff
@@ -6,55 +6,54 @@
...
-!>  - Trilinos
-!>
-!> @par EXAMPLES
-!> @code
-!>
-!> @endcode
-!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module that encompasses the native implementation of
+!>        Anderson Acceleration in Futility.
 MODULE AndersonAccelerationTypes
```
aarograh, line 11: A pretty minor one, but our style guide says modules should be named "<filename without extension>Module"... so for this, "AndersonAccelerationModule".  Granted, virtually nothing in Futility has been updated to follow this, but since this module is unused in downstream code so far, let's try to start it off correctly.

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
     PROCEDURE,PASS :: reset => reset_AndersonAccelerationType
 ENDTYPE AndersonAccelerationType
 
-!> Exception Handler for use in Anderson Acceleration
-TYPE(ExceptionHandlerType),SAVE :: eAndersonAccelerationType
-
 !> Name of module
-CHARACTER(LEN=*),PARAMETER :: modName='ANDERSONACCELERATIONTYPE'
-
+CHARACTER(LEN=*),PARAMETER :: modName='AndersonAccelerationTypes'
```
aarograh, line 75: And you'll want to change this to match the updated module name

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
 !
 !===============================================================================
 CONTAINS
 !
 !-------------------------------------------------------------------------------
-!> @brief Initializes the Anderson Acceleration Type with a parameter list
+!> @brief Initializes the Anderson Acceleration Type with an input parameter list
 !> @param pList the parameter list
```
aarograh, line 82: looks like this is unneeded since you have `Params`

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
-  IF(.NOT. MPIEnv%isInit()) THEN
-    CALL eAndersonAccelerationType%raiseError('Incorrect input to '// &
-        modName//'::'//myName//' - MPI Environment is not initialized!')
-  ELSE
-    solver%MPIparallelEnv => MPIEnv
-  ENDIF
+  REQUIRE(Params%has('AndersonAccelerationType->n'))
+  REQUIRE(Params%has('AndersonAccelerationType->depth'))
+  REQUIRE(Params%has('AndersonAccelerationType->beta'))
+  REQUIRE(Params%has('AndersonAccelerationType->start'))
```
aarograh, line 99: Just out of curiosity, is it possible to set reasonable defaults for some of these?  Or are the values fairly application-specific?

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
-  CALL validParams%get('AndersonAccelerationType->depth',depth)
-  CALL validParams%get('AndersonAccelerationType->beta',beta)
-  CALL validParams%get('AndersonAccelerationType->start',start)
-  CALL validParams%clear()
+  CALL Params%get('AndersonAccelerationType->n',solver%n)
+  CALL Params%get('AndersonAccelerationType->depth',solver%depth)
+  CALL Params%get('AndersonAccelerationType->beta',solver%beta)
+  CALL Params%get('AndersonAccelerationType->start',solver%start)
 
   IF(.NOT. solver%isInit) THEN
```
aarograh, line 107: So, instead of a giant IF block, just `REQUIRE(.NOT.solver%isInit)` instead

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
+      !Construct coefficient matrix, right hand side, and solve for fit coefficients
+      DO i=1,depth_s
+        CALL BLAS_copy(solver%r(i),x_new)
+        CALL BLAS_axpy(solver%r(depth_s+1),x_new,-1.0_SRK)
+        tmpb=BLAS_dot(solver%r(depth_s+1),x_new)
+        CALL solver%LS%b%set(i,tmpb)
+        DO k=1,i
+          CALL BLAS_copy(solver%r(depth_s+1),solver%tmpvec)
+          CALL BLAS_axpy(solver%r(k),solver%tmpvec,-1.0_SRK)
+          tmpA=BLAS_dot(solver%tmpvec,x_new)
```
aarograh, line 247: bugs me that all the BLAS stuff are subroutines except this one... but I know we can't do anything about that

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
-!> @param solver The anderson solver to act on
-!>
-!> This routine resets the counters for the anderson acceleration
+!> @brief Set or reset the initial iterate and linear solver state for the Anderson solver
+!> @param solver The Anderson solver to act on
+!> @param x Initial iterate solve is starting from
 !>
-SUBROUTINE reset_AndersonAccelerationType(solver)
+SUBROUTINE reset_AndersonAccelerationType(solver,x)
+  CHARACTER(LEN=*),PARAMETER :: myName='reset_AndersonAccelerationType'
```
aarograh, line 300: slide this down below the dummy arguments

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
+  REQUIRE(x%n == solver%n)
+
+  !If this is the first call to set/reset must actually create vectors of needed type
+  IF(solver%s == 0) THEN
+    IF(.NOT.ALLOCATED(solver%x)) THEN
+
+      !!!TODO Once missing BLAS interfaces have been added for the following vector types
+      !do away with this error catch to allow use of these with Anderson Acceleration.
+#ifdef HAVE_MPI
+      SELECT TYPE(x);TYPE IS(NativeDistributedVectorType)
```
aarograh, line 315: `TYPE IS(...)` should go on its own line

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
+        CALL solver%ce%exceptHandler%raiseError('Incorrect call to '//modName// &
+            '::'//myName//' - Input vector type not supported!')
 #endif
+#ifdef FUTILITY_HAVE_Trilinos
+      TYPE IS(TrilinosVectorType)
+        CALL solver%ce%exceptHandler%raiseError('Incorrect call to '//modName// &
+            '::'//myName//' - Input vector type not supported!')
+      ENDSELECT
+#endif
```
aarograh, line 324: So the fact that SELECTTYPE and ENDSELECT are in 2 different sets of preproc macros isn't good... I'd recommend this:

```
SELECTTYPE
#ifdef HAVE_MPI
  !Check for native distributed vector
#endif
#ifdef FUTILITY_HAVE_Trilinos
  !Check for trilinos vector
#endif
CLASS DEFAULT
  CONTINUE
ENDSELECT
```

src/AndersonAcceleration.f90
```diff
@@ -72,183 +71,288 @@ MODULE AndersonAccelerationTypes
...
+  ENDIF
+
+  !Reset iteration counter
+  solver%s=0_SIK
+
+  !Grab initial iterate to initiate Anderson from
+  CALL BLAS_copy(x,solver%x(1))
+
+  !Reset Anderson coefficient matrix
+  IF(solver%LS%isinit) THEN
```
aarograh, line 343: Should we just REQUIRE this?  Seems like you shouldn't be calling reset on something that was never initialized to begin with

src/AndersonAcceleration.f90
jpjones6: I'm glad you mentioned this. I was leaning toward changing it, but since the module already existed before I gutted and remade it I didn't want to change it without a second opinion.

src/AndersonAcceleration.f90
jpjones6: Got it, also changed the copybriefs/details

src/AndersonAcceleration.f90
jpjones6: Removed it

src/AndersonAcceleration.f90
jpjones6: So, defaults are/will be set in XMLProc for a normal run by a user. Depth=1, beta=0.5, start=1 is a conservative choice that should at least work "okay" for essentially any problem.

aarograh: Ok.  I'm kind of in favor of setting these defaults internally as opposed to using XMLProc... thoughts @stimpsonsg @bscollin ?

src/AndersonAcceleration.f90
jpjones6: I like it.

src/AndersonAcceleration.f90
jpjones6: Are we changing our format to do it that way?

src/AndersonAcceleration.f90
jpjones6: I agree. I'll change it here. We have a LOT of instances in MPACT with the ; for this.

src/AndersonAcceleration.f90
aarograh: We've pretty much always done it that way...  Most often you see this:

```
SUBROUTINE stuff(arg1,arg2)
  TYPE(*) :: arg1
  TYPE(*) :: arg2
  !
  CHARACTER(LEN=*),PARAMETER :: myName='stuff'

  !Rest of subroutine
ENDSUBROUTINE stuff
```

src/AndersonAcceleration.f90
jpjones6: Whoa, yeah that's a bug. That should fail the ci.

src/AndersonAcceleration.f90
aarograh: So the standard allows the `;` if there is only one branch.  But since you have 2 (or 3 with my proposed change to this section), each should be on its own line for consistency within the block.

src/AndersonAcceleration.f90
jpjones6: No, not for this guy. If depth <= 1 the linear solver is never inited because it isn't needed since the alphas can be calculated with a simple formula.

aarograh: Oh, gotcha.  I missed the `%LS` part.  Thought that was the Anderson object you were checking.

src/AndersonAcceleration.f90
aarograh: It will some day when Trilinos finished building.

src/AndersonAcceleration.f90
aarograh: So I'm fine with what you did on this... if it works.  If we were to build without trilinos and without mpi, though, it would compile to this:
```
SELECT TYPE(x)
ENDSELECT
```
Can you have a SELECT block with no branches?  I'm not even sure.  I'm fine leaving it this way as long as it works for a non-mpi, non-trilinos build.

src/AndersonAcceleration.f90
aarograh: So I'm fine with what you did on this... if it works. If we were to build without trilinos and without mpi, though, it would compile to this:

SELECT TYPE(x)
ENDSELECT
Can you have a SELECT block with no branches? I'm not even sure. I'm fine leaving it this way as long as it works for a non-mpi, non-trilinos build.

src/AndersonAcceleration.f90
aarograh: Also, is this something that would even be set up in XMLProc?  I feel like at most someone would simply enable anderson acceleration for a particular application, and then the code would have the appropriate values for that application.

jpjones6: Any change in the problem (different geometry, power, flow, anything) will lead to a change in what's optimal for depth and beta (especially beta). The default for beta should be 0.5 because that is safe (beta acts as a damping factor), but it is likely far from optimal for most problems. So, really the user should be aware of what Anderson is set to if they are going to use it.

src/AndersonAcceleration.f90
jpjones6: We must be looking at different parts of the code entirely. I almost can't find an example of this in Futility or MPACT.

src/AndersonAcceleration.f90
jpjones6: Apparently yes. Worked locally. Good catch though looks janky.

src/AndersonAcceleration.f90
aarograh: ok, that's fair.  But, for any of these values that could have a default, let's set a safe one here and then we can expose those to the user as needed (instead of ALWAYS requiring them to be set).  `beta` sounds like it could just default to 0.5.  It seems like `start` and `depth` could have some reasonable defaults too?  Obviously `n` has to be passed in, so that one is fine to REQUIRE the presence of.

src/AndersonAcceleration.f90
aarograh: Well, that's how the standard is written.  Much of MPACT follows this.  Futility is probably much more a mixed bag since it's so old.  But yes, the rule is that the the dummy arguments each be on their own line immediately following the function/subroutine declaration line.  Other variables come after.

src/AndersonAcceleration.f90
aarograh: Ok, cool.  If it works, it works.

src/AndersonAcceleration.f90
jpjones6: I'm good with that. I'll remove the defaults from XML_Proc on the MPACT side, but still warn for missing parameters there so the warnings are printed in the input processing. Does that sound good? Start can always be 1 (only time higher should really be chosen is when you know your initial guess is either garbage or not actually used for the vector you're accelerating such as when the initial vector is just defaulted to something). For depth 1 is safest, higher values for depth result in more aggressive acceleration that may or may not be stable or even faster.

aarograh: Yeah that sounds good.  Thanks

src/AndersonAcceleration.f90
jpjones6: Okay, taken care of.

PR #250, opened 2020-05-21
String Sorting

aarograh:
Adds equality operators for strings, string sorting,
and string duplicate removal

### DIFF
--- a/src/IO_Strings.f90
+++ b/src/IO_Strings.f90
@@ -1752,7 +1752,6 @@ SUBROUTINE removeDuplicates(list)
     IF(old_list(iList) /= old_list(iList-1)) THEN
       nUnique=nUnique+1
     ENDIF
-WRITE(*,*) iList,CHAR(old_list(iList-1)),CHAR(old_list(iList)),nUnique
   ENDDO !iList
 
   !Allocate the new list and fill with unique entries
--- a/src/IO_Strings.f90
+++ b/src/IO_Strings.f90
@@ -71,6 +71,7 @@ USE ISO_FORTRAN_ENV
 USE Futility_DBC
 USE IntrType
 USE Strings
+USE Sorting
 USE ExceptionHandler
 IMPLICIT NONE
 PRIVATE !Default private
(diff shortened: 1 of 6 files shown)


src/IO_Strings.f90
```diff
@@ -1722,4 +1724,49 @@ FUNCTION getRowStr(maxcolsize) RESULT(rowstr)
...
+  CALL MOVE_ALLOC(list,old_list)
+  CALL sort(old_list)
+
+  !Count the unique entries
+  nUnique=1
+  DO iList=2,SIZE(old_list)
+    IF(old_list(iList) /= old_list(iList-1)) THEN
+      nUnique=nUnique+1
+    ENDIF
+WRITE(*,*) iList,CHAR(old_list(iList-1)),CHAR(old_list(iList)),nUnique
```
HendersonSC, line 1755: Doh!

aarograh: Dangit, thought I got all those

aarograh: removed

src/IO_Strings.f90
```diff
@@ -1722,4 +1724,48 @@ FUNCTION getRowStr(maxcolsize) RESULT(rowstr)
...
+!>
+SUBROUTINE removeDuplicates(list)
+  TYPE(StringType),ALLOCATABLE,INTENT(INOUT) :: list(:)
+  !
+  INTEGER(SIK) :: nUnique,iList
+  TYPE(StringType),ALLOCATABLE :: old_list(:)
+
+  IF(.NOT.ALLOCATED(list)) THEN
+    ALLOCATE(list(0))
+  ENDIF
```
stimpsonsg, line 1742: would we prefer a DBC for allocation instead of just having it return?

stimpsonsg:
overall looks good

src/IO_Strings.f90
HendersonSC: I lean toward returning (not strongly), though after considering it couldn't it return if size is < 2. Since a size 1 list can't have a duplicate.

aarograh: I dunno.  I don't really have a preference.  I just went for robustness on this one

PR #253, opened 2020-05-28
Provides interfaces for converting temperatures

HendersonSC:


src/Constants_Conversion.f90
```diff
@@ -232,4 +236,130 @@ ELEMENTAL FUNCTION K_to_F(in_temp) RESULT(out_temp)
...
+  SELECTCASE(units)
+  CASE('C')
+    out_temp = C2K + in_temp
+  CASE('R')
+    out_temp = in_temp * R2K
+  CASE('F')
+    out_temp = (in_temp - 32.0_SRK)* R2K + 273.15_SRK
+  CASE DEFAULT
+    out_temp = in_temp
+  ENDSELECT
```
aarograh, line 260: So you should just call `K_to_C`, `K_to_F` and so on in all of these routines instead of duplicating the math here

src/Constants_Conversion.f90
```diff
@@ -27,6 +27,10 @@ MODULE Constants_Conversion
 PUBLIC :: C_to_F
 PUBLIC :: F_to_K
 PUBLIC :: K_to_F
+PUBLIC :: tempTo_K
+PUBLIC :: tempTo_C
+PUBLIC :: tempTo_F
+PUBLIC :: tempTo_R
```
aarograh, line 33: Do we use Rankine for anything?  I was just expecting K, C, and F

src/Constants_Conversion.f90
```diff
@@ -232,4 +236,130 @@ ELEMENTAL FUNCTION K_to_F(in_temp) RESULT(out_temp)
...
+  CASE('F')
+    out_temp = (in_temp - 32.0_SRK)* R2K + 273.15_SRK
+  CASE DEFAULT
+    out_temp = in_temp
+  ENDSELECT
+
+ENDFUNCTION tempTo_K
+!
+!-------------------------------------------------------------------------------
+!> @brief converts temperature to Kelvin from Kelvin, Rankine, or Fahrenheit,
```
aarograh, line 265: to Celsius

src/Constants_Conversion.f90
```diff
@@ -232,4 +236,130 @@ ELEMENTAL FUNCTION K_to_F(in_temp) RESULT(out_temp)
...
+  CASE('F')
+    out_temp = (in_temp - 32.0_SRK) * R2K
+  CASE DEFAULT
+    out_temp = in_temp
+  ENDSELECT
+
+ENDFUNCTION tempTo_C
+!
+!-------------------------------------------------------------------------------
+!> @brief converts temperature to Kelvin from Celsius, Rankine, or Kelvin,
```
aarograh, line 290: to Fahrenheit

src/Constants_Conversion.f90
```diff
@@ -232,4 +236,130 @@ ELEMENTAL FUNCTION K_to_F(in_temp) RESULT(out_temp)
...
+  CASE('C')
+    out_temp = (in_temp / R2K) + 32.0_SRK
+  CASE DEFAULT
+    out_temp = in_temp
+  ENDSELECT
+
+ENDFUNCTION tempTo_F
+!
+!-------------------------------------------------------------------------------
+!> @brief converts temperature to Kelvin from Celsius, Rankine, or Fahrenheit,
```
aarograh, line 315: to Rankine, from Celsius, Kelvin, or Fahrenheit

src/Constants_Conversion.f90
```diff
@@ -232,4 +236,130 @@ ELEMENTAL FUNCTION K_to_F(in_temp) RESULT(out_temp)
...
+  CASE('C')
+    out_temp = (in_temp + C2K) / R2K
+  CASE DEFAULT
+    out_temp = in_temp
+  ENDSELECT
+
+ENDFUNCTION tempTo_R
+!
+!-------------------------------------------------------------------------------
+!> @brief converts temperature to Kelvin from Celsius, Rankine, or Fahrenheit,
```
aarograh, line 340: This is just wrong

src/Constants_Conversion.f90
```diff
@@ -232,4 +236,130 @@ ELEMENTAL FUNCTION K_to_F(in_temp) RESULT(out_temp)
...
+!> if any other unit is presented in_temp is simply returned
+!> @param in_temp the temperature input as the specified units
+!> @param units character indicating the units of in_temp
+!> @return out_temp the temperature converted to Fahrenheit
+!>
+ELEMENTAL FUNCTION convertTemp(in_temp,in_units,out_units) RESULT(out_temp)
```
aarograh, line 346: Good call on this one, but need to make it public

unit_tests/testConversions/testConversions.f90
```diff
@@ -31,75 +31,166 @@ SUBROUTINE testTemperatures()
...
+  ASSERT_APPROXEQ(out_temp(4),291.85_SRK,'557.33 F_to_C (1D)')
+  !To Celsius from Rankine
+  in_temp = (/0.0_SRK,273.15_SRK,546.3_SRK,1475.01_SRK/)
+  out_temp = tempTo_C(in_temp,'R')
+  ASSERT_APPROXEQF(out_temp(1),-273.15_SRK,'0 R_to_C (1D)')
+  ASSERT_APPROXEQF(out_temp(2),-121.4_SRK,'273.15 R_to_C (1D)')
+  ASSERT_APPROXEQF(out_temp(3),30.35_SRK,'565 R_to_C (1D)')
+  ASSERT_APPROXEQF(out_temp(4),546.3_SRK,'1200 R_to_C (1D)')
+
+  COMPONENT_TEST('C_to_K')
```
aarograh, line 126: Is this different from the above "C_to_K" component test?

unit_tests/testConversions/testConversions.f90
```diff
@@ -31,75 +31,166 @@ SUBROUTINE testTemperatures()
...
+  ASSERT_APPROXEQF(out_temp(2),491.67_SRK,'491.67 K_to_R (1D)')
+  ASSERT_APPROXEQF(out_temp(3),983.34_SRK,'565 K_to_R (1D)')
+  ASSERT_APPROXEQF(out_temp(4),1475.01_SRK,'1200 K_to_R (1D)')
+  !To Rankine from Fahrenheit
+  in_temp = (/-459.67_SRK,0.0_SRK,459.67_SRK,919.34_SRK/)
+  out_temp = tempTo_R(in_temp,'F')
+  ASSERT_APPROXEQF(out_temp(1),0.0_SRK,'0 F_to_R (1D)')
+  ASSERT_APPROXEQF(out_temp(2),459.67_SRK,'491.67 F_to_R (1D)')
+  ASSERT_APPROXEQF(out_temp(3),919.34_SRK,'565 F_to_R (1D)')
+  ASSERT_APPROXEQF(out_temp(4),1379.01_SRK,'1200 F_to_R (1D)')
```
aarograh, line 193: Could be good to see some tests for `convertTemp` too

src/Constants_Conversion.f90
jpjones6: I don't believe so and I also don't believe we allow for it in inputs. All the dictionary checks in VERAIn are for F, C, and K.

stimpsonsg: as long as it's correct, don't remove it

src/Constants_Conversion.f90
stimpsonsg: are we planning to replace the `K_to_C` etc interfaces with these, though?  If so, it's just going to be temporarily duplicated.

src/Constants_Conversion.f90
stimpsonsg: yeah, this is a good one

src/Constants_Conversion.f90
aarograh: I wasn't... If you know exactly what temperature something's in already, just use the direct conversion.  This is more for annoyances like input processing where there are multiple possibilities.  That's my opinion at least.

bscollin: Do we really want to deal with string matching for these interfaces?

src/Constants_Conversion.f90
aarograh: Yeah, it's fine.  Just didn't expect that one.

src/Constants_Conversion.f90
bkochuna: I second Ben's comment on string matching. Is enumeration possible?

aarograh: It's certainly possible... we just do it in exactly 0 places in Futility or MPACT.  We need the string version at the very least, but could add enumerations (`TEMP_K`, `TEMP_C`, etc.) and provide second versions of the interfaces as well.

HendersonSC: Ah so I was expecting we wouldn't keep the K_to_C, etc. If we do plan on it its an easy switch. As for using enumerations I can't think of a single reason not to, but I also don't really see a lot of benefit either. If all cases remain a single character and the only valid inputs are capital, then the select case is effectively identical.

bscollin: I"m okay having this interface IF we keep the old ones.  Most of the time you know what the units are and you can directly use `K_to_C` (or whatever).  I just didn't want a giant loop that is doing string matching a billion times.  Another option (definitely more complicated) is to have an interface that returns the function pointer based on the desired input and output units.  Honestly, if we keep `X_to_Y` I think this is fine.

unit_tests/testConversions/testConversions.f90
HendersonSC: Yeah, that should have changed to "tempTo_K"

src/Constants_Conversion.f90
HendersonSC: Alright cleaned up the duplicate math

src/Constants_Conversion.f90
HendersonSC: fixed

src/Constants_Conversion.f90
HendersonSC: fixed

unit_tests/testConversions/testConversions.f90
HendersonSC: fixed

src/Constants_Conversion.f90
```diff
@@ -250,19 +251,19 @@ ELEMENTAL FUNCTION tempTo_K(in_temp,units) RESULT(out_temp)
 
   SELECTCASE(units)
   CASE('C')
-    out_temp = C2K + in_temp
+    out_temp = C_to_K(in_temp)
   CASE('R')
     out_temp = in_temp * R2K
   CASE('F')
-    out_temp = (in_temp - 32.0_SRK)* R2K + 273.15_SRK
+    out_temp = C_to_K(F_to_C(in_temp))
```
aarograh, line 258: Use F_to_K instead of the combination

src/Constants_Conversion.f90
```diff
@@ -300,19 +301,19 @@ ELEMENTAL FUNCTION tempTo_F(in_temp,units) RESULT(out_temp)
 
   SELECTCASE(units)
   CASE('K')
-    out_temp = (in_temp - C2K) / R2K + 32.0_SRK
+    out_temp=C_to_F(K_to_C(in_temp))
```
aarograh, line 304: Same... K_to_F

PR #254, opened 2020-05-29
Fix build errors related to Native solvers w/o PETSc

bkochuna:
CASL Ticket # - 5855

### DIFF
--- a/src/LinearSolverTypes.f90
+++ b/src/LinearSolverTypes.f90
@@ -84,6 +84,10 @@ IMPLICIT NONE
 #endif
 !petscisdef.h defines the keyword IS, and it needs to be reset
 #undef IS
+#else
+#ifdef HAVE_MPI
+#include <mpif.h>
+#endif
 #endif
 
 PRIVATE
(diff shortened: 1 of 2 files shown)


bkochuna:
Verified the build with MPI and no PETSc works.

@djabaay has been looking into some SERIAL_DEBUG test failures. I'm going to start digging into these too. If we identify the issues for that we can include them here in this PR.

aarograh:
Brendan, thanks for pushing this up.  I'll merge once the pipeline has passed.  The issues with the timeouts can go in separately.  This by itself gets an entire pipeline up and running that isn't right now.

bkochuna:
@aarograh sounds good. Thanks.

stimpsonsg:
So, this PR could go in as is (even though nothing in LinearSolvers needs this header explicitly.  I'd prefer we fold it into the new PR that's created for round 2 of the Native GMRES stuff.

bkochuna:
@stimpsonsg  Yeah, I think that makes the most sense. Matt is still getting to the bottom of things. Most recent update is:
Previously, these tests were being executed w/o a preconditioner, so some logic changed that caused these tests to run with the preconditioner. Agreement w/o the preconditioner was very good although not exact. He's tracking down the reason for this presently. The other thing we wish to investigate and resolve is if the preconditioner has problems with single pin cell cases as appeared to be evident from what was experienced here.

My best guess is that Matt should be able to get to the bottom of these issues today sometime today tomorrow.

@mattkab2 (so he sees this comment)

PR #255, opened 2020-05-29
Revert "Native GMRES"

stimpsonsg:
Reverts CASL/Futility#231


bkochuna:
@stimpsonsg Thanks Shane. was just sitting down to do this after getting the kids to bed.

PR #259, opened 2020-06-04
Hotfix for testAnderson and testVectorTypes

jpjones6:
This is a fix for the build failures in testVectorTypes in petsc and trilinos builds, and the Anderson test that fails in GCC 8.3 builds. The Anderson test failure was due to a change in the random number generator from GCC 5.4 to GCC 8.3.


unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
```diff
@@ -339,7 +340,7 @@ SUBROUTINE testStep()
   ENDDO
 
   !Esure rate of convergence is correct:
-  ASSERT_APPROXEQ(AccErr(10)/AccErr(9),0.90109849144468168_SRK,'Wrong Anderson Conv. Rate')
+  ASSERT_APPROXEQ(AccErr(10)/AccErr(9),0.90463663552029550_SRK,'Wrong Anderson Conv. Rate')
```
aarograh, line 343: Why did this change?  I thought RAND and RANDOM_NUMBER were the same in 5.4?

jpjones6: I thought that was the case, but apparently not. It could still be the same RNG under the hood in 5.4 but the seed could have been different. I changed it to force the random seed to be 0 at the beginning just to be thorough.

aarograh: Alright, yeah, definitely in favor of fixing the seed to 0.  I'm good with letting this ride.  If we happen to have issues again then I think we switch to Futility's StochasticSampler package, but I'd rather not do that if we can help it.

bscollin: whats wrong with StochasticSampler??

aarograh: Maybe nothing.  But the versions Jesse uses here should be fixed by the standard, and technically StochasticSampler could be changed by future work.  I'd just as soon have as few moving parts as possible.

bscollin: The interface is fixed, but do we know that Intel doesn't use a different seed, or a completely different sampling method?  Sampler will allow you to choose the RNG type and the seed so that you know its portable.  The standard doesn't say what the RNG methodology needs to be so I don't think you can count on the output being standardized

stimpsonsg: also, it says esure instead of ensure in the comment

HendersonSC:
Apparently late the the party, but I am pretty sure `srand` and `rand` are actually GNU extension, I am not sure we can count on them being static in anyway.

jpjones6:
They are extensions, but they are included specifically for backwards compatibility with g77. [https://gcc.gnu.org/onlinedocs/gfortran/RAND.html](url) However, the point is moot, there is no guaranteeing the Intel implementation would be identical to the GCC implementation. How about we go with the pregenerated and read from file approach.

jpjones6:
Ben pointed out offline that there have been reliability issues in the past with using files for what needs to be done here. That being the case the argument against just using the Futility native RNG doesn't hold much air. In reality if anything changes with the Futility RNG the one line in the Anderson test that checks the exact convergence rate is all that would need to change. I can even add a comment to that effect above the check to clue in future devs.

bscollin:
Aaron pointed out that I was incorrect in where the issues were.  I think we're in agreement that whichever direction is easier to implement is the right direction that will achieve a similar capability (file or internal RNG).

unit_tests/testAndersonAcceleration/testAndersonAcceleration.f90
```diff
@@ -47,6 +49,7 @@ PROGRAM testAndersonAcceleration
 
 CREATE_TEST('Test Anderson Acceleration Solver')
 
+CALL NativeRNG%init(1_SIK)
```
aarograh, line 52: One small thing: I think there are enumerations for this... should use the enumeration instead of a hard-coded integer.  Here and at line 68

aarograh: Thanks for fixing that incorrect example too

jpjones6: Yeah, no problem.

PR #260, opened 2020-06-04
Add python packages to travisci

bkochuna:
Initially testing to see if this works on travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,14 @@ before_install:
  - docker pull mbairdphys/vera_tpls
 
 script:
- - docker run -v $PWD:/mnt/Futility mbairdphys/vera_tpls /bin/bash -c "cmake --version; mkdir /build && cd /build && cmake -DFutility_ENABLE_DBC:BOOL=ON /mnt/Futility && make -j2 |& tee /mnt/Futility/make.out && ctest -j2"
+ - docker run -a STDOUT -a STDERR mbairdphys/vera_tpls /bin/bash -c "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py && pip install pandas"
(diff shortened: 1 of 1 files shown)


bkochuna:
Alright. This should do what we discussed. This will help us to merge #184

bkochuna:
@bscollin can you review? Do you want this as its own MR or should I just pull it under #184 ?

bkochuna:
@bscollin ping

@stimpsonsg or @aarograh do either of you have time to review and approve? This is holding up #184

.travis.yml
```diff
@@ -8,12 +8,14 @@ before_install:
  - docker pull mbairdphys/vera_tpls
 
 script:
- - docker run -v $PWD:/mnt/Futility mbairdphys/vera_tpls /bin/bash -c "cmake --version; mkdir /build && cd /build && cmake -DFutility_ENABLE_DBC:BOOL=ON /mnt/Futility && make -j2 |& tee /mnt/Futility/make.out && ctest -j2"
+ - docker run -a STDOUT -a STDERR mbairdphys/vera_tpls /bin/bash -c "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python get-pip.py && pip install pandas"
```
aarograh, line 11: So is this essentially installing pandas every single time we run the CI?  Is it possible instead to just update the docker image to contain pandas already?

bkochuna: Yeah this is what Ben suggested we do until Mark can update the docker container. I agree that would be the right way to do it

aarograh: Gotcha, wasn't aware of that conversation.  In that case, since the CI passed I'm good with everything here.  I'll approve and let you merge when ready.  I'd just as soon have this as it's own merge separate from #184

PR #261, opened 2020-06-05
Native Solver 2: The GMRESurrection

mattkab2:
Now with 10% more testing than the next leading brand.
Previous pull request link: https://github.com/CASL/Futility/pull/231

The problem was discovered to be related to the orthogonalization section of the GMRES solver. This has been addressed by changing the communication pattern slightly, which was easiest to do in a slightly older version of the solver.

This branch passes all tests in the MPACT checkin script I could run. There was only one Heavy test I could not run, which depends on ORIGEN. I will upload the logs from these tests soon (Friday, most likely).


aarograh:
Thanks, Matt.  I'll check this out soon, but regardless we'll probably hold the merge until the VERA release candidate has been tagged (probably next week some time).  Once that and the review are done, we can merge this in.

mattkab2:
@aarograh no problem.

I'm attaching a diff of the differences between this pull request and the previous one.
Note some changes show up as they pertain to VectorResemble_Alloc and VectorResembleAlloc; it seems someone else created these routines as well, and they were merged into Futility while I worked on fixing the merge.

Pretty much all the changes are concentrated in the solveGMRES_partial routine.
[Diff/Patch file](https://github.com/CASL/Futility/files/4736001/ns.txt)

mattkab2:
Checkin test log on the NativeGMRES branch:
[checkin_test.log](https://github.com/CASL/Futility/files/4737156/checkin_test.log)

src/LinearSolverTypes.f90
```diff
@@ -1615,228 +1662,216 @@ SUBROUTINE solveBiCGSTAB(solver)
...
-      c(it)=t/temp
-      s(it)=h/temp
-      R(it,it)=temp
-      g(it)=c(it)*phibar
-      phibar=-s(it)*phibar
-#ifdef FUTILITY_DEBUG_MSG
-      WRITE(668,*) '         GMRES-NOPC',it,ABS(phibar)
-#endif
-      IF(ABS(phibar) <= tol) EXIT
     ENDDO
```
aarograh, line 1714: The last version had a few lines right inside this ENDDO:
```
-      CALL thisLS%getResidual(residVec=u)
-      IF(PRESENT(thisPC)) CALL thisPC%apply(u)
-      thisLS%residual = BLAS_nrm2(u)
```
These deletions were intentional?  Why aren't they needed?

src/LinearSolverTypes.f90
```diff
@@ -1615,228 +1662,216 @@ SUBROUTINE solveBiCGSTAB(solver)
...
-      phibar=-s(it)*phibar
-#ifdef FUTILITY_DEBUG_MSG
-      WRITE(668,*) '         GMRES-LP',it,ABS(phibar)
+  ENDSELECT
+
+  ALLOCATE(h(thisLS%nRestart))
+  ALLOCATE(R(thisLS%nRestart+1,thisLS%nRestart+1))
+  ALLOCATE(givens_cos(thisLS%nRestart+1))
+  ALLOCATE(givens_sin(thisLS%nRestart+1))
+  ALLOCATE(g(thisLS%nRestart+1))
```
aarograh, line 1786: These 4 allocations didn't have `+1` in them last time.  Why that change?

src/LinearSolverTypes.f90
mattkab2: These were moved to the end of the solveGMRES_partial routine. Ultimately it doesn't really matter so much whether they're put here or there, since either way they get called at the end of every restart.

aarograh: Ok, works for me.  I missed it in the other location.

src/LinearSolverTypes.f90
mattkab2: This is wrapped up in the whole "older version of the solver" thing, and it's related to how the orthogonalization and given's rotation constants are stored. What it comes down to is that in the current version, you loop in such a way that you generate these constants for the basis vector i+1 (having started with the constants for basis vector 1 already initialized).
Before, you would work on basis vector i+1 while generating the constants for basis vector i.

I believe there is an argument to be made that this should be more efficient in terms of achieving the linear solver tolerance, since it effectively lets you use an additional basis vector which is wasted in the previous case. This is part of the rationale for changing the solver to this version.

aarograh: Alright, cool.  Just making sure this was an intended change.

aarograh:
And one final question for @mattkab2 : you ran all those transient cases that failed before and they passed, correct?  I see the check-in script log, but it doesn't actually list the tests that were run.

mattkab2:
@aarograh That's correct, those transient cases all work perfectly well now.

aarograh:
Alright, good.

As I mentioned, we'll leave this on a branch until the RC6 tag, but once that's tagged I'll go ahead and merge assuming no other concerns come up before then.

aarograh:
RC6 has been tagged and the sync ran and passed last night, so I think this is a good time to go ahead and merge this.

PR #263, opened 2020-06-11
Remove Call Restriction On Anderson Reset

jpjones6:
The native Anderson Acceleration ability had a restriction in place that required that a solution step be taken at least once before the solver was reset. This normally makes sense, however, for multistate cases where only a single iteration will be taken on a state this restriction becomes a problem. This MR simply removes the restriction.

### DIFF
--- a/src/AndersonAcceleration.f90
+++ b/src/AndersonAcceleration.f90
@@ -296,6 +296,9 @@ SUBROUTINE reset_AndersonAccelerationType(solver,x)
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
   CLASS(VectorType),INTENT(INOUT) :: x
 
+#ifdef HAVE_MPI | FUTILITY_HAVE_Trilinos
+  CHARACTER(LEN=*),PARAMETER :: myName='reset_AndersonAccelerationType'
+#endif
   INTEGER(SIK) :: i,j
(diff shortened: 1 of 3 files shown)


src/AndersonAcceleration.f90
```diff
@@ -296,39 +296,33 @@ SUBROUTINE reset_AndersonAccelerationType(solver,x)
   CLASS(AndersonAccelerationType),INTENT(INOUT) :: solver
   CLASS(VectorType),INTENT(INOUT) :: x
 
-  CHARACTER(LEN=*),PARAMETER :: myName='reset_AndersonAccelerationType'
```
aarograh, line 299: Let's leave the parameter in there and use it in the error messages.  That's our typical pattern (especially since it's used in multiple places), so let's stick with it.

jpjones6: Can't unfortunately, because the only uses of it are wrapped in ifdefs for tpls that the Futility ci does builds without making it look like that char variable is unused making the ci fail.

aarograh: Just wrap it in ifdefs too... that's what we do elswhere

jpjones6: Gross, alright will do.

PR #267, opened 2020-06-22
Fix 2 valgrind defects

aarograh:


stimpsonsg:
I believe @mattkab2 was going to fix the MatrixTypesParallel one...so just be aware you don't need to worry about that now, Matt

PR #273, opened 2020-07-10
Add PL 4-D Arrays

aarograh:
Adds support for 4D arrays to ParameterLists, including reading them
from HDF5 files.


src/ParameterLists.f90
```diff
@@ -488,6 +535,12 @@ MODULE ParameterLists
 PROCEDURE :: procXMLTree
 ENDTYPE ParamType
 
+!> @brief Wrapper type for an array of ParamType pointers
+TYPE :: ParamTypePtr
```
djabaay, line 539: Maybe `ParamTypePtrArray` to be consistent with the rest of the code elsewhere?

unit_tests/testParameterLists/testParameterLists.f90
```diff
@@ -3765,6 +3773,886 @@ SUBROUTINE testSDKa3()
...
+  valsnka4(2,2,1,:)=4_SNK
+  valsnka4(1,1,2,:)=5_SNK
+  valsnka4(2,1,2,:)=6_SNK
+  valsnka4(1,2,2,:)=7_SNK
+  valsnka4(2,2,2,:)=8_SNK
+  !test init
+
+  COMPONENT_TEST('%init(...)')
+  ASSERT(.NOT.ASSOCIATED(testParam%pdat),'%pdat 1')
+  ASSERT(LEN(testParam%name) == 0,'%name 1')
```
djabaay, line 3793: Use `ASSERT_EQ`?

aarograh: Yeah... once I figure out what's up with the build error I plan to just search/replace as many of these ASSERT to ASSERT_EQ/ASSERT_APPROXEQ as I can

aarograh: I've decided not to do this right now... there are a LOT of things that can't be passed through the ASSERT_EQ interfaces because they're using an == operator that we wrote that isn't recognized during the preprocessing step.  I'm not clear on why it matters, but it apparently does. And with the unit test being nearly 7000 lines long I don't have time to actually look at every single replacement I'm doing.

PR #277, opened 2020-07-23
Hotfix test times

aarograh:
Differences in time keeping on virtual machines and the window subsystem for linux can be problematic for the current testTimes.  This reworks the test to simply check that the elapsed time is increasing properly and not to anything insane, rather than relying on more precise amounts of time.

### DIFF
--- a/unit_tests/testTimes/testTimes.f90
+++ b/unit_tests/testTimes/testTimes.f90
@@ -22,15 +22,10 @@ CHARACTER(LEN=5) :: adum3
 CHARACTER(LEN=2) :: adum4
 CHARACTER(LEN=MAXLEN_DATE_STRING) :: adate
 CHARACTER(LEN=MAXLEN_CLOCK_STRING) :: aclock
+REAL(SRK) :: totalElapsed
 !
 !Check the timer resolution
 CREATE_TEST('TIMERS')
-!
-!Print module constants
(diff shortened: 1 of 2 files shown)


unit_tests/testTimes/testTimes.f90
```diff
@@ -113,10 +105,23 @@ PROGRAM testTimes
...
 CALL sleep(1)
 CALL testTimer%toc()
-ASSERT(ABS(testTimer%elapsedtime-1._SRK) < 0.05_SRK,'tic/toc')
-FINFO() 'testTimer%toc()= ',testTimer%elapsedtime
+totalElapsed=testTimer%elapsedtime
+ASSERT(totalElapsed > 0.0_SRK,'tic/toc')
+CALL testTimer%tic()
+CALL sleep(1)
+CALL testTimer%toc()
+ASSERT(testTimer%elapsedtime > totalElapsed,'tic/toc')
```
HendersonSC, line 113: Nitpicky, but given its a small file and we are trying to move this direction, `ASSERT_GT` and others. Afterwhich `FINFO` isn't required.

aarograh: Nah, good call.  I forgot those particular flavors existed

aarograh: Fixed

PR #278, opened 2020-07-24
Adds Least Common Multiple Routine

HendersonSC:
Adds two least common multiple routines, one for a pair of scalar integers and one for an array of intergers


src/ExtendedMath.f90
```diff
@@ -165,6 +170,56 @@ ELEMENTAL FUNCTION GreatestCommonDivisor(a1,a2) RESULT(b)
...
+  lcm = abs(u(1))
+  IF(ALL(u == u(1))) THEN
+    RETURN
+  ENDIF
+
+  DO i=2,SIZE(u)
+    a = abs(u(i))
+    !In all likelihood there exists an 'a' and 'b' which will cause overflow
+    lcm = (lcm/GreatestCommonDivisor(lcm,a)) * a
+  ENDDO
```
aarograh, line 196: So... seems like this entire function can be reduced to this and reduce some redundant logic.  Disagree?
```
lcm = 0
IF(SIZE(u) == 0) RETURN
IF(ANY(u) == 0) RETURN

lcm=a
DO i=2,SIZE(u)
  lcm = LeastCommonMultiple_scalar(ABS(lcm),ABS(u(i))
ENDDO
```

src/VTUFiles.f90
```diff
@@ -465,7 +465,7 @@ SUBROUTINE writepvtu_VTUXMLFileType(fileobj,funit,case,filen,procs,rank)
   INTEGER(SIK) :: i,j,iord
   CHARACTER(LEN=128) :: fname,fmtStr
   !
-  IF((procs>1).AND.rank == 0) THEN
+  IF(rank == 0) THEN
```
aarograh, line 468: Didn't we already merge this?  Or am I crazy?

unit_tests/testExtendedMath/testExtendedMath.f90
```diff
@@ -81,67 +76,66 @@ SUBROUTINE testRatFrac()
...
+  ASSERT_EQ(LeastCommonMultiple(5,0),0, 'LCM')
+
+  ASSERT_EQ(LeastCommonMultiple(2,3),6, 'LCM')
+  ASSERT_EQ(LeastCommonMultiple(3,2),6, 'LCM')
+  ASSERT_EQ(LeastCommonMultiple(201,3),201, 'LCM')
+  ASSERT_EQ(LeastCommonMultiple(33,198),198, 'LCM')
+  ASSERT_EQ(LeastCommonMultiple(8,12),24, 'LCM')
+  ASSERT_EQ(LeastCommonMultiple(8,36),72, 'LCM')
+
+  ASSERT_EQ(LeastCommonMultiple((/8,12,36/)),LeastCommonMultiple(8,36),"LCM array")
```
aarograh, line 115: Might be good to have a couple tests with longer arrays and things in non-ascending orders and such just to make sure nothing crazy happens.  But overall the test is in pretty decent shape

src/ExtendedMath.f90
```diff
@@ -165,6 +170,56 @@ ELEMENTAL FUNCTION GreatestCommonDivisor(a1,a2) RESULT(b)
...
+
+  INTEGER(SIK) :: lcm
+  INTEGER(SIK) :: a,b
+
+  lcm = 0
+  IF(u == 0 .OR. v == 0) RETURN
+  a = ABS(u)
+  b = ABS(v)
+
+  !In all likelihood there exists an 'a' and 'b' which will cause overflow
```
aarograh, line 218:
```
a=HUGE(1_SIK)
b=HUGE(1_SIK)-1
```
;)

src/VTUFiles.f90
HendersonSC: We did, it might be that this was the branch I pushed it up on.....but I am surprised it is being shown as a difference

src/ExtendedMath.f90
HendersonSC: Yeah, implemented this less the call to `ABS`. Since its calling the scalar version and the scalar version is agnostic to sign by design, I don't think its necessary.

aarograh: Good point.  Thanks for cleaning up

unit_tests/testExtendedMath/testExtendedMath.f90
HendersonSC: Added a few more plus a couple more "wonky" cases.

src/ExtendedMath.f90
HendersonSC: lol, yes that is one. Now just to identify how ever many more and we can guard against it.

src/VTUFiles.f90
HendersonSC: Something weird with not having merged master I guess. It seems to recognize that isn't a change now

PR #279, opened 2020-07-29
Adding array routines to round reals and query for remainders.

djabaay:
Description:
These routines are needed for checking input value tolerances during the
input processing.

CASL Ticket # - PHI 5989


djabaay:
@aarograh I am not using the `applyTol` routine and the issue does not involve `.APPROXEQA.`.  I am just using the `hasAnyRemainder` function for warnings in MPACT, which is a derivative of the `applyTol`.  I left it in the PR, but can take it out if need be.  If you want me to try to use integer math instead, that's fine.  I agree, I recognized the 1e-20 limit and I add 1e-14 to values to ensure consistent behavior across different machines.  It may be necessary to change the DBC tolerance to 1e-14 instead.

The MPACT MR that I have right now is here which shows the issue more specifically: https://ners-arc-05.engin.umich.edu/MPACT/MPACT/-/merge_requests/312

aarograh:
Alright, so this is essentially supported warning messages related to 5989 then.  I'm ok with that for now.

But I would still like to see either 1e-20 upped to 1e-14, or some test cases showing that tolerances below 1e-20 work correctly.

djabaay:
Will do on the DBC. Also, Do you want me to just take out the `applyTol` routines?  Or leave them in and retrofit them with integer arithmatic?

aarograh:
If `applyTol` isn't being used I'd just as soon see it taken out.  I think there are some potential issues with rounding to a particular tolerance.

PR #281, opened 2020-08-21
Adding subroutines for Find interface for searching strings.

djabaay:
Description:
Fixing another issue with array utils assigning the first value as
unique, even if it is null.  Changing that to loop over all values
instead.

CASL Ticket # - PHI 6419


aarograh:
It looks like a Travis build ran successfully on this and the status juts isn't making it back.  I'm going to go ahead and merge.

PR #282, opened 2020-10-09
Add replace_entry interface to ArrayUtils

aarograh:
This allows a single entry of an array to be replaced by an array.  It's basically a convenience function to handle certain types of array resize operations.

### DIFF
--- a/src/ArrayUtils.f90
+++ b/src/ArrayUtils.f90
@@ -35,6 +35,7 @@ PUBLIC :: isMonotonic
 PUBLIC :: isIncreasing
 PUBLIC :: isDecreasing
 PUBLIC :: hasAnyRemainder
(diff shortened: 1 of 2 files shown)


HendersonSC (approved):
Looks to be in good shape.

djabaay:
Just curious @aarograh, where did you find yourself seeing this operation used a bunch?

aarograh:
@HendersonSC I'm going to push up a unit test in a bit before we merge this

@djabaay I reworked editsFactory in MPACT and there are several spots I use this there to expand the user-defined edits groups

HendersonSC:
Yeah good point, this is why I shouldn't review thing later in the night.

HendersonSC:
Unit test is definitely a good add. I am good with this as is.

PR #283, opened 2020-10-09
Add constructor for string types

aarograh:
Adds a constructor for string types.  This allows interfaces to be defined
only taking in string types, and the client code can simply cast native characters
to the string type in the call to the routine if needed.  Specifically this can
be useful for testing where hard-coded strings will be common.

### DIFF
--- a/src/Strings.f90
+++ b/src/Strings.f90
@@ -134,6 +134,12 @@ TYPE :: StringType
     !FINAL :: clean_str
 ENDTYPE StringType
 
+INTERFACE StringType
+  !> @copybrief Strings::charToStringType
+  !> @copydetails Strings::charToStringType
+  MODULE PROCEDURE charToStringType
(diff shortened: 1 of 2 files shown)


HendersonSC (approved):
Overall I don't think I have any objections to this. It would be nice if Fortran allowed something like `list = StringType((/'a','aa','aaa'/))` for this, but I expect it would puke on those being differently sized character arrays. I am guessing the same could be accomplished as `list =(/StringType('a'),StringType('aa'),StringType('aaa')/)`, but that is just awkward. Though it would be more awkward to pad each character array with whitespace to make them the same length and have to `TRIM` everywhere.

Not really sure that has to much to do with this. I just wanted to document my thoughts on using StringTypes. Principally that when you make a StringType it should really be the string you want to work with and not some whitespace padded version of it.

PR #284, opened 2020-10-12
Adds pop_entry interface to ArrayUtils

aarograh:
Removes an entry from an array

### DIFF
--- a/src/ArrayUtils.f90
+++ b/src/ArrayUtils.f90
@@ -38,6 +38,7 @@ PUBLIC :: isDecreasing
 PUBLIC :: hasAnyRemainder
 PUBLIC :: replaceEntry
 PUBLIC :: popEntry
+PUBLIC :: insertEntry
 !PUBLIC :: findIntersection
 !Need a routine in here that compares a 1-D array to a 2-D array for a given dimension
 !to see if the 1-D array exists in the 2-D array...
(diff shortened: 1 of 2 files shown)


src/ArrayUtils.f90
```diff
@@ -1372,4 +1393,89 @@ SUBROUTINE replaceEntry_1DString(oldlist,sublist,entry)
...
+!>
+!> After insertion, the @c entry-th index of the list will be the first element
+!> of @c insert, with the original values at @c entry and later being moved back
+!> by one.  If @c entry is given a value of @c SIZE(list)+1, then @c insert is
+!> added to the back of the list
+!>
+SUBROUTINE insertEntry_1DString_1Dentry(list,insert,entry)
+  TYPE(StringType),ALLOCATABLE,INTENT(INOUT) :: list(:)
+  TYPE(StringType),INTENT(IN) :: insert(:)
+  INTEGER(SIK),INTENT(IN) :: entry
```
HendersonSC, line 1465: Just so we can be clear on the expected behavior, what happens if `insert` is size 0? I personally believe it should go through this interface and return the original list, but there are a handful of assumptions in that.

HendersonSC (approved):

src/ArrayUtils.f90
aarograh: Talked offline, but for the sake of completeness: that's exactly what happens.  Updated unit tests show that.

PR #285, opened 2020-10-14
Fix getUnique to always return allocated array

aarograh:
Ensures that an allocated array is always returned by getUnique

### DIFF
--- a/src/ArrayUtils.f90
+++ b/src/ArrayUtils.f90
@@ -62,8 +62,7 @@ INTERFACE getDelta
   MODULE PROCEDURE getDelta_1DReal
 ENDINTERFACE getDelta
 
-!> @brief Generic interface to ...
-!>
+!> @brief Generic interface to retrieve the unique entries of an array
 INTERFACE getUnique
   !> @copybrief ArrayUtils::getUnique_1DReal
   !> @copydetails ArrayUtils::getUnique_1DReal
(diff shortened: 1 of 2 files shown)


aarograh:
@djabaay this is related to https://code.ornl.gov/vera/vera-dev/-/issues/4634 that I just tagged you in.  Mind doing a quick review?  It's just a quick change, but I think ensuring the return value is always allocated, even if to size 0, is a little safer than before.

djabaay:
I guess if you want the routine to be able to accept an unallocated array to execute on, unallocated or allocated to size 0 shouldn't really matter.  Is the goal to just ensure that _something_ is allocated, regardless of the input?

I haven't got my ucams unlocked so I haven't seen the ticket yet.

aarograh:
Yeah I think it's better to return something allocated no matter what.  To say there are 0 unique entries in a size-0 or unallocated array seems like a reasonable statement to me.

The issue this is aiming to address is a crash in BWR jumpin cases that Mehdi is dealing with.

djabaay:
Understood, thanks for the brief explanation.

src/ArrayUtils.f90
```diff
@@ -524,6 +524,8 @@ SUBROUTINE getUnique_1DReal(r,rout,delta,tol)
 
     !Deallocate
     DEALLOCATE(tmpr)
+  ELSE
+    ALLOCATE(rout(0))
```
djabaay, line 528: The only thing I can think of is add a comment in the subroutine header to describe when there are 0 unique entries.  Other than comments, looks good.

djabaay (changes requested):

src/ArrayUtils.f90
aarograh: good call.  Just pushed updates

djabaay (approved):
Thanks @aarograh

PR #287, opened 2020-10-23
Add 2 deallocate statements that shouldn't be necessary

aarograh:


HendersonSC (approved):
If someone were to add 2 deallocate statements that shouldn't be necessary, then this is the way it should be done.

PR #289, opened 2020-10-29
Add gather collective for StringTypes

HendersonSC:
Adds a wrapper in parallel env which works similarly to mpi_gather for futility string types..

### DIFF
--- a/src/ParallelEnv.f90
+++ b/src/ParallelEnv.f90
@@ -527,7 +527,6 @@ ENDSUBROUTINE partition_indices_ParEnvType
 !> 0-nremainder
 !>
 SUBROUTINE partition_greedy_ParEnvType(myPE,iwgt,n1,n2,ipart,idxmap)
-  CHARACTER(LEN=*),PARAMETER :: myName='partition_greedy_ParEnvType'
   CLASS(ParEnvType),INTENT(IN) :: myPE
   INTEGER(SIK),INTENT(IN) :: iwgt(:)
   INTEGER(SIK),INTENT(IN),OPTIONAL :: n1
--- a/src/ParallelEnv.f90
+++ b/src/ParallelEnv.f90
@@ -196,20 +196,19 @@ TYPE,EXTENDS(ParEnvType) :: MPI_EnvType
     !> @copybrief ParallelEnv::gather_SLK1_MPI_Env_type
     !> @copydetails ParallelEnv::gather_SLK1_MPI_Env_type
     PROCEDURE,PASS,PRIVATE :: gather_SLK1_MPI_Env_type
+    !> @copybrief ParallelEnv::gather_str1D_MPI_ENV_type
+    !> @copydetails ParallelEnv::gather_str1D_MPI_ENV_type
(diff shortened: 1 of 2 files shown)


unit_tests/testParallelEnv/testParallelEnv.f90
```diff
@@ -350,6 +342,34 @@ SUBROUTINE testMPIEnv()
...
+      ALLOCATE(tstRecv(2))
+      CALL testMPI%gatherV(tstString,0)
+      ASSERT_EQ(CHAR(tstString(1)),"","test gather at master, 1")
+      ASSERT_EQ(CHAR(tstString(2)),"blah_not_master","test gather at master, 2")
+
+      CALL testMPI%gatherV(tstString,1)
+      ASSERT_EQ(CHAR(tstString(1)),"blah_master","test gather at master, 1")
+      ASSERT_EQ(CHAR(tstString(2)),"blah_not_master","test gather at master, 2")
+    ENDIF
+  ENDIF
```
aarograh, line 371: Wouldn't be terrible to also test the `nproc == 1` case as well.

aarograh: I would also like to see a test for the rank-2 array as well

src/ParallelEnv.f90
```diff
@@ -200,6 +201,16 @@ SUBROUTINE ParEnvType_clear_absintfc(myPE)
...
         gather_SIK1_MPI_Env_type
+
+    !> @copybrief ParallelEnv::gatherV_str1D_MPI_ENV_type
+    !> @copydetails ParallelEnv::gatherV_str1D_MPI_ENV_type
+    PROCEDURE,PASS,PRIVATE :: gatherV_str1D_MPI_ENV_type
+    !> @copybrief ParallelEnv::gatherV_str2D_MPI_ENV_type
+    !> @copydetails ParallelEnv::gatherV_str2D_MPI_ENV_type
+    PROCEDURE,PASS,PRIVATE :: gatherV_str2D_MPI_ENV_type
+    !>
+    GENERIC :: gatherV => gatherV_str1D_MPI_ENV_type,gatherV_str2D_MPI_ENV_type
```
aarograh, line 212: Couldn't get this under the `gather` generic?

aarograh (changes requested):
Looks good overall.  Mainly would like to see a tad more testing.

Thanks for cleaning up some of the other doxygen comments too

src/ParallelEnv.f90
HendersonSC: I started with the same generic, but hit a snag (probably of my own making) with the arguments in the interface. Not sure which way around gather/gatherv makes more sense in the actual name scheme.

aarograh: Hmm... ok... do you remember the error?  If not, could you try it again and let me know what it was?  I've run into errors with `GENERIC` when we used deferred shape arrays (`REAL(SRK),DIMENSION(*) :: ...`), but I feel like the current `gather` procedures should work fine.

aarograh (approved):

PR #290, opened 2020-10-30
Relaxed nnz requirement on serial banded matrices

mattkab2:
The serial banded matrix requires an exact nnz value count to function. However, this was improved for the parallel versions, which instead treat nnz as an upper bound.

This MR updates the serial behavior to match the parallel behavior, which resolves several ongoing problems with implementing native CMFD in the serial debug test cases.

### DIFF
--- a/src/MatrixTypes_Native.f90
+++ b/src/MatrixTypes_Native.f90
@@ -695,6 +695,7 @@ SUBROUTINE assemble_BandedMatrixType(thisMatrix)
 
   ALLOCATE(diagRank(thisMatrix%counter))
   ALLOCATE(idxOrig(thisMatrix%counter))
+  IF (thisMatrix%isAssembled) RETURN
   DO i=1,thisMatrix%counter
     iLong=INT(thisMatrix%iTmp(i),kind=SLK)
     jLong=INT(thisMatrix%jTmp(i),kind=SLK)
@@ -1369,6 +1370,7 @@ SUBROUTINE set_BandedMatrixType(matrix,i,j,setval)
   IF((j <= matrix%m) .AND. (i <= matrix%n) .AND. (i >= 1) .AND. (j >= 1)) THEN
     IF(.NOT. matrix%isAssembled) THEN
       ! If it is not assembled, add to tmp variables
+      REQUIRE(matrix%counter < matrix%nnz)
       matrix%counter=matrix%counter+1
       matrix%iTmp(matrix%counter)=i
       matrix%jTmp(matrix%counter)=j
--- a/src/MatrixTypes_Native.f90
+++ b/src/MatrixTypes_Native.f90
@@ -444,7 +444,8 @@ SUBROUTINE init_BandedMatrixParam(matrix,Params)
 
   ! Pull Data From Parameter List
   CALL validParams%get('MatrixType->n',n)
-  CALL validParams%get('MatrixType->m',m)
+  m=n
+  IF (validParams%has('MatrixType->m')) CALL validParams%get('MatrixType->m',m)
   CALL validParams%get('MatrixType->nnz',nnz)
   CALL validParams%clear()
 
@@ -691,11 +692,10 @@ SUBROUTINE assemble_BandedMatrixType(thisMatrix)
   INTEGER(SLK),ALLOCATABLE :: diagRank(:)
 
   REQUIRE(thisMatrix%isInit)
-  REQUIRE(thisMatrix%nnz == thisMatrix%counter)
 
-  ALLOCATE(diagRank(SIZE(thisMatrix%iTmp)))
-  ALLOCATE(idxOrig(SIZE(thisMatrix%iTmp)))
-  DO i=1,SIZE(thisMatrix%iTmp)
+  ALLOCATE(diagRank(thisMatrix%counter))
+  ALLOCATE(idxOrig(thisMatrix%counter))
+  DO i=1,thisMatrix%counter
     iLong=INT(thisMatrix%iTmp(i),kind=SLK)
     jLong=INT(thisMatrix%jTmp(i),kind=SLK)
     nLong=INT(thisMatrix%n,kind=SLK)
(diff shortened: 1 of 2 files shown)


src/MatrixTypes_Native.f90
```diff
@@ -691,11 +692,10 @@ SUBROUTINE assemble_BandedMatrixType(thisMatrix)
   INTEGER(SLK),ALLOCATABLE :: diagRank(:)
 
   REQUIRE(thisMatrix%isInit)
-  REQUIRE(thisMatrix%nnz == thisMatrix%counter)
```
djabaay, line 694: Would the requirement be for `thisMatrix%nnz >= thisMatrix%counter` ?

src/MatrixTypes_Native.f90
```diff
@@ -691,11 +692,10 @@ SUBROUTINE assemble_BandedMatrixType(thisMatrix)
...
 
-  ALLOCATE(diagRank(SIZE(thisMatrix%iTmp)))
-  ALLOCATE(idxOrig(SIZE(thisMatrix%iTmp)))
-  DO i=1,SIZE(thisMatrix%iTmp)
+  ALLOCATE(diagRank(thisMatrix%counter))
+  ALLOCATE(idxOrig(thisMatrix%counter))
+  DO i=1,thisMatrix%counter
```
djabaay, line 698: What's the significance of changing from `thisMatrix%iTmp` to `thisMatrix%counter`?

mattkab2: Essentially this means it will loop over the added entries instead of the full iTmp array.
iTmp is an allocated array which stores up to nnz entries, but now because of these changes it may not be full.
So we loop up to counter, which counts the number of entries added prior to the assemble call (which makes the matrix structure read-only)

src/MatrixTypes_Native.f90
mattkab2: Correct, all though this requirement is better placed in the set routine, because otherwise an array access error will happen before the requirement is called (counter is equal to the index of the last set element in iTmp).

I'll make the change and push up.

src/MatrixTypes_Native.f90
djabaay: Got it, so `%counter` is the number of entries set on the `%iTmp` array.

unit_tests/testMatrixTypes/testMatrixTypes.f90
```diff
@@ -1416,6 +1441,28 @@ SUBROUTINE testMatrix()
     bool = bool .AND. thisMatrix%bands(1)%elem(1) == 9
     ASSERT(bool, 'banded%set(...)')
   ENDSELECT
+  WRITE(*,*) "Finished set B"
```
djabaay, line 1444: Delete the debug statement

unit_tests/testMatrixTypes/testMatrixTypes.f90
```diff
@@ -1402,6 +1402,31 @@ SUBROUTINE testMatrix()
...
+    bool = .TRUE.
+    bool = bool .AND. thisMatrix%bands(1)%elem(1) == 1
+    bool = bool .AND. thisMatrix%bands(1)%elem(2) == 3
+    bool = bool .AND. thisMatrix%bands(1)%elem(3) == 5
+    bool = bool .AND. thisMatrix%bands(1)%elem(4) == 7
+    bool = bool .AND. thisMatrix%bands(2)%elem(1) == 2
+    bool = bool .AND. thisMatrix%bands(2)%elem(2) == 6
+    ASSERT(bool, 'banded%set(...)')
+  ENDSELECT
+  WRITE(*,*) "Finished set A"
```
djabaay, line 1419: Delete the debug statement

src/MatrixTypes_Native.f90
```diff
@@ -1369,6 +1370,7 @@ SUBROUTINE set_BandedMatrixType(matrix,i,j,setval)
   IF((j <= matrix%m) .AND. (i <= matrix%n) .AND. (i >= 1) .AND. (j >= 1)) THEN
     IF(.NOT. matrix%isAssembled) THEN
       ! If it is not assembled, add to tmp variables
+      REQUIRE(matrix%counter < matrix%nnz)
```
djabaay, line 1373: `<` or `<=` ?

mattkab2: It needs to be < because counter is incremented before the set call. Otherwise counter would equal the number of entries plus one, which isn't a disaster but I chose to set it up this way.

djabaay: Perhaps just make a note of that somewhere if you haven't already.

unit_tests/testMatrixTypes/testMatrixTypes.f90
mattkab2: Removed

unit_tests/testMatrixTypes/testMatrixTypes.f90
mattkab2: Removed

djabaay:
Quoting @mattkab2 from our offline conversation in regards to changing the DBC from `<=` to `<`:

I guess it would be more accurate to say nnz is the max nonzeros the matrix can hold.
The counter starts at zero when initialized
11:21
Then when set is called, it is incremented and iTmp(counter) gets set
11:21
But iTmp is allocated to have nnz entries, so counter can never be equal to nnz at the start of the set call or there will be a memory access error
11:22
At least, before the assemble routine is called.
After assemble is called it is assumed that you're modifying an existing entry

This fully explains what is going on, and makes sense to me.

@aarograh Do you want to take a quick look through before merging?

djabaay (approved):

mattkab2:
@aarograh Mind taking a quick look?

aarograh:
@mattkab2 I don't see any issues that need to be addressed with these changes.  We're tagging VERA 4.3-RC1 soon, so I'm going to hold off merging this until after the tag just to be on the safe side.  I'll merge as soon as I know we've tagged.

aarograh:
Sounds like things are going a bit slower with the tag than I realized, so I'm going ahead with this.  We have time to back it out if something were to go wrong.

PR #292, opened 2020-11-24
Simplify string conversions

HendersonSC:
Simplify some of the operations in the string to integer and string to float routines and add testing.

### DIFF
--- a/src/Strings.f90
+++ b/src/Strings.f90
@@ -1254,18 +1254,29 @@ FUNCTION str_to_sik(this,stt,stp) RESULT(i)
   INTEGER(SIK),INTENT(IN),OPTIONAL :: stp
   INTEGER(SIK) :: i
   !
-  CHARACTER(LEN=:),ALLOCATABLE :: scratch
+  INTEGER(SIK) :: begStr,endStr,ioerr
 
-  IF(PRESENT(stt) .OR. PRESENT(stp)) THEN
-    scratch = this%s(stt:stp)
-  ELSE
-    scratch = this%s
+  !Bypass the world if the string isn't allocated
+  ioerr = -HUGE(ioerr)
+  IF(ALLOCATED(this%s)) THEN
+    !Find the beginning
+    IF(PRESENT(stt)) THEN
+      begStr = stt
+    ELSE
+      begStr = 1
+    ENDIF
+    !Find the end
+    IF(PRESENT(stp)) THEN
+      endStr = stp
+    ELSE
+      endStr = LEN(this%s)
+    ENDIF
+    READ(this%s(begStr:endStr),*,IOSTAT=ioerr) i
   ENDIF
 
-  IF(LEN(scratch) > 0) THEN
-    READ(scratch,*) i
-  ELSE
-    i = -HUGE(i)
+  !Test to see if it worked, return 0 on any error code
+  IF(ioerr /= 0) THEN
+    i = 0_SIK
   ENDIF
 ENDFUNCTION str_to_sik
 !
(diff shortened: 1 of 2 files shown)


aarograh (changes requested):

src/Strings.f90
```diff
@@ -1254,18 +1254,29 @@ FUNCTION str_to_sik(this,stt,stp) RESULT(i)
...
+    ENDIF
+    READ(this%s(begStr:endStr),*,IOSTAT=ioerr) i
   ENDIF
 
-  IF(LEN(scratch) > 0) THEN
-    READ(scratch,*) i
-  ELSE
-    i = -HUGE(i)
+  !Test to see if it worked, return 0 on any error code
+  IF(ioerr /= 0) THEN
```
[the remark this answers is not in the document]
HendersonSC: So ioerr would be the normal error codes returned by the READ statement. Which I only returns 0 if there was no error. I am not convinced what codes GCC returns when it encounters an error, but I also am not really convinced it matters in this case, it could matter if we were trying to anticipate some common errors, but that feels scary.

I also don't really care for returning zero, I just wasn't really sure what other number to pick. Too bad there isn't an integer NaN that would be ideal. I suppose `-HUGE(i)` is probably the the best bu it is also valid input albeit not one we should really use often.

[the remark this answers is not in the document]
aarograh: Yeah... I would prefer `NaN`, then `-Infinity`, and then `-HUGE(i)`... I'm guessing neither of the first 2 exist for integers, but `-HUGE(i)` is way less likely to cause a problem than `0`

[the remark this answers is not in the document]
HendersonSC: fixed

src/Strings.f90
```diff
@@ -1283,18 +1294,32 @@ FUNCTION str_to_srk(this,stt,stp) RESULT(i)
   REAL(SRK) :: i
```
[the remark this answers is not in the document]
HendersonSC: fixed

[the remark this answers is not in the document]
HendersonSC: fixed

unit_tests/testStrings/testStrings.f90
```diff
@@ -742,9 +742,9 @@ SUBROUTINE testStrFunct()
   str = '-0.2468e-000'
   ASSERT(str%isNumeric(),'with e-000')
   str = '-.2468e-000'
-  ASSERT(.NOT.str%isNumeric(),'no leading digit')
+  ASSERT(str%isNumeric(),'no leading digit')
   str = '-0.2468-000'
-  ASSERT(.NOT.str%isNumeric(),'mixing exponent symbol')
+  ASSERT(str%isNumeric(),'mixing exponent symbol')
```
aarograh, line 747: Is this actually correct?  The missing `e` makes that not really a single numeric... unless we're supposed to catch artihmetic expressions too...

unit_tests/testStrings/testStrings.f90
```diff
@@ -698,9 +698,9 @@ SUBROUTINE testStrFunct()
   str = '-0.2468e-000'
   ASSERT(str%isFloat(),'with e-000')
   str = '-.2468e-000'
-  ASSERT(.NOT.str%isFloat(),'no leading digit')
+  ASSERT(str%isFloat(),'no leading digit')
   str = '-0.2468-000'
-  ASSERT(.NOT.str%isFloat(),'mixing exponent symbol')
+  ASSERT(str%isFloat(),'missiing exponent symbol')
```
aarograh, line 703: Same question here as below... not really sure this is correct...

Also, `missiing` is spelled wrong

aarograh (commented):
Comments were resolved, but I hvae one additional question now about the correctness of part of a unit test

unit_tests/testStrings/testStrings.f90
HendersonSC: So I think this `10.7.2.3.2 F editing` in the standard applies 
> The input field is either an IEEE exceptional specification or consists of an optional sign, followed by a string of
 one or more digits optionally containing a decimal symbol, including any blanks interpreted as zeros. The d has
 no effect on input if the input field contains a decimal symbol. If the decimal symbol is omitted, the rightmost
 d digits of the string, with leading zeros assumed if necessary, are interpreted as the fractional part of the value
 represented. The string of digits may contain more digits than a processor uses to approximate the value. The
 basic form may be followed by an exponent of one of the following forms:
   • a sign followed by a digit-string;
   • the letter E followed by zero or more blanks, followed by a signed-digit-string;
   • the letter D followed by zero or more blanks, followed by a signed-digit-string.

If I am reading that correctly then
```
1234.567e89
1234.567E89
1234.567d89
1234.567D89
1234.567+89
```
should all be the same. So would adding a `+` after any of `e, E, d, or D` and likewise the negative exponent versions.

unit_tests/testStrings/testStrings.f90
HendersonSC: same answer here......except for missing has more than enough i's.

unit_tests/testStrings/testStrings.f90
aarograh: Well... that's bizarre... I would never have guessed that was a permissible way of defining a float.  Kind of wish it weren't, but it doesn't prove your test is correct (and that the previous implementation was actually deficient).

unit_tests/testStrings/testStrings.f90
HendersonSC: removed missiing's extra i

aarograh (approved):

aarograh:
Going to hold this for the VERA 4.3 RC1 tag, just in case (unless it gets delayed)

PR #294, opened 2020-12-07
Metric conversions

bscollin:


src/Constants_Conversion.f90
```diff
@@ -75,27 +75,121 @@ MODULE Constants_Conversion
...
+REAL(SRK),PUBLIC,PARAMETER :: kg2g=kilo2base
 !> Conversion factor from g to kg
-REAL(SRK),PUBLIC,PARAMETER :: g2kg=ONE/kg2g
+REAL(SRK),PUBLIC,PARAMETER :: g2kg=base2kilo
 !> Conversion factor from g to mg
-REAL(SRK),PUBLIC,PARAMETER :: g2mg=1000.0_SRK
+REAL(SRK),PUBLIC,PARAMETER :: g2mg=base2milli
 !> Conversion factor from mg to g
-REAL(SRK),PUBLIC,PARAMETER :: mg2g=ONE/g2mg
+REAL(SRK),PUBLIC,PARAMETER :: mg2g=milli2base
```
aarograh, line 192: One of these days (maybe as we wrap up the BWR STH stuff) I'll switch downstream uses over to the new ones and get rid of these redundant conversions.  Same for the length conversions further down.

aarograh (approved):

PR #295, opened 2020-12-10
Fix potential for bad ranks to be used in string gathers

aarograh:
In some cases, HUGE(...) was not being overwritten if there
were empty strings, which could then lead to errors in the
send/receive.  This was fixed by using -HUGE(...) instead,
and then only performing the send/receive for rank >= 0

### DIFF
--- a/src/ParallelEnv.f90
+++ b/src/ParallelEnv.f90
@@ -1043,7 +1043,7 @@ SUBROUTINE gather_str2D_MPI_ENV_type(myPE,sendbuf,root)
 
   !Need to know which process is responsible for which string
   ALLOCATE(charProc(UBOUND(sendbuf,DIM=1),UBOUND(sendbuf,DIM=2)))
-  charProc = HUGE(rank)
+  charProc = -HUGE(rank)
   iEntry = 0
   DO j=1,UBOUND(charProc,DIM=2)
     DO i=1,UBOUND(charProc,DIM=1)
@@ -1052,7 +1052,7 @@ SUBROUTINE gather_str2D_MPI_ENV_type(myPE,sendbuf,root)
     ENDDO
   ENDDO
   !This is somewhat arbitrary, but giving preference to lowest rank's data
-  CALL myPE%allReduceMinI(SIZE(charProc),charProc)
+  CALL myPE%allReduceMaxI(SIZE(charProc),charProc)
 
   !Set-up individual send and receive for each string
   IF(myPE%rank /= rank) THEN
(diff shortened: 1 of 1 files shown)


src/ParallelEnv.f90
```diff
@@ -1052,7 +1052,7 @@ SUBROUTINE gather_str2D_MPI_ENV_type(myPE,sendbuf,root)
     ENDDO
   ENDDO
   !This is somewhat arbitrary, but giving preference to lowest rank's data
-  CALL myPE%allReduceMinI(SIZE(charProc),charProc)
+  CALL myPE%allReduceMaxI(SIZE(charProc),charProc)
```
HendersonSC, line 1055: This change actually reverses the above comment. Like the comment says its some what arbitrary, but it had the effect of always preserving the data on the master process. I don't know that one way is better than the other, but the comment should at least be conssitent.

src/ParallelEnv.f90
```diff
@@ -1070,7 +1070,7 @@ SUBROUTINE gather_str2D_MPI_ENV_type(myPE,sendbuf,root)
     iEntry = 0
     DO j=1,UBOUND(sendbuf,DIM=2)
       DO i=1,UBOUND(sendbuf,DIM=1)
-        IF(charProc(i,j) /= rank) THEN
+        IF(charProc(i,j) /= rank .AND. charProc(i,j) >= 0) THEN
```
HendersonSC, line 1073: If we would rather maintain the master process's data, then instead of using `-HUGE` we could keep `HUGE` and change `charProc(i,j) >= 0` to `charProc(i,j) <= myPE%nProc`. I have no real preferences either way, just thought I would mention it.

HendersonSC: The same two comments would apply to the the 1D version as well.

HendersonSC (changes requested):

src/ParallelEnv.f90
aarograh: oh, yeah, I didn't even read the comment.  I'll update that.

src/ParallelEnv.f90
aarograh: talked offline.  We'll stick with how I did it since the checks are a bit simpler and it doesn't really make a significant difference anyways.

src/ParallelEnv.f90
aarograh: fixed

HendersonSC (approved):

rks171 (approved):

PR #296, opened 2020-12-17
Hotfix string gather

aarograh:
The tag was only incremented when the non-root process was doing a send, when it actually needed to increment for every element of the array.


aarograh:
whoops, pushed an old branch instead.  Ignore this.

PR #299, opened 2021-01-17
Implements Base Solution Acceleration Methods (#4717)

bscollin:
Implements a base class for relaxed Picard, Modified Picard, and AndersonAcceleration as described in:

https://code.ornl.gov/vera/vera-dev/-/issues/4717


bscollin:
Do not merge until MPACT change is ready to merge.

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+!                             All rights reserved.                             !
+!                                                                              !
+! Futility is a jointly-maintained, open-source project between the University !
+! of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
+! can be found in LICENSE.txt in the head directory of this repository.        !
+!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+!> @brief A Fortran 2003 module that encompasses a base class for solution
+!>        acceleration.  Module also includes both picard iteration and a
+!>        modified picard iteration.  Anderson Acceleration extends this base
+!>        class in the AndersonAcceleration module.
```
aarograh, line 12: Typically we have another line of `!++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!` before the actual code starts...

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+  CONTAINS
+    !> @copybrief SolutionAccelerationType::init
+    !> @copydetails SolutionAccelerationType::init
+    PROCEDURE(init_absintfc),DEFERRED,PASS :: init
+    !> @copybrief SolutionAccelerationType::init_base_SolutionAccelerationType
+    !> @copydetails SolutionAccelerationType::init_base_SolutionAccelerationType
+    PROCEDURE,PASS :: init_base => init_base_SolutionAccelerationType
+    !> @copybrief SolutionAccelerationType::clear
+    !> @copydetails SolutionAccelerationType::clear
+    PROCEDURE(clear_absintfc),DEFERRED,PASS :: clear
```
aarograh, line 56: You could actually do `PROCEDURE(clear_base_SolutionAccelerationType)` here and not define the abstract interface.  Reduces a few lines of code and basically says "the concrete clear routines need to have the same interface as the base clear routine"

You might be able to do the same thing with `init`, but I haven't gotten far enough down to see if that interface holds or not

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+    !> @copybrief ModifiedPicardType::setInitial_ModifiedPicardType
+    !> @copydetails ModifiedPicardType::setInitial_ModifiedPicardType
+    PROCEDURE,PASS :: setInitial => setInitial_ModifiedPicardType
+    !> @copybrief ModifiedPicardType::step_ModifiedPicardType
+    !> @copydetails ModifiedPicardType::step_ModifiedPicardType
+    PROCEDURE,PASS :: step => step_ModifiedPicardType
+ENDTYPE ModifiedPicardType
+
+!> Name of module
+CHARACTER(LEN=*),PARAMETER :: modName='SolutionAccelerationModules'
```
aarograh, line 152: no `s` at the end

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+!
+!===============================================================================
+CONTAINS
+!
+!-------------------------------------------------------------------------------
+!> @brief Initializes the Solution Acceleration base class
+!> @param solver The Solution Acceleration solver to act on
+!> @param Params The options parameter list
+!> @param ce The computing environment to use for the calculation
```
aarograh, line 161: should have these in the order they appear in

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+        DO i=1,SIZE(solver%x)
+          CALL solver%x(i)%clear()
+        ENDDO
+        DEALLOCATE(solver%x)
+      ENDIF
+      solver%s=0
+      solver%n=-1
+      solver%start=1
+    ENDIF
+  ENDSUBROUTINE clear_base_SolutionAccelerationType
```
aarograh, line 208: Need to set `isInit` to false

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+  TYPE(ParamType),INTENT(IN) :: Params
+
+  CHARACTER(LEN=*),PARAMETER :: myName='init_RelaxedPicardType'
+
+  REQUIRE(.NOT.solver%isInit)
+
+  CALL solver%init_base(ce,Params)
+
+  IF(Params%has('RelaxedPicardType->alpha')) &
+      CALL Params%get('RelaxedPicardType->alpha',solver%alpha)
```
aarograh, line 228: I'm ok with this being optional... is `1.0` the right default though?

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+
+  IF(solver%alpha <= 0.0_SRK) CALL ce%exceptHandler%raiseError('Incorrect input to '//modName// &
+      '::'//myName//' - Relaxation coefficient must be greater than 0!')
```
aarograh, line 231: do we want a limit on over-relaxation too?  Or just leave it to the client not to be insane?

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+!> @param solver The solver to act on
+!>
+SUBROUTINE clear_RelaxedPicardType(solver)
+  CLASS(RelaxedPicardType),INTENT(INOUT) :: solver
+
+  CALL solver%clear_base()
+
+  IF(solver%isInit) THEN
+    solver%alpha=1.0_SRK
+    solver%isInit=.FALSE.
```
aarograh, line 247: Oh, I see it now... really these should probably just be in `clear_base` in my opinion... calling `init_base` is not sufficient to set `isInit=.TRUE.`, but calling clear_base is definitely sufficient to set it to `.FALSE.`

aarograh (changes requested):
Overall looks good.  A few comments.  I'm also not quite done looking through it but my computer's about to force a restart and I didn't want the review to be weird from stopping in the the middle

src/SolutionAcceleration.f90
bscollin: But if the base class doesn't have the interface defined, then we'll need select types everywhere to handle.  Or are you saying use clear and clear_base as the same?

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+SUBROUTINE step_ModifiedPicardType(solver,x_new)
+  CLASS(ModifiedPicardType),INTENT(INOUT) :: solver
+  CLASS(VectorType),INTENT(INOUT) :: x_new
+
+  REQUIRE(x_new%n == solver%n)
+
+  !Update iteration counter
+  solver%s=solver%s+1
+
+  CALL BLAS_copy(x_new,solver%tmpvec)
```
aarograh, line 384: Is it required that `setInital` get called before `step`?  If not, then I think this call might fail due to `%tmpvec` not yet being allocated

src/SolutionAcceleration.f90
bscollin: I just did it the same direction @jpjones6 did it :)

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,395 @@
...
+
+  IF(solver%s >= solver%start) THEN
+    !Get under-relaxed solution using mixing parameter
+    CALL BLAS_scal(x_new,solver%alpha)
+    CALL BLAS_axpy(solver%x(1),x_new,1.0_SRK-solver%alpha)
+  ENDIF
+  CALL BLAS_copy(solver%tmpvec,solver%x(1))
```
aarograh, line 391: Do we really need `%tmpvec` on the class?  Seems like it could just be a local variable for this routine

src/SolutionAcceleration.f90
bscollin: I thought it should be part of the extending class to set isInit to false.  This is why I didn't duplicate it as the main clear routine.  My view is isInit shouldn't be set to false until everything is cleared.

src/SolutionAcceleration.f90
bscollin: I think so, it defaults to a standard Picard iteration.  0 is not an option.  You could maybe argue 0.5, but I don't have a huge preference.

src/SolutionAcceleration.f90
bscollin: There are definitely many cases where overrelaxation makes sense.  We could probably limit it to something like 10, but at this point of the code, I don't really want to be limiting the use.

src/SolutionAcceleration.f90
bscollin: Yes, setInitial needs to be called first.  Honestly, this is consistent with the behavior of Anderson only you had to call reset first in Anderson.

src/SolutionAcceleration.f90
bscollin: Because it could potentially be a large memory allocation depending on vector size... and potentially a distributed vector, I thought it would be better to store it then determine the polymorphic state of a vector and allocate it and destroy it each time.  It was potentially an optimization without a real need, but it would definitely get rid of some code.

src/SolutionAcceleration.f90
aarograh: I'm saying you can pass an actual subroutine name into the `PROCEDURE(...)`.  You only need the abstract interface `clear_absintfc` if there's no routine in the module that defines the right interface.

Another option is to not have `clear` be deferred... just point it at `clear_base_SolutionAccelerationType` and let it get overridden later.  That's probably simpler actually

src/SolutionAcceleration.f90
aarograh: leave things better than you found them ;)

src/SolutionAcceleration.f90
aarograh: Hmm... my view is that `isInit` should only be true if the object is fully initialized... and therefore can be set to false if any part of the object has been cleared...

src/SolutionAcceleration.f90
aarograh: yeah, ok, I'm good with a non-relaxed Picard iteration being the default.  Just wanted to ask.

src/SolutionAcceleration.f90
aarograh: ok, works for me

src/SolutionAcceleration.f90
aarograh: Ok... if we're going to require `setInitial` to be called, then `step` should really have some sort of DBC statement ensuring that the vector is set up.

That DBC statement should really be in the other 2 classes too, even if it's not strictly required.  We should ensure the same set of calls for all classes.

src/SolutionAcceleration.f90
aarograh: That's fair... I think my bigger concern is just that we seem to have required `setInitial` to be called for this class and not others, and making `tmpVec` completely local would resolve that... I could be wrong on the other classes though

bscollin: yeah, they all need `x(:)` to be allocated too... which `setInitial` also handles

src/SolutionAcceleration.f90
bscollin: Yeah, that's fair enough.

src/SolutionAcceleration.f90
bscollin: Done

src/SolutionAcceleration.f90
bscollin: Fixed

src/SolutionAcceleration.f90
bscollin: Fixed both here and in Anderson

src/SolutionAcceleration.f90
bscollin: Fixed

src/SolutionAcceleration.f90
bscollin: Fixed

src/SolutionAcceleration.f90
bscollin: we good on this one?

src/SolutionAcceleration.f90
jpjones6: Definitely wouldn't follow my example :P.

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,385 @@
...
+  !Pull Data from Parameter List
+  CALL Params%get('SolutionAcceleration->num_unknowns',solver%n)
+  IF(Params%has('SolutionAcceleration->starting_iteration')) &
+      CALL Params%get('SolutionAcceleration->starting_iteration',solver%start)
+
+  IF(solver%n < 1) CALL ce%exceptHandler%raiseError('Incorrect input to '//modName// &
+      '::'//myName//' - Number of unkowns (n) must be greater than 0!')
+
+  IF(solver%start < 1) CALL ce%exceptHandler%raiseError('Incorrect input to '//modName// &
+      '::'//myName//' - Starting point must be greater than 1!')
```
jpjones6, line 174: Greater than 0.

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,385 @@
...
+
+  REQUIRE(.NOT.solver%isInit)
+
+  CALL solver%init_base(ce,Params)
+
+  IF(Params%has('SolutionAcceleration->picard_relaxation_factor')) &
+      CALL Params%get('SolutionAcceleration->picard_relaxation_factor',solver%alpha)
+
+  IF(solver%alpha <= 0.0_SRK) CALL ce%exceptHandler%raiseError('Incorrect input to '//modName// &
+      '::'//myName//' - Relaxation coefficient must be greater than 0!')
```
jpjones6, line 222: Leaving open the possibility of alpha > 1.0 are we? How devious.

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,385 @@
...
+
+  solver%alpha=1.0_SRK
+  solver%isInit=.FALSE.
+
+ENDSUBROUTINE clear_RelaxedPicardType
+!
+!-------------------------------------------------------------------------------
+!> @brief Sets the initial vector for relaxation
+!> @param solver Solver to take step with
+!> @param x_new New solution iterate and under-relaxed / accelerated return vector
```
jpjones6, line 244: Seems like you changed to just x, or maybe never changed to x_new and meant to...

src/SolutionAcceleration.f90
```diff
@@ -0,0 +1,385 @@
...
+    CALL solver%tmpvec%clear()
+    DEALLOCATE(solver%tmpvec)
+  ENDIF
+  solver%isInit=.FALSE.
+ENDSUBROUTINE clear_ModifiedPicardType
+!
+!-------------------------------------------------------------------------------
+!> @brief Sets the initial vector for relaxation
+!> @param solver Solver to take step with
+!> @param x_new New solution iterate and under-relaxed / accelerated return vector
```
jpjones6, line 345: x

src/SolutionAcceleration.f90
jpjones6: DBC checks to guard against what Aaron mentioned would be nice. Just need one on `tmpVec` I think.

src/AndersonAcceleration.f90
```diff
@@ -29,21 +30,11 @@ MODULE AndersonAccelerationModule
...
-  INTEGER(SIK) :: n=-1
-  !> Current iteration count
-  INTEGER(SIK) :: s=0
+TYPE,EXTENDS(SolutionAccelerationType) :: AndersonAccelerationType
   !> Iteration depth of Anderson solver
   INTEGER(SIK) :: depth=1
-  !> Starting iteration for Anderson
-  INTEGER(SIK) :: start=1
   !> Value of mixing parameter
   REAL(SRK) :: beta=0.5_SRK
```
jpjones6, line 37: Seems like the mixing parameter for Anderson (beta) and the under-relaxation parameter (alpha) for the Picard classes could just be put on the base class as beta and combined. They both serve the same function.

bscollin: my basic thought on this was they are called different things in the literature.  I also don't think there's any literature out there that would allow beta > 1, but there'a a lot out there on overrelaxing.  I'm not just thinking about this for coupling, but potentially replacing our current relaxation for currents in cmfd, etc.  Basically anywhere we have a relaxation applied, to use this class... then allow us to more easily experiment with trying Anderson in other areas.

src/SolutionAcceleration.f90
bscollin: Added

src/SolutionAcceleration.f90
bscollin: added DBC checks

src/SolutionAcceleration.f90
bscollin: Fixed

src/SolutionAcceleration.f90
bscollin: yes

src/SolutionAcceleration.f90
bscollin: copy/paste error.  Fixed

src/SolutionAcceleration.f90
bscollin: Fixed

src/AndersonAcceleration.f90
jpjones6: Okay. I'm good with leaving them differentiated.

src/SolutionAcceleration.f90
jpjones6: Looks good to me. @aarograh are you happy with this one?

src/SolutionAcceleration.f90
jpjones6: After looking at the next routine I figured as much. All good now.

aarograh (approved):

aarograh:
This is ready once MPACT is ready.

PR #301, opened 2021-01-25
Hotfix geom graph combine

aarograh:


aarograh:
Was just using this to compare a rework.  That second branch is now much further along than this one, so I'm going to close this

PR #302, opened 2021-01-28
Hotfix geom graph combine take2

aarograh:
This rewrites the GraphType combine method to account for several robustness issues.


src/Geom_Graph.f90
```diff
@@ -294,32 +361,32 @@ ELEMENTAL FUNCTION getAdjacentVert_graphType(thisGraph,v0,i) RESULT(v1)
 !>
 !>
 !>
-ELEMENTAL FUNCTION getCWMostVert_graphType(thisGraph,v0,vCurr) RESULT(vNext)
-  CLASS(GraphType),INTENT(IN) :: thisGraph
-  INTEGER(SIK),INTENT(IN) :: vCurr
+ELEMENTAL FUNCTION getCWMostVert_graphType(this,v0,vCurr) RESULT(vNext)
```
HendersonSC, line 364: Looks like you got a lot of Doxygen added but some is still missing.

src/Geom_Graph.f90
```diff
@@ -645,16 +820,16 @@ PURE SUBROUTINE getMidPointOnEdge_graphType(thisGraph,v1,v2,m)
   REAL(SRK) :: a(2),b(2),c(2),r,alp1,alp2,theta,scal
 
   m=-HUGE(m)
-  n=nVert_graphType(thisGraph)+1
+  n=this%nVert()+1
   IF(v1 > 0 .AND. v2 > 0 .AND. v1 < n .AND. v2 < n) THEN
-    IF(thisGraph%edgeMatrix(v1,v2) == 1) THEN
-      m=thisGraph%vertices(:,v1)+thisGraph%vertices(:,v2)
+    IF(this%edgeMatrix(v1,v2) == 1) THEN
```
HendersonSC, line 825: Looks like you can use the enumerations here.

HendersonSC (commented):
Overall good start...You already know about the debug statements so I will give you a chance to clear those up before commenting on them.

It looks like you have added some comments on how it all works, but the is large enough its hard to expand it and make sure it all flows. I will try to look at that in a little more detail on a second look.

HendersonSC:
As far as comments and process go, it is clearly better. I don't know where you want to draw the line, someone just getting into graphs might not fully keep up, but I think generally between the naming and comments its in a decent place.

src/Geom_Graph.f90
aarograh: Yeah... there are a handful (including this one) that I simply don't know exactly what they're doing because I didn't really have to rely on them at all.  I just left them blank since I don't want to put something wrong in there.

src/Geom_Graph.f90
aarograh: good catch.  I need to do another pass for enumerations... unfortunately searching for 1, -1, and 0 is kind of annoying

aarograh: fixed as many of these as I could find

aarograh:
Alright, I think this is all set if you want to give it a final look over @HendersonSC .  The only test that failed is testGenReqTables, which has nothing to do with my changes.  The docker container downloads a bunch of python packages each time it runs and we think that one of the packages got changed in a way that broke that test.

src/Geom_Graph.f90
HendersonSC: Understood...I think we should at least try to tag them with an issue or something.

aarograh: I can get on board with that

aarograh: #303 has it

HendersonSC (approved):

aarograh:
the failed test is testGenReqTables due to python package changes and has nothing to do with these changes.  Going ahead with the merge.

PR #304, opened 2021-02-05
Fix for PETSc 3.12.4

rks171:
Description:
For this newer version of PETSc, a parallel run using SuperLU_dist needs to
specifically call that solver (not SuperLU, which only works for serial).  Older
versions of PETSc allowed the solver to be set to SuperLU whether it was serial
or parallel.

CASL Ticket # - 4661


aarograh (approved):

aarograh:
Failing test is a known python package issue and unrelated to these changes.  Going ahead with the merge.

PR #306, opened 2021-02-10
Fix superlu_dist issues by using it for serial solvers too

aarograh:
Just use superlu_dist for both serial and parallel solves to fix valgrind defects


rks171 (approved):
Looks good.

aarograh:
Failure is not relevant to these changes.  merging.

PR #309, opened 2021-02-25
Adding overloaded assignments for geometries, and 90 degree rotation.

djabaay:
Description:
These functions are needed to help speed up the geometry processing in
MPACT.

VERA-dev Issue # - ####

### DIFF
--- a/src/Geom_Box.f90
+++ b/src/Geom_Box.f90
@@ -30,7 +30,6 @@ PRIVATE
 PUBLIC OBBoxType
 PUBLIC ABBoxType
 PUBLIC :: OPERATOR(==)
-PUBLIC :: ASSIGNMENT(=)
 
 !> @brief Type for a Box
 TYPE OBBoxType
@@ -107,15 +106,6 @@ INTERFACE OPERATOR(==)
   !> @copybrief Geom_Box::isequal_OBBoxType
   !> @copydetails Geom_Box::isequal_OBBoxType
   MODULE PROCEDURE isequal_OBBoxType
-ENDINTERFACE
-
-!> @brief Generic interface for assignment operator (=)
-!>
-!> Adds assignment capability for OBBox types.
-INTERFACE ASSIGNMENT(=)
-  !> @copybrief GeomPoints::assign_OBBoxType
-  !> @copydetails GeomPoints::assign_OBBoxType
-  MODULE PROCEDURE assign_OBBoxType
 ENDINTERFACE
 !
 !===============================================================================
@@ -516,25 +506,6 @@ ELEMENTAL FUNCTION isequal_OBBoxType(b0,b1) RESULT(bool)
   IF(b0%p0 == b1%p0 .AND. ALL(b0%u .APPROXEQA. b1%u)) &
       bool=ALL(b0%e.APPROXEQA. b1%e)
 ENDFUNCTION isequal_OBBoxType
-!
-!-------------------------------------------------------------------------------
-!> @brief Defines the assignment operation between two OBBoxes
-!> e.g. b0 = b1
-!> @param b0 the first box
-!> @param b1 the second box
-!>
-!> Function is elemental so it can be used on an array of OBBoxes.
-ELEMENTAL SUBROUTINE assign_OBBoxType(b0,b1)
-  TYPE(OBBoxType),INTENT(INOUT) :: b0
-  TYPE(OBBoxType),INTENT(IN) :: b1
-  IF(b1%p0%dim > 0) THEN
-    b0%p0=b1%p0
-    b0%u=b1%u
-    b0%e=b1%e
-  ELSE
-    CALL b0%clear()
-  ENDIF
-ENDSUBROUTINE assign_OBBoxType
 !
 !-------------------------------------------------------------------------------
 !> @brief Constructor for the ABBoxType
--- a/src/Geom_Box.f90
+++ b/src/Geom_Box.f90
@@ -30,6 +30,7 @@ PRIVATE
 PUBLIC OBBoxType
 PUBLIC ABBoxType
 PUBLIC :: OPERATOR(==)
+PUBLIC :: ASSIGNMENT(=)
 
 !> @brief Type for a Box
 TYPE OBBoxType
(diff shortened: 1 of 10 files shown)


djabaay:
CI failures are from the genReqTables test.

src/Geom_Box.f90
```diff
@@ -107,6 +108,15 @@ MODULE Geom_Box
...
 ENDINTERFACE
+
+!> @brief Generic interface for assignment operator (=)
+!>
+!> Adds assignment capability for OBBox types.
+INTERFACE ASSIGNMENT(=)
+  !> @copybrief GeomPoints::assign_OBBoxType
+  !> @copydetails GeomPoints::assign_OBBoxType
+  MODULE PROCEDURE assign_OBBoxType
+ENDINTERFACE
```
aarograh, line 119: So... why is this actually necessary?  Fortran has intrinsic assignment for derived types.  Since there are no pointer components or anything like that, it seems like the intrinsic assignment should be sufficient (and more robust since we don't need to ever add components to an assignment procedure we define)

src/Geom_CircCyl.f90
```diff
@@ -107,6 +108,15 @@ MODULE Geom_CircCyl
...
 ENDINTERFACE
+
+!> @brief Generic interface for assignment operator (=)
+!>
+!> Adds assignment capability for Circle types.
+INTERFACE ASSIGNMENT(=)
+  !> @copybrief GeomPoints::assign_CircleType
+  !> @copydetails GeomPoints::assign_CircleType
+  MODULE PROCEDURE assign_CircleType
+ENDINTERFACE
```
aarograh, line 119: Same question here about the necessity of assignment

src/Geom_Poly.f90
```diff
@@ -156,6 +158,13 @@ MODULE Geom_Poly
   !> @copydetails Geom_Poly::isequal_PolygonType
   MODULE PROCEDURE isequal_PolygonType
 ENDINTERFACE
+
+!> @brief
+INTERFACE ASSIGNMENT(=)
+  !> @copybrief Geom_Poly::assign_PolygonType
+  !> @copydetails Geom_Poly::assign_PolygonType
+  MODULE PROCEDURE assign_PolygonType
+ENDINTERFACE
```
aarograh, line 167: Same question here about the necessity of assignment

aarograh (commented):
This is mostly fine.  My main concern is about the necessity of writing our own assignment routines.  They tend to be brittle and have gotten us in the past.

To be honest, you can leave those tests in place and just delete the assignment procedures and see if it still works.  It should unless the compiler is completely retarded.

src/Geom_Box.f90
djabaay: Understood.  These types aren't going to change, but your point is valid.  I'll try deleting the assignment overload and seeing if the tests still pass.

src/Geom_Poly.f90
djabaay: The other two are probably fine to delete.  The polygontype does have pointer assignments to other polygontypes, so this one might be good to keep.

aarograh: Ah, good point.   I missed those 2 pointers at the bottom.  So yeah, definitely keep this one.

src/Geom_Poly.f90
```diff
@@ -2003,6 +2012,35 @@ RECURSIVE FUNCTION isequal_PolygonType(p1,p2) RESULT(bool)
...
+    ALLOCATE(p1%vert(p2%nVert))
+    ALLOCATE(p1%edge(2,p2%nVert))
+    p1%vert=p2%vert
+    p1%edge=p2%edge
+    ALLOCATE(p1%quad2edge(p2%nQuadEdge))
+    ALLOCATE(p1%quadEdge(3,p2%nQuadEdge))
+    p1%quad2edge=p2%quad2edge
+    p1%quadEdge=p2%quadEdge
+    p1%nextPoly => p2%nextPoly
+    p1%subRegions => p2%subRegions
```
aarograh, line 2039: Just want to double check... a shallow copy is the right way to go here?  Or do we want to do a deep copy?  (i.e., allocate `nextPoly` and `subRegions` then assign the `p2` versions to `p1`)

djabaay: Good question.  Let me look through the general pin mesh type and think on how the polygon type is used there.  I think that will dictate whether it should be shallow or deep.

aarograh: Sounds good.  Please add a comment on this routine's header that explicitly states whether it's deep or shallow copy too.

src/Geom_Box.f90
djabaay: Tests passed.  Removed.

src/Geom_CircCyl.f90
djabaay: Tests passed when removed.

src/Geom_Poly.f90
```diff
@@ -156,6 +158,13 @@ MODULE Geom_Poly
   !> @copydetails Geom_Poly::isequal_PolygonType
   MODULE PROCEDURE isequal_PolygonType
 ENDINTERFACE
+
+!> @brief
```
aarograh, line 162: Also should fill this in if you're going to have it

aarograh:
Just 2 comments about comments... then we can go ahead and merge

src/Geom_Poly.f90
djabaay: I'm going to say that since the `%clear` routine only nullifies, and when the `%subregions` pointer and `%nextpoly` pointer are assigned, they are only allocated, it should be a shallow copy.  Otherwise, if we allocate the pointers when we assign, and we clear the assigned polygon, those allocated pointers will still exist and be lost.

djabaay:
Updated.  Filled in the rest of the Geom_Poly subroutine headers while I was at it.

aarograh (approved):

src/Geom_Poly.f90
aarograh: Great, thanks for filling all that in.  There were a lot more empty ones than I realized.

src/Geom_Poly.f90
aarograh: Yep, makes sense.  Thanks!

aarograh:
Everything looks good and that failure is unrelated to these changes.  Merging...

PR #314, opened 2021-03-11
Disables the FMU unit test if BUILD_SHARED_LIBS is FALSE (#4647)

wgurecky:
Disables the FMU unit test if BUILD_SHARED_LIBS is FALSE since FMU test requires shared libs.


aarograh:
@wgurecky Just confirming here that you tested this version with a static mpact build and did not encounter linking issues?

aarograh:
will confirmed that static builds work with this change.

aarograh (approved):

PR #315, opened 2021-03-11
Add IAPWS Water Property Tables (#4795)

bscollin:


src/IAPWS_WaterProperties.f90
```diff
@@ -0,0 +1,4258 @@
...
+      d1 = DCrit
+    ENDIF
+  ENDIF
+
+  CALL WNPT3(d1, d2, p, t, eps, x)
+  DITER3 = x
+ENDFUNCTION diter3
+!
+!-------------------------------------------------------------------------------
+!TODO needs header
```
aarograh, line 3892: Several functions have these TODO... if possible it would be good to fill those in (but I certainly couldn't give a reasonable description of most of them)

src/IAPWS_WaterProperties.f90
```diff
@@ -0,0 +1,4258 @@
...
+  F3 = nullp3n(X3, T, P)
+
+  x = 0.0_SRK
+  DO I = 1, 40
+    IF (F1 /= F3) THEN
+      X = X1 + (X3 - X1)*F1/(F1 - F3)
+    ELSE
+      EXIT
+    ENDIF
+    IF (X < 0.0_SRK) X = (X1 + X3)/2.0_SRK
```
aarograh, line 3931: This function and `wnph3` (and probably others) have a decent number of `/2.0_SRK` that could be changed to multiplication.  Might be good to try to search on any divisions like that... maybe regex `/[0-9]*\.[0-9]*_` or something like that to get divisions by reals and not divisions by variables...

src/IAPWS_WaterProperties.f90
```diff
@@ -0,0 +1,4258 @@
...
+    a1 = 646.836937239618_SRK, &
+    a2 = 62.7217723183194_SRK, &
+    a3 = -18.0176207184787_SRK, &
+    a4 = 2.01708278241311_SRK, &
+    a5 = 247.531572315595_SRK, &
+    a6 = -41.1717709139789_SRK, &
+    a7 = -150.208861198343_SRK, &
+    a8 = 1431.77901876243_SRK, &
+    a9 = -862.734816079672_SRK, &
+    a0 = 49.9470109089612_SRK
```
aarograh, line 3996: So continuation lines should be 4 spaces indented compared to the start of the line.  This occurs in quite a few places throughout the module.  If you can fix that without spending an inordinate amount of time, please do.

src/IAPWS_WaterProperties.f90
```diff
@@ -0,0 +1,4258 @@
...
+  REAL(SRK) :: hr, pr
+
+  EPS = 1.E-12
```
aarograh, line 3999: seems like this should be a parameter?  But not really a huge deal.  I imagine something like this occurs elsewhere too.

src/IAPWS_WaterProperties.f90
```diff
@@ -0,0 +1,4258 @@
...
+  ENDIF
+ENDFUNCTION vest3
+!
+!-------------------------------------------------------------------------------
+!> Returns the specific volume of the phase.  Works for saturated liquid or
+!> vapor.
+!> @param P    The saturation pressure
+!> @param T    The saturation temperature
+!> @param DEST The density of the fluid [kg/m**3]
+!> @param EPS  TODO - what is EPS
```
aarograh, line 3858: Through pure luck, happened to notice this... would be good to fill in the TODO if possible

unit_tests/testIAPWSWaterProperties/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+#++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
+#                          Futility Development Group                          !
+#                             All rights reserved.                             !
+#                                                                              !
+# Futility is a jointly-maintained, open-source project between the University !
+# of Michigan and Oak Ridge National Laboratory.  The copyright and license    !
+# can be found in LICENSE.txt in the head directory of this repository.        !
+#++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++!
```
aarograh, line 8: Do we usually put our headers on cmake files?  I thought it was just .f90 and .C files.  It's not hurting anything, just more of a question on whether we should make this change elsewhere too.  I definitely haven't added these to MPACT cmake files.

unit_tests/testIAPWSWaterProperties/testIAPWSWaterProperties.f90
```diff
@@ -0,0 +1,512 @@
...
+  ! Unit Test comparing function to LITERATURE values
+  ! These reference values are provided in:
+  ! J. Cooper, "The International Association for the Properties of Water and Steam", Berlin,
+  ! Germany, 2008.
+  ! For the verification of computer program implementations of the tables.
+  ! See Table 4.
+  REAL(SRK), PARAMETER, DIMENSION(11) :: & ! Temp. in [K], Density in [kg/m^3], and viscosity [uPa-s] (from literature)
+    T = [298.15E0_SRK, 298.15E0_SRK, 373.15E0_SRK, 433.15E0_SRK, 433.15E0_SRK, 873.15E0_SRK, 873.15E0_SRK, 873.15E0_SRK, 1173.15E0_SRK, 1173.15E0_SRK, 1173.15E0_SRK], &
+    Rho = [998.0E0_SRK, 1200.0E0_SRK, 1000.0E0_SRK, 1.0E0_SRK, 1000.0E0_SRK, 1.0E0_SRK, 100.0E0_SRK, 600.0E0_SRK, 1.0E0_SRK, 100.0E0_SRK, 400.0E0_SRK], &
+    U_lit = [889.735100_SRK, 1437.649467_SRK, 307.883622_SRK, 14.538324_SRK, 217.685358_SRK, 32.619287_SRK, 35.802262_SRK, 77.430195_SRK, 44.217245_SRK, 47.640433_SRK, 64.154608_SRK]
```
aarograh, line 301: Same comments on continuation lines here and throughout the test

aarograh (changes requested):
Had a few comments on this.  I'm open to some discussion and judgment calls about the time required to address them since most of them are formatting, but figured I'd at least point them out.

src/IAPWS_WaterProperties.f90
bscollin: @rks171 would have to add those because I have no idea what they are doing either.  Thats why I documented them so they were easy enough to find.

src/IAPWS_WaterProperties.f90
bscollin: yeah.... thought about it, but its technically not wrong.  I can do these pretty quickly, at least 2.0.  Note that I'm a little fearful to use Constants module because pi is used as a variable to be consistent with the documentation.  (there are also some cases where PI is 3.14...  I will change /2.0  to *0.5, some for /4.0... maybe /8.0

src/IAPWS_WaterProperties.f90
bscollin: if I knew what it was... I'd fill it in :)

unit_tests/testIAPWSWaterProperties/CMakeLists.txt
bscollin: I"m 99.999% positive I just copied this from another cmake file.

src/IAPWS_WaterProperties.f90
bscollin: maybe... definitely didsn't search out use of any variable

src/IAPWS_WaterProperties.f90
bscollin: I'm not sure I follow.  There are several places where they are trying to align digits, should all of those be 4 spaces and not aligh numbers?  I thought there was some lattitude to improve readability.

src/IAPWS_WaterProperties.f90
bscollin: Fixed all `/2.0_SRK`'s... also there were several `*1.0_SRK`'s.  I didn't find anything else (no divide by 4, 5, 8, or 10.

src/IAPWS_WaterProperties.f90
bscollin: Tried to align these things better.  DIdn't fully enforce 4 spaces, but when it made sense, I did, Also aligned decimal points for several large lists of parameters.

src/IAPWS_WaterProperties.f90
bscollin: Fixed this one.  Not sure how to find them generally.

src/IAPWS_WaterProperties.f90
aarograh: Yeah, I'll take a look... to be more precise, continuation lines should be _at least_ 4 spaces.  I don't want to see less than that, but using more than that can certainly improve readability (sometimes I'll just add a couple spaces to the start of the line so that the continuation lines will align with it when using 4 spaces for them)

src/IAPWS_WaterProperties.f90
aarograh: yeah me neither.  The review was somewhat just spot checking, so stuff like this I noticed was pretty much just pure chance.

src/IAPWS_WaterProperties.f90
aarograh: I'm pretty good with how you did it.  The times you didn't do 4 spaces, it was always greater than 4 spaces and made sense to do something else.

aarograh:
It'd be good if @rks171 could help out with some of the `TODO` comments in the headers too.  Those are my only remaining comments from the review.

aarograh (approved):

PR #317, opened 2021-03-17
Fix path to get-pip

aarograh:
Specifies python 2.7 in the get-pip URL


wgurecky (approved):
I reviewed the changes to the get-pip url, verified this url is correct.

PR #319, opened 2021-03-23
XDMF

KyleVaughn:
Adds XDMF mesh type and file type. Allows for reading and writing of 2D XMDF mesh files.


KyleVaughn:
Note that XML element members were made public because the PRIVATE attribute was applied selectively, leaving things like `content` and `name` public despite having `setName` or `getContent` and because the setters and getters are largely unnecessary. For instance, `getParent` is just a call to `parent => thisXMLE%parent` and `getContent` is just `content=thisXMLE%content`.

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+!> @param mesh the mesh
+!> @param xmle the XML element
+!> @param strpath the string holding the path to the XDMF file
+!> @param h5 the HDF5 file
+!>
+RECURSIVE SUBROUTINE export_leaf_XDMFFileType(mesh, xmle, strpath, h5)
+  CHARACTER(LEN=*),PARAMETER :: myName='export_leaf_XDMFFileType'
+  TYPE(XDMFMeshType),INTENT(IN)  :: mesh
+  TYPE(XMLElementType),TARGET,INTENT(INOUT) :: xmle
+  TYPE(StringType),INTENT(INOUT) :: strpath
```
thfolk, line 835: Can this be INTENT(IN)?

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+  INTEGER(SLK) :: xdmf_id, nverts, ncells, i, j, vctr
+  CHARACTER(LEN=200) :: charpath
+  INTEGER(SLK), ALLOCATABLE :: vertex_list_2d(:, :), vertex_list_1d(:), cell_list_1d(:)
+
+  ! Create HDF5 group
+  CALL h5%mkdir(CHAR(mesh%name))
+
+  ! Determine number of xmle children
+  ! Geometry, Topology, Material, Cell sets
+  ! Assumes mesh has geometry and topology
```
thfolk, line 849: Add DBC statement?

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+IMPLICIT NONE
+PRIVATE
+
+#ifdef FUTILITY_HAVE_HDF5
+! Public members
+PUBLIC :: XDMFFileType
+PUBLIC :: XDMFMeshType
+PUBLIC :: XDMFTopologyList
+PUBLIC :: ASSIGNMENT(=)
+
```
thfolk, line 36: I see a lot of indexing with 1, 2, 3, 4 integers which can be confusing when newly looking at the code. Can you assign named integers here so that more context is given for the indexing? I.e. GEOMETRY_IDX=1, TOPOLOGY_IDX=2, etc

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+  ! Determine number of xmle children
+  ! Geometry, Topology, Material, Cell sets
+  ! Assumes mesh has geometry and topology
+  nchildren = 2
+  IF(ALLOCATED(mesh%material_ids)) nchildren = nchildren + 1
+  IF(ALLOCATED(mesh%cell_sets)) nchildren = nchildren + SIZE(mesh%cell_sets)
+  ALLOCATE(xmle%children(nchildren))
+
+  ! GEOMETRY
+  current_xml => xmle%children(1)
```
thfolk, line 856: GEOMETRY_IDX = 1

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+  str_name="Geometry"
+  CALL current_xml%setName(str_name) 
+  current_xml%nAttr=0
+  current_xml%parent => xmle
+
+  str_name= "GeometryType"
+  str_value = "XYZ"
+  CALL current_xml%setAttribute(str_name, str_value)
+
+  ALLOCATE(current_xml%children(1))
```
thfolk, line 866: GEOMETRY_IDX

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+
+  str_name= "DataType"
+  str_value = "Float"
+  CALL child_xml%setAttribute(str_name, str_value)
+
+  nverts=SIZE(mesh%vertices, DIM=2)
+  str_name="Dimensions"
+  str1 = nverts
+  str2 = "3"
+  str_value = str1//" "//str2
```
thfolk, line 881: Can this be consolidated to one line without str1 and str2?

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+  CALL child_xml%setAttribute(str_name, str_value)
+
+  i = LEN_TRIM(strpath)
+  charpath = CHAR(strpath)
+  child_xml%content = charpath(1:i-4)//"h5:/"//mesh%name//"/vertices"
+
+  CALL h5%fwrite(CHAR(mesh%name)//'->vertices',mesh%vertices)
+
+  ! TOPOLOGY
+  current_xml => xmle%children(2)
```
thfolk, line 899: TOPOLOGY_IDX

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+    child_xml%nAttr=0
+    child_xml%parent => current_xml
+
+    str_name= "DataType"
+    str_value = "Int"
+    CALL child_xml%setAttribute(str_name, str_value)
+
+    str_name="Dimensions"
+    str1 = ncells
+    str2 = nverts
```
thfolk, line 937: nverts is set to str1 at line 879. can set str2 to ncells instead of reassigning both.

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+        CALL xmle_children(i)%getAttributeValue(strIn,strOut)
+        IF(strOut /= 'XYZ') THEN
+          CALL eXDMF%raiseWarning(modName//'::'//myName// &
+            ' - GeometryType only supports XYZ right now.')
+        ENDIF
+        ! Format
+        CALL xmle_children(i)%getChildren(ele_children)
+        REQUIRE(SIZE(ele_children) == 1)
+        strIn='Format'
+        CALL ele_children(1)%getAttributeValue(strIn,strOut)
```
thfolk, line 282: Instead of indexing children(1) throughout, add a named INTEGER. Is (1) the parent or a top level? Could do PARENT_IDX.

thfolk (commented):
Hey Kyle, overall looks great! Awesome job. I followed it pretty well for the most part. Just a few things I think could improve the readability.
1) Used named integers instead of indexing with 1, 2, 3, etc. Examples include GEOMETRY_IDX, TOPOLOGY_IDX, etc. for your indexes that are fixed. 
2) Most of the variables were descriptive. Some were not. I suggested a few changes to some variables that could show more context.
3) There are a lot of DBC statements filtered throughout. It's good that you are checking things as you go. From what I've learned about DBC statements they normally go at the beginning of a module. Some of your statements could be turned into IF blocks with an exception handler. Since there are so many DBC statements throughout, you could create a module to check all of these things in one place. 

Overall, great work!

aarograh:
@thfolk @KyleVaughn I'm seeing a couple comments about DBC vs. exception handlers.  First off, thanks for adding one or the other... something is always better than nothing!  I figured I'd take a minute to write a note about picking between the two though in case it's relevant to the discussion going on here:

* Exception Handling: Use this for bad conditions that could potentially be caused by user input.  For example, two input options are related to each other: input A can only have certain values depending on what value input B has.  That situations warrants a descriptive error message to provide feedback to the user.  We often can't reasonably test all input combinations during input processing, so we need to provide feedback downstream too.
* DBC: I think of this more as feedback to the developer.  In other words, if a DBC statement trips, then that probably means the developer called the routine in a way that it should never be called.  It should be rare for a user to be able to cause a DBC statement to appear, since "Design by Contract" really refers to how we as developers tie various functions and classes together.  If the developer obeyed the "contract", the user should never see this.  This is also partly why we disable DBC checks in release builds, since they shouldn't be encountered by users anyway.

I haven't looked in detail at this code so I don't know which of these is appropriate (maybe both are appropriate in different places), but hopefully that helps in deciding whether or not things should be changed from one to the other.

KyleVaughn:
Thanks @aarograh @thfolk !
It looks like I should change a decent amount of the DBC to exception handling, since it's primarily applied to assumptions about valid XMDF, valid HDF5, and what structure the mesh file will have.

thfolk:
Thanks @aarograh. I'm saving your comment to keep in the memory bank!

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+!> @param h5 the HDF5 file containing mesh data
+!>
+RECURSIVE SUBROUTINE create_XDMFMesh_from_file(mesh, xmle, h5)
+  CHARACTER(LEN=*),PARAMETER :: myName='create_XDMFMesh_from_file'
+  TYPE(XDMFMeshType),TARGET, INTENT(INOUT)  :: mesh
+  TYPE(XMLElementType), INTENT(INOUT) :: xmle
+  TYPE(HDF5FileType), INTENT(INOUT) :: h5
+  TYPE(XMLElementType), POINTER :: xmle_children(:)
+  TYPE(StringType) :: strIn, strOut
+  INTEGER(SIK) :: i, grid_ctr, mesh_ctr
```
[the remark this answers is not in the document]
KyleVaughn: Changed

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+        ! GeometryType
+        strIn='GeometryType'
+        CALL xmle_children(i)%getAttributeValue(strIn,strOut)
+        IF(strOut /= 'XYZ') THEN
+          CALL eXDMF%raiseWarning(modName//'::'//myName// &
+            ' - GeometryType only supports XYZ right now.')
+        ENDIF
+        ! Format
+        CALL xmle_children(i)%getChildren(ele_children)
+        REQUIRE(SIZE(ele_children) == 1)
```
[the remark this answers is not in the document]
KyleVaughn: Changed to error handler

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+        CALL ele_children(1)%getAttributeValue(strIn,strOut)
+        IF(strOut /= 'HDF') THEN
+          CALL eXDMF%raiseWarning(modName//'::'//myName// &
+            ' - only supports HDF5 geometry data right now.')
+        ENDIF
+        ! Vertex Data
+        strIn='Dimensions'
+        CALL ele_children(1)%getAttributeValue(strIn,strOut)
+        strArray=strOut%split()
+        REQUIRE(strArray(2) == '3')
```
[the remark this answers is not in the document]
KyleVaughn: Changed to error handler. Also, the error checking happens as each component is read in. The code below double checks that the h5 and xdmf data match and are the correct shape.

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+        nverts=strArray(1)%stoi()
+        ! This is all awful string manipulation to get the h5 group
+        content=ele_children(1)%getContent()
+        segments=content%split(':')
+        group=segments(2)%substr(2,LEN(segments(2)))
+        group = group%replace(NEW_LINE("A"),"")
+        REQUIRE(h5%pathExists(CHAR(group)))
+        group = group%replace("/", "->")
+        ! Data shape
```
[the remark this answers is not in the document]
KyleVaughn: Fixed string manipulation to be more readable.

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+  REQUIRE(children(1)%name%upper() == 'DOMAIN')
+
+  ! Information
+  ! NOTE: It is assumed that material information is before any grids
+  ! and that all grids are contained in one overall grid.
+  CALL children(1)%getChildren(children)
+  IF (SIZE(children) == 2) THEN
+    REQUIRE(children(1)%name%upper() == 'INFORMATION')
+    REQUIRE(children(2)%name%upper() == 'GRID')
+    i = 2
```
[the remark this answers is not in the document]
KyleVaughn: Changed

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1236 @@
...
+    DEALLOCATE(thismesh%cells)
+  ENDIF
+  IF( ALLOCATED(thismesh%material_ids) ) DEALLOCATE(thismesh%material_ids)
+  IF( ALLOCATED(thismesh%cell_sets) ) THEN
+    DO i=1, SIZE(thismesh%cell_sets)
+      DEALLOCATE(thismesh%cell_sets(i)%cell_list)
+    ENDDO
+    DEALLOCATE(thismesh%cell_sets)
+  ENDIF
+ENDSUBROUTINE clear_XDMFMeshType
```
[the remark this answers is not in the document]
KyleVaughn: No, but I checked that every member is deallocated from bottom up, so the only memory that should stick around is the memory needed to hold the uninitialized mesh object. Testing in Valgrind seems like overkill when the clear routine is only 20 lines.

src/FileType_XDMF.f90
KyleVaughn: Strpath can be. Changed

src/FileType_XDMF.f90
KyleVaughn: Added where possible. A lot of the 1 indexing is due to the assumption that there is only one child. There are checks for this, but now that I have it as error handling it should be more clear that the assumption exists.

src/FileType_XDMF.f90
KyleVaughn: Added

src/FileType_XDMF.f90
KyleVaughn: This one is actually the allocation of an xml element with one child, hence the 1. It will always only have 1 child, so GEOMETRY_IDX doesn't work here even if they are equivalent

src/FileType_XDMF.f90
KyleVaughn: Not without changing to a formatted write statement into a character array. The str1 = nverts is an assignment from integer to string. Strings have the // operator defined, so this was easiest for me.

src/FileType_XDMF.f90
KyleVaughn: Added

src/FileType_XDMF.f90
KyleVaughn: True, but 50 lines is a ways away and assignment is extremely fast, so I think this makes sense.

src/FileType_XDMF.f90
KyleVaughn: The (1) refers to the only child in a lot of places. I feel (1) is better than (ONLY_CHILD) or something like that.

KyleVaughn:
I believe the tests are failing now due to CASL Trilinos being moved.
```
git clone https://github.com/CASL/Trilinos.git
Cloning into 'Trilinos'...
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/CASL/Trilinos.git/'
The command "git clone https://github.com/CASL/Trilinos.git" exited with 128.
```

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1324 @@
...
+  CONTAINS
+    !> @copybrief FileType_XDMF::fopen_XDMFFileType
+    !> @copydoc FileType_XDMF::fopen_XDMFFileType
+    PROCEDURE,PASS :: fopen => fopen_XDMFFileType
+    !> @copybrief FileType_XDMF::fclose_XDMFFileType
+    !> @copydoc FileType_XDMF::fclose_XDMFFileType
+    PROCEDURE,PASS :: fclose => fclose_XDMFFileType
+    !> @copybrief FileType_XDMF::fdelete_XDMFFileType
+    !> @copydoc FileType_XDMF::fdelete_XDMFFileType
+    PROCEDURE,PASS :: fdelete => fdelete_XDMFFileType
```
yl5, line 113: Is any information used from the BaseFileType? If not, I don't see the XDMF should be derived from it. Right now, you have to implement these deferred procedures (and they don't do anything). Maybe XDMF is a data structure or format that is not suitable to be defined as a file type?

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1324 @@
...
+!> @param strpath the string holding the path to the XDMF file
+!> @param mesh the XDMF mesh object
+!>
+SUBROUTINE importFromDisk_XDMFFileType(thisXDMFFile, strpath, mesh)
+  CHARACTER(LEN=*),PARAMETER :: myName='importFromDisk_XDMFFileType'
+  CLASS(XDMFFileType),INTENT(INOUT) :: thisXDMFFile
+  TYPE(StringType),INTENT(INOUT) :: strpath
+  TYPE(XDMFMeshType),INTENT(OUT),TARGET  :: mesh
+  TYPE(XMLFileType) :: xml
+  TYPE(HDF5FileType) :: h5
```
yl5, line 675: Looks like you are reading in multiple files to construct the XDMF data structure so that suggests again it shouldn't be derived from the BaseFileType.

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1324 @@
...
+      DEALLOCATE(thismesh%cell_sets(i)%cell_list)
+    ENDDO
+    DEALLOCATE(thismesh%cell_sets)
+  ENDIF
+ENDSUBROUTINE clear_XDMFMeshType
+!
+!-------------------------------------------------------------------------------
+!> @brief Gets the number of grid levels to a leaf node
+!> @param thismesh the XDMF mesh object
+!>
```
yl5, line 790: Add return variable to comment, e.g.,
!> @returns n blabla...

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1324 @@
...
+RECURSIVE FUNCTION distanceToLeaf_XDMFMeshType(thismesh) RESULT(n)
+  CLASS(XDMFMeshType), INTENT(INOUT) :: thismesh
+  INTEGER(SNK) :: n
+  ! It is assumed that all siblings have children if any sibling has children
+  ! Hence it is sufficient to assess thismesh%children(1)'s depth recursively
+  ! Ex:     Possible                Not Possible
```
yl5, line 796: Nice mini figure:-)

src/FileType_XDMF.f90
```diff
@@ -0,0 +1,1324 @@
...
+  TYPE(XMLElementType),POINTER :: xmle, children(:)
+  TYPE(StringType) :: strIn, strOut
+  INTEGER(SIK) :: i, gridIdx
+  CHARACTER(LEN=200) :: charpath
+
+  ! Initialize the XDMFTopologyList if it has not been
+  IF(.NOT.XDMFTopologyList%has('Topology')) CALL init_XDMFTopologyList()
+
+  !H5
+  ! NOTE: it is assumed that the h5 and xml files have the same name.
```
yl5, line 685: Maybe better to add an exception if prefer to enforce user doing this.

src/FileType_XML.f90
```diff
@@ -55,19 +55,19 @@ MODULE FileType_XML
 !> Derived type for an XML element
 TYPE :: XMLElementType
   !> The number of attributes defined on the element
-  INTEGER(SIK),PRIVATE :: nAttr=0
```
yl5, line 58: If they were set to private for some purpose (e.g., avoid accidentally overriding it), it should be better to write setter/getter to access them. Same thing applies to the below changes.

unit_tests/testFileType_XDMF/testFileType_XDMF.f90
```diff
@@ -0,0 +1,618 @@
...
+#include "UnitTest.h"
+USE ISO_FORTRAN_ENV
+USE UnitTest
+USE IntrType
+USE Strings
+USE FileType_XDMF
+IMPLICIT NONE
+
+REAL(SDK) :: two_pins_pin1_vertices(3,109) = RESHAPE( (/ &
+    0.0000000000000000,        0.0000000000000000,        0.0000000000000000,&
```
yl5, line 19: need _SDK otherwise they will be single precision

unit_tests/testFileType_XDMF/testFileType_XDMF.f90
```diff
@@ -0,0 +1,618 @@
...
+0, 2, 4, &
+1, 2, 0, &
+3, 0, 4  &
+/), (/3, 3/))
+
+
+CREATE_TEST('XDMF TYPE')
+REGISTER_SUBTEST('two_pins',test_two_pins)
+REGISTER_SUBTEST('three_level_grid',test_three_level_grid)
+REGISTER_SUBTEST('three_level_grid_IH', test_three_level_grid_implicit_hierarchy)
```
yl5, line 206: SUBTEST should be to test each procedure (constructor/destructor, type-bound procedure, etc.) rather than each type of problems. COMPONENT_TEST may be used to test each case/problem.

yl5:
Overall looks good, but I'd suggest to reconsider deriving the type from the BaseFileType and rewriting the unit tests.

KyleVaughn:
@yl5 Addressed all of your comments except for making XML element's children attribute private again. I need XML elements to persist in memory after leaving a subroutine, so making an allocatable XML element array, setting a pointer to it, then using the XML setChildren method isn't sufficient, since it just does `thisXMLE%children => children` and the underlying XML elements will be deallocated once leaving the subroutine, so the pointer persists longer than the target. I could potentially write a function just to allocate the children, then use getChildren to work with children one by one, but this goes back to the point of unnecessary functions. I don't think that hiding these attributes makes sense, since the XML class does not have any derived classes, so there is no benefit to writing setters, getters, etc. to provide a general interface to the object. I'm fine with leaving the other attributes private, since they have the methods they need to be useful for the most part, but not being able to access children is a real hindrance to working with the type.

HendersonSC:
> @yl5 Addressed all of your comments except for making XML element's children attribute private again. I need XML elements to persist in memory after leaving a subroutine, so making an allocatable XML element array, setting a pointer to it, then using the XML setChildren method isn't sufficient, since it just does `thisXMLE%children => children` and the underlying XML elements will be deallocated once leaving the subroutine, so the pointer persists longer than the target. I could potentially write a function just to allocate the children, then use getChildren to work with children one by one, but this goes back to the point of unnecessary functions. I don't think that hiding these attributes makes sense, since the XML class does not have any derived classes, so there is no benefit to writing setters, getters, etc. to provide a general interface to the object. I'm fine with leaving the other attributes private, since they have the methods they need to be useful for the most part, but not being able to access children is a real hindrance to working with the type.

I would claim that there are multiple reasons for a setter/getter and there are many reasons for one to exists. Further its existence and/or relevance is wholly unrelated to its internal logic (or lack thereof).

To the benefit of a developer some `setFoo` or `getBar` tied to a private data component, provides an absolute guarantee that you know where the data is changing if it changes. Since you know where it changes, during debugging sessions you can know when it changes, which inevitably leads to why it changes. Which makes things way easier than hunting for where a variable changes.

Take note, that I haven't reviewed the underlying code, just agreeing with @yl5. It is plausible that it needs to be public, but the reason it needs to be public can not be that the appropriate setter/getter is too simple. If it is just the technical problem of a pointer outliving target or I have somehow missed something, feel free to email me and/or ping me on slack.

KyleVaughn:
@yl5 Talked with Brendan, and I'm going to keep it private. Will update later today.

yl5:
@KyleVaughn, just a couple things as I saw your updates:

1) If you agree that XDMF is not a file type, maybe change the module and source file names as well. Maybe XMDFDataType or XMDFDataLibrary?

2) I might not be clear about the unit tests. The purpose is to test each subroutine independently to make sure we get expected outputs from the input parameters. We should write a SUBTEST for each type-bound or public procedure as the minimum. Within each SUBTEST, you can test different problem conditions (by varying the input parameters) to thoroughly cover the codes in that procedure, and this is where you may use the COMPONENT_TEST. The current implementation seems to cover the ballpark of the testing but you will need to breakdown/restructure it to test each procedure independently. For an integration test of multiple routines and the overall feature, that is normally what a regression test does.

KyleVaughn:
@yl5 Renamed the module to XDMFMesh and removed the XDMFFileType, so import and export are simply importXDMFMesh and exportXDMFMesh. 

The unit tests were rewritten to maintain independence as best as possible. Functions are used in other tests once that test has passed, like the clear routine and assignment operator. The export tests use the import subroutine to verify, which I'm not happy about, but setting up all the meshes that should be tested by hand would be tedious and a lot of code. Importing the exported meshes also allows me to access the written data easier. 

The import and export now have component tests for the various cases that test the major branches in logic, as denoted by the table in each function. 

Lastly, the XML children attribute was made private again.

yl5 (approved):
Comments were addressed properly. Looks good to me now.

PR #321, opened 2021-04-09
Fix uninitialized value in IAPWS module

aarograh:
Fixes an uninitialized variable conditional jump in IAPWS

### DIFF
--- a/src/IAPWS_WaterProperties.f90
+++ b/src/IAPWS_WaterProperties.f90
@@ -3209,7 +3209,7 @@ ENDFUNCTION tph5n
 SUBROUTINE regsopt(P, T, ireg)
   REAL(SRK), INTENT(IN) :: P
   REAL(SRK), INTENT(IN) :: T
-  INTEGER(SIK), INTENT(INOUT) :: ireg
+  INTEGER(SIK), INTENT(OUT) :: ireg
 
   REAL(SRK), PARAMETER :: &
       rpsr = 5.E-4_SRK, &! relative tolerance value.  should not be lower than 2.E-4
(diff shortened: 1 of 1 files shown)


src/IAPWS_WaterProperties.f90
```diff
@@ -3219,8 +3219,6 @@ SUBROUTINE regsopt(P, T, ireg)
   REAL(SRK), SAVE :: told = 0.0_SRK, pold = 0.0_SRK
   INTEGER(SIK), SAVE :: iregold = 0
 
-  IF (ireg /= 0 .AND. ireg /= 1 .AND. ireg /= 2 .AND. ireg /= 3 .AND. ireg /= 5 .AND. ireg /= 9) ireg = 0
```
bscollin, line 3222: Should we leave an IF(PRESENT(ireg)) do this check.

bscollin: nevermind... its not optional

bscollin: maybe make ireg intent out only?

HendersonSC: Dang beat me to it. It appears to me that `ireg` is intent(out) in all cases. If possible it might be worth while to update that and then set it to 0 on entry.

aarograh: Yeah, good suggestions.  I made it intent out.  I also moved the `ireg=0` assignment up some, though not quite to the top.  I figured there was no reason to do 2 assignments in a row if t and p haven't changed.  I doubt a performance hit would ever be detectable but I think it's still pretty clear what's going on with the current version.

bscollin (approved):

src/IAPWS_WaterProperties.f90
HendersonSC: Looks totally reasonable to me.

HendersonSC (approved):

PR #327, opened 2021-05-13
Add gatherv generic to MPI environment

aarograh:
gatherv is similar to gather, but it allows each process to send a differently sized array, which is useful for many cases.  This implements the specific procedures for 1D arrays of SSK and SDK reals.  Other interfaces can be added later.


HendersonSC (approved):
All things considered this looks good.

PR #330, opened 2021-05-24
Hdf5 auto open

aarograh:
Adds some better DBC checks in HDF5 and automatically opens the file if it's initialized and the client is trying to read or query the file.


src/MemProf.f90
```diff
@@ -74,11 +74,12 @@ MODULE MemProf
...
 !-------------------------------------------------------------------------------
-!> @brief the procudere to
+!> @brief the procedure to intialize a memory profiler object
 !> @param thisMP the memory profiler
-!> @param params
+!> @param pe the parallel environment
+!> @param myLog the log file; optional
+!> @param params the input parameter list; optional
 !>
-!TODO  Need to get control rod movement passed in
```
HendersonSC, line 81: All of these lines get an emphatic approval.

HendersonSC (approved):
I like the changes to hdf5, I have contempalted doing that myself a few times. Now edits doesn't have to constantly be pestering with the `isOpen` mess.

HendersonSC (approved):

PR #333, opened 2021-06-04
Hotfix for Distributed Matrix Initialization

mattkab2:
The MPI_IN_PLACE keyword needs to be swapped from the send to recv buffer in an allreduce call.
Current implementation fails (silently) to correctly sum data on some problems (including 3-mini in the native CMFD heavies.

This was not observed before, and I don't know why it is an issue now. The layout of the CMFD matrix should not have changed. The only change made before observing this issue was the update of the MPACT branch to latest master.

Additionally, the documentation for MPI seems to indicate the MPI_IN_PLACE keyword is valid in either the send/recv buffer. However, this is clearly not the case here.


djabaay:
Looks fine to me @mattkab2.  I grep-ed other calls for the MPI_IN_PLACE and all the other calls are consistent with your change.

djabaay (approved):
Since the change is tested in MPACT, nothing else should be needed.

PR #334, opened 2021-06-15
WIP: resolve invalid math

HendersonSC:
Turned on fpe-trap and found some invalid math....

Some of which is resolved in this PR. There are changes in this PR that will likely have downstream effects, mostly because there is for sure 1 instance of non-standard code fixed.

Edit: the non-standard code was actually not in Futility...so that is kind of a good thing.

### DIFF
--- a/cmake/Futility_Configurations.cmake
+++ b/cmake/Futility_Configurations.cmake
@@ -251,7 +251,7 @@ ELSEIF(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
     IF(${PROJECT_NAME}_FPE_TRAP)
         SET(Fortran_FLAGS
             ${Fortran_FLAGS}
-            ${CSYM}ffpe-trap=invalid,zero,overflow,underflow,denormal
+            ${CSYM}ffpe-trap=invalid,zero#,overflow,underflow,denormal
            )
     ENDIF()
 
--- a/cmake/Futility_Configurations.cmake
+++ b/cmake/Futility_Configurations.cmake
@@ -251,7 +251,7 @@ ELSEIF(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
     IF(${PROJECT_NAME}_FPE_TRAP)
         SET(Fortran_FLAGS
             ${Fortran_FLAGS}
-            ${CSYM}ffpe-trap=invalid,zero#,overflow,underflow,denormal
+            ${CSYM}ffpe-trap=invalid,zero,overflow,underflow,denormal
            )
     ENDIF()
 
--- a/src/Geom_Line.f90
+++ b/src/Geom_Line.f90
@@ -359,7 +359,7 @@ ELEMENTAL FUNCTION distance_LineType_to_PointType(line,p) RESULT(d2)
   CLASS(LineType),INTENT(IN) :: line
   TYPE(PointType),INTENT(IN) :: p
   REAL(SRK) :: d2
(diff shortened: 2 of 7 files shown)


cmake/Futility_Configurations.cmake
```diff
@@ -251,7 +251,7 @@ ELSEIF(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
     IF(${PROJECT_NAME}_FPE_TRAP)
         SET(Fortran_FLAGS
             ${Fortran_FLAGS}
-            ${CSYM}ffpe-trap=invalid,zero,overflow,underflow,denormal
+            ${CSYM}ffpe-trap=invalid,zero#,overflow,underflow,denormal
```
HendersonSC, line 254: Its unlikely we can handle all of these checks in one swoop.

src/Geom_Line.f90
```diff
@@ -359,7 +359,7 @@ ELEMENTAL FUNCTION distance_LineType_to_PointType(line,p) RESULT(d2)
   CLASS(LineType),INTENT(IN) :: line
   TYPE(PointType),INTENT(IN) :: p
   REAL(SRK) :: d2
-  d2=-1._SRK
+  d2=-HUGE(d2)
```
HendersonSC, line 362: This is probably unecessary. The reason I initially added this is that the routine can generate a negative value by accident, contrary to the description. That being said I believe the magnitude of that value is guaranteed to be quite low.

src/Geom_Poly.f90
```diff
@@ -411,7 +411,9 @@ SUBROUTINE calcCentroid(this)
     ENDDO
   ENDIF
   CALL this%centroid%clear()
-  CALL this%centroid%init(DIM=2,X=xcent/this%area,Y=ycent/this%area)
+  IF(this%area > 0.0_SRK) THEN
+    CALL this%centroid%init(DIM=2,X=xcent/this%area,Y=ycent/this%area)
+  ENDIF
```
aarograh, line 416: But what if it's not > 0?

HendersonSC: Yeah I guess it could either be the be the centroid of a point, in which case  maybe this is appropriate
```
IF(this%area > 0.0_SRK) THEN
  CALL this%centroid%init(DIM=2,X=xcent/this%area,Y=ycent/this%area)
ELSE
  CALL this%cnetroid%init(DIM=2,X=xcent,Y=ycent)
ENDIF
```
or the centroid of negative area, which should probably throw an error.

cmake/Futility_Configurations.cmake
HendersonSC: No the PR has no real reason to keep this, this was for my convenience and to make sure anyone that follows behind me is certain to be on the same configuration.

src/Geom_Poly.f90
aarograh: that seems like a reasonable approach

aarograh:
@HendersonSC Do you know if we run MPACT master with this branch if there are any failures or anything?  If not, we could go ahead and merge this in without the corresponding MPACT MR

HendersonSC:
Going to try to figure that out....I think that there are some unit tests that fail with these changes, but I am thinking those can be resolved pretty easily

HendersonSC:
Pushed some changes that fixed the Futility unit tests which failed locally. None of the tests failed upstream as far as I can tell.

PR #336, opened 2021-07-08
Adds gatherv for rank-1 SNK and SLK integers

aarograh:


aarograh:
@HendersonSC I've run the test locally and it passes.  I'll likely merge this without the CI since it's pretty fubar right now and MPACT requests are waiting on it.  The MPACT pipeline will expose any issues.

HendersonSC (approved):
It appears that added code is just an integer versions of existing algorithms for reals. I think the risk associated with merging without ci testing is low.

HendersonSC (approved):
Well that is wild

PR #339, opened 2022-02-03
Add extrapolation to interpolators.

thfolk:
A 1-D extrapolation function is needed for the critical buckling calculation in MPACT.

I added a 1-D extrapolation to the 1-D interpolation, however, I broke unit tests in the process. Should extrapolation be expanded to the 2-D and 3-D bilinear interpolation routines as well? Or should I keep the extrapolation separate, find a new home for it, in order to maintain backwards compatibility to any routine that was calling interpolation and not expecting to return an extrapolated value?

The original interpolation routine worked like this: if x is between the range of given table, interpolate. If x is outside of range, return closest table value but don't extrapolate. 

These current changes did not catch any failures downstream in MPACT.


aarograh:
@thfolk Futility is now managed at https://code.ornl.gov/futility/Futility.  Please open a merge request there and we can review it.

aarograh:
> A 1-D extrapolation function is needed for the critical buckling calculation in MPACT.
> 
> I added a 1-D extrapolation to the 1-D interpolation, however, I broke unit tests in the process. Should extrapolation be expanded to the 2-D and 3-D bilinear interpolation routines as well? Or should I keep the extrapolation separate, find a new home for it, in order to maintain backwards compatibility to any routine that was calling interpolation and not expecting to return an extrapolated value?
> 
> The original interpolation routine worked like this: if x is between the range of given table, interpolate. If x is outside of range, return closest table value but don't extrapolate.
> 
> These current changes did not catch any failures downstream in MPACT.

Let's have this discussion on the new server so that it will be appropriately tied to the final merge request.  Just copy paste that comment on the new MR once you've opened it

